PR #7, opened 2021-06-24
Add WPT compatibility + CLI arguments

sadym-chromium:
To provide compatibility with WPT tests, server has to accept some CLI parameters.


README.md
```diff
@@ -10,30 +10,40 @@ This is a Node.js project, so install dependencies as usual:
...
 
     PORT=8081 npm run bidi-server
+    npm run bidi-server -- --port=8081
 
 Use the `DEBUG` environment variable to see debug info:
 
     DEBUG=* npm run bidi-server
 
-Use the `HEADLESS=false` environment variable to run browser in headful mode:
+Use the argument `--headless=false` environment variable to run browser in headful mode:
```
sadym-chromium, line 38:
```suggestion
Use the CLI argument `--headless=false` to run browser in headful mode:
```

README.md
```diff
@@ -10,30 +10,40 @@ This is a Node.js project, so install dependencies as usual:
...
     DEBUG=* npm run bidi-server
 
-Use the `HEADLESS=false` environment variable to run browser in headful mode:
+Use the CLI argument `--headless=false` to run browser in headful mode:
+
+    npm run bidi-server -- --headless=false
+
+### Starting on Linux
+
+You can alson run the Server by using the shell script `.shell/runBiDiServer.sh`. It will write output to the file `log.txt`:
```
foolip, line 44: Does this script need to be inside a `.shell/` directory? This would make sense if the script is only used internally by `npm run something` and shouldn't be seen/used by users, but that's not the case here.

README.md
```diff
@@ -10,30 +10,40 @@ This is a Node.js project, so install dependencies as usual:
...
 
-    HEADLESS=false npm run bidi-server
+    sh .shell/runBiDiServer.sh --port=8081 --browser="example/path/to/browser" --headless=false
```
foolip, line 46: The `sh` prefix here isn't needed, `.shell/runBiDiServer.sh` is executable and starts with `#!/bin/sh`. My rule of thumb is that scripts should either be executable and be called directly, or they shouldn't be executable and be called via `bash` or `sh`.

README.md
```diff
@@ -10,30 +10,40 @@ This is a Node.js project, so install dependencies as usual:
...
 Use the `DEBUG` environment variable to see debug info:
 
     DEBUG=* npm run bidi-server
 
-Use the `HEADLESS=false` environment variable to run browser in headful mode:
+Use the CLI argument `--headless=false` to run browser in headful mode:
+
+    npm run bidi-server -- --headless=false
+
+### Starting on Linux
```
foolip, line 42: There's probably not much work needed to make it work on macOS and Windows too. What would this section be called if it's OS-agnostic?

.shell/runBiDiServer.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+# Get location of the script.
+SCRIPT=$(readlink -f $0)
```
foolip, line 4: `readlink -f` doesn't work on macOS, you get the error "readlink: illegal option -- f"

Just `cd "$(dirname $0)/.."` will work here I think. (Comment will need tweaking.)

package.json
```diff
@@ -6,7 +6,8 @@
   "main": "index.ts",
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
-    "bidi-server": "npm run build && node ./src/.build/index.js",
+    "bidi-server": "npm run build && npm run server --",
```
foolip, line 9: What does the trailing `--` do here? Unless it avoid the need for `--` between `npm run bidi-server` and arguments I suspect it doesn't do anything :)

package.json
```diff
@@ -16,8 +17,8 @@
   "dependencies": {
     "@types/rollup": "^0.54.0",
     "@types/websocket": "^1.0.2",
+    "argparse": "^2.0.1",
     "debug": "^4.3.1",
-    "http": "*",
```
foolip, line 20: Makes sense, https://nodejs.org/api/http.html is built in.

src/bidiServerRunner.ts
```diff
@@ -27,19 +27,31 @@ export class BidiServerRunner {
...
     onBidiConnectionClosed: (browserProcess: BrowserProcess) => void
   ) {
     const self = this;
-    const bidiPort = parseInt(process.env.PORT) || 8080;
 
     const server = http.createServer(function (request, response) {
       debugInternal(new Date() + ' Received request for ' + request.url);
-      response.writeHead(404);
+
+      // Needed for WPT compatibility.
```
foolip, line 39: I guess this works, but how about giving this response only for `POST /session` requests to make it easier to figure out what this is used for?

src/bidiServerRunner.ts
```diff
@@ -27,19 +27,31 @@ export class BidiServerRunner {
...
+      response.writeHead(200, { 'Content-Type': 'text/plain' });
```
foolip, line 40:
```suggestion
      response.writeHead(200, { 'Content-Type': 'application/json; charset=utf-8', 'Cache-Control': 'no-cache' });
```

Matching https://w3c.github.io/webdriver/#dfn-send-a-response. Maybe link that here too.

src/browserLauncher.ts
```diff
@@ -43,10 +43,11 @@ export class BrowserProcess {
   }
 }
 
-export async function launchBrowser(): Promise<BrowserProcess> {
+export async function launchBrowser(
+  browserExecutablePath,
+  headless
```
foolip, line 48:
```suggestion
  headless: boolean
```

I don't know if this is valid TypeScript, but giving types of arguments and return value is what I think will catch the most mistakes when working with type systems like this. (I have some experience with mypy, none with TypeScript.)

src/browserLauncher.ts
```diff
@@ -97,9 +100,16 @@ async function _waitForWSEndpoint(browserProcess): Promise<string> {
     _addEventListener(rl, 'line', onLine);
 
     function onLine(line) {
+      const errorMatch = line.match(/Unable to open X display/);
```
foolip, line 103: Is this the error you get on macOS and Windows too? Or does the error never happen there in the first place?

src/index.ts
```diff
@@ -20,22 +20,68 @@ import mapperReader from './mapperReader.js';
...
+import argparse from 'argparse';
+
+function parseArguments() {
+  var parser = new argparse.ArgumentParser({
+    add_help: true,
+    exit_on_error: false,
+  });
+
+  parser.add_argument('-p', '--port', {
+    help: 'Port, BiDi server should listen to. Default is 8080.',
```
foolip, line 32:
```suggestion
    help: 'Port that BiDi server should listen to. Default is 8080.',
```

src/index.ts
```diff
@@ -20,22 +20,68 @@ import mapperReader from './mapperReader.js';
...
+  return args[0];
+}
 
 (async () => {
   try {
     console.log('Launching BiDi server.');
-    BidiServerRunner.run(_onNewBidiConnectionOpen, _onBidiConnectionClosed);
+
+    const args = parseArguments();
+    const _bidiPort = parseInt(args.port);
```
foolip, line 56: Hopefully the `parseInt` can be avoided here by pushing the type into the argument definition.

Also, leading underscores for local variables seems odd, they're always "private" in some sense and can't be accidentally used elsewhere, which I guess is why `_onNewBidiConnectionOpen` is prefixed.

src/index.ts
```diff
@@ -20,22 +20,68 @@ import mapperReader from './mapperReader.js';
...
+    add_help: true,
+    exit_on_error: false,
+  });
+
+  parser.add_argument('-p', '--port', {
+    help: 'Port, BiDi server should listen to. Default is 8080.',
+    default: 8080,
+  });
+
+  parser.add_argument('-hl', '--headless', {
```
foolip, line 36: Can `type: 'boolean'` be added here to avoid the string&rarr;boolean conversion later?

src/index.ts
```diff
@@ -20,22 +20,68 @@ import mapperReader from './mapperReader.js';
...
 import { IServer } from './iServer.js';
+import argparse from 'argparse';
+
+function parseArguments() {
+  var parser = new argparse.ArgumentParser({
+    add_help: true,
+    exit_on_error: false,
+  });
+
+  parser.add_argument('-p', '--port', {
```
foolip, line 31: Can `type: 'int'` be added here to take care of the int parsing too?

package.json
sadym-chromium: trailing `--` is needed to pass the cli arguments inside the script. otherwise they will be interpreted as an NPM params.
https://docs.npmjs.com/cli/v7/commands/npm-run-script

src/browserLauncher.ts
sadym-chromium: I faced this issue when I tried to run a headful browser via `ssh`. The script just silently hang. I decided to add the specific error, and of cause It can be extended with other use cases.

README.md
sadym-chromium: I consider the script as a helper, which is not that different form `.vscode` files. So I decided to add it there. Can you think of any better place for the script?

src/index.ts
sadym-chromium: `argparse` supports only `int`, `float` and `str`: https://github.com/nodeca/argparse/blob/master/doc/port_difference.md

src/index.ts
sadym-chromium: done

README.md
sadym-chromium: we can add it later, but I don't want to focus on it now

sadym-chromium: Works on Mac. Added TODO to verify if the script works on Windows as well.

.shell/runBiDiServer.sh
sadym-chromium: done

src/browserLauncher.ts
sadym-chromium: Extended the match logic with an array of regex, so the list can be adjusted with any other problems with launching Chrome.

src/index.ts
sadym-chromium: done

src/bidiServerRunner.ts
sadym-chromium: Added `/session*` handler. I don't want to specify all the proper HTTP handlers yet.

foolip: Did you forget to push a commit? I don't see anything limiting the handler to `/session*` in the diff.

README.md
```diff
@@ -10,30 +10,42 @@ This is a Node.js project, so install dependencies as usual:
...
 ## Starting the Server
 
     npm run bidi-server
 
-This will run the server on port 8080. Use the `PORT` environment variable to
-run it on another port:
+This will run the server on port 8080. Use the `PORT` environment variable or `--port=...` argument to run it on another port:
 
     PORT=8081 npm run bidi-server
+    npm run bidi-server -- --port=8081
```
foolip, line 32: Since the `--browser` argument is now required, can you include it in these examples too?

README.md
```diff
@@ -10,30 +10,42 @@ This is a Node.js project, so install dependencies as usual:
...
+
 If it's a newly downloaded binary, first run it manually once to ensure the operating system trusts it.
 
 ## Starting the Server
 
     npm run bidi-server
 
-This will run the server on port 8080. Use the `PORT` environment variable to
-run it on another port:
+This will run the server on port 8080. Use the `PORT` environment variable or `--port=...` argument to run it on another port:
```
foolip, line 29: Looks like the `PORT` environment variable no longer works. It doesn't seem very important to keep it working, but can you either remove it from the docs or make it work again?

src/index.ts
```diff
@@ -20,22 +20,69 @@ import mapperReader from './mapperReader.js';
...
+}
 
 (async () => {
   try {
     console.log('Launching BiDi server.');
-    BidiServerRunner.run(_onNewBidiConnectionOpen, _onBidiConnectionClosed);
+
+    const args = parseArguments();
+
+    const bidiPort = args.port;
```
foolip, line 58: This is where you'd have to add `PATH` support, but `args.port || process.env.PATH` won't work because `args.port` has a default. And `process.env.PATH || args.port` would make the environment variable take precedent over the argument, which is unusual and not great.

So just dropping it seems fine :)

README.md
sadym-chromium: Added back respect of environment variable in case of no arg passed.

README.md
sadym-chromium: Made `--browser` optional in case of environment variable `BROWSER_PATH` is set. It'll be useful for `.vscode` configurations.

src/index.ts
sadym-chromium: Done by tricking the default value.

foolip (approved):

PR #12, opened 2021-07-02
!Don't merge yet! Add unit tests

sadym-chromium:
# UPD: Still not sure if the `ts-mockito` is the best choice. It has a nice interface and some advantages comparing to `Sinon`, like automatic creating mock by interfaces. But it looks like it has poor support: https://github.com/NagRock/ts-mockito/issues/212


.gitignore
```diff
@@ -3,4 +3,4 @@ src/bidiMapper/.build/
 tests/__pycache__/
 package-lock.json
 src/.build/
-log.txt
+log.txt
```
mathiasbynens, line 6: Please configure your editor to add trailing newlines at EOF. It reduces unnecessary line changes in future commits and avoids Git warnings. In VS Code: `"files.insertFinalNewline": true`

src/bidiMapper/utils/log.ts
```diff
@@ -15,16 +15,19 @@
...
   return (...messages: any[]) => {
-    const elementId = type + '_log';
+    console.log.apply(null, [type].concat(messages));
 
-    if (!window.document.getElementById(elementId)) {
-      window.document.documentElement.innerHTML += `<h3>${type}:</h3><pre id='${elementId}'></pre>`;
-      window.document.getElementById(elementId);
-    }
+      // If run in browser, add debug message to the page.
+    if (typeof window !== 'undefined' && window && window.document && window.document.documentElement) {
```
mathiasbynens, line 21: This could be simplified using [optional chaining](https://v8.dev/features/optional-chaining)

src/bidiMapper/utils/log.ts
```diff
@@ -15,16 +15,19 @@
  */
 export function log(type: string): (...message: any[]) => void {
   return (...messages: any[]) => {
-    const elementId = type + '_log';
+    console.log.apply(null, [type].concat(messages));
```
mathiasbynens, line 18: Simpler:

```suggestion
    console.log(type, ...messages);
```

src/bidiMapper/utils/log.ts
```diff
@@ -15,16 +15,19 @@
...
-      window.document.getElementById(elementId);
-    }
+      // If run in browser, add debug message to the page.
+    if (typeof window !== 'undefined' && window && window.document && window.document.documentElement) {
+        const elementId = type + '_log';
 
-    const element = window.document.getElementById(elementId);
+        if (!window.document.getElementById(elementId)) {
+          window.document.documentElement.innerHTML += `<h3>${type}:</h3><pre id='${elementId}'></pre>`;
+          window.document.getElementById(elementId);
```
mathiasbynens, line 26: What is the purpose of this line?

.gitignore
sadym-chromium: done

src/bidiMapper/utils/log.ts
sadym-chromium: `typeof window !== 'undefined'` is still needed, but the next chain is simplified. Thanks!

src/bidiMapper/utils/log.ts
sadym-chromium: Artefact. Removed.

src/bidiMapper/utils/log.ts
mathiasbynens: Looks like this wasn't applied?

mathiasbynens (approved):

src/bidiMapper/utils/cdpServer.spec.ts
```diff
@@ -0,0 +1,22 @@
+import { CdpServer } from './cdpServer';
+import { ServerBinding } from './iServer';
+import { mock, instance, verify } from 'ts-mockito';
+
+describe('CdpServer tests.', async () => {
+  it('given CdpServer, when `sendMessage` is called, then cdpBindings should be called with proper values', async () => {
```
bwalderman, line 6: Minor: With Mocha's interface, I think `describe('CdpServer')` serves the same purpose of saying "given CdpServer" and then "given CdpServer" doesn't need to be repeated in the test case names.

Suggestion for mocha-compatible style:

```js
describe('CdpServer', async () => {
  it('calls cdpBindings with proper values when `sendMessage` is called', async () => {
```

bwalderman (approved):
LGTM! Just left one minor comment on test naming.

src/bidiMapper/utils/cdpServer.spec.ts
sadym-chromium: done

src/bidiMapper/utils/log.ts
sadym-chromium: my bad. Fixed now.

src/bidiMapper/utils/log.ts
```diff
@@ -15,16 +15,18 @@
  */
 export function log(type: string): (...message: any[]) => void {
   return (...messages: any[]) => {
-    const elementId = type + '_log';
+    console.log(type, ...messages);
```
foolip, line 18: This (or some other `console.log`) is showing up in the output of `npm run unit`, I see this:
```
  CdpServer tests.
cdp sent > {"someAttribute":"someValue","id":0}
    ✔ given CdpServer, when `sendMessage` is called, then cdpBindings should be called with proper values
```

That "cdp sent" line would be nice to get rid of. If there is logging that's useful for debugging, then we should probably use a logging mechanism which can be disabled when running tests.

src/bidiServerRunner.ts
```diff
@@ -27,19 +27,38 @@ export class BidiServerRunner {
...
     onBidiConnectionClosed: (browserProcess: BrowserProcess) => void
   ) {
     const self = this;
-    const bidiPort = parseInt(process.env.PORT) || 8080;
 
     const server = http.createServer(function (request, response) {
       debugInternal(new Date() + ' Received request for ' + request.url);
-      response.writeHead(404);
+
+      // Needed for WPT compatibility.
```
foolip, line 39: This was already landed in https://github.com/GoogleChromeLabs/chromium-bidi/pull/7/files, why is it also showing up in this PR despite https://github.com/GoogleChromeLabs/chromium-bidi/pull/7 being merged?

Hmm, I think it's because there are conflicts in `package.json`, so GitHub can't merge into `main` in order to compare against `main`...

foolip:
On the mocking library to use, I've used Sinon with pure JS before. I don't exactly love it, but it gets the job done. Would it work to depend on Sinon and https://www.npmjs.com/package/@types/sinon perhaps? That has a lot of downloads at least, suggesting we'd not be alone :)

src/bidiServerRunner.ts
sadym-chromium: done

src/bidiMapper/utils/log.ts
sadym-chromium: done. Log only if run in browser.

bwalderman:
@sadym-chromium , thanks for adding this. My next PR will bring some big changes to CdpClient, so it might be best to wait before creating a CdpClient unit test. Thanks!

sadym-chromium:
Closed in favour of #20 as `Sinon` appeared to be quite ergonomic for our needs.

PR #15, opened 2021-07-13
Typed CDP Bindings - Part 1

bwalderman:
This change adds a dependency on https://www.npmjs.com/package/devtools-protocol which contains updated .json and .d.ts files for CDP.

To enable the mapper script to import this module, I've added a rollup config to be run during build. It is configured to allow node module imports.

As a quick test, I've imported the `version` property from `devtools-protocol` module to verify that this builds and is rolled up into the output file as expected.

A follow-up change will add a strongly-typed layer on top of CdpClient.

### DIFF
--- a/src/bidiMapper/utils/cdpClient.ts
+++ b/src/bidiMapper/utils/cdpClient.ts
@@ -17,6 +17,10 @@ import { ServerBinding, AbstractServer } from './iServer';
 import { log } from './log';
 const logCdp = log('cdp');
 
+// TODO: Remove. This is a quick test to verify we can import a node module.
+import { version } from 'devtools-protocol/json/browser_protocol.json';
+logCdp(`Using CDP version: ${version}`);
+
 export class CdpClient extends AbstractServer {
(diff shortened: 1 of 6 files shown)


src/bidiMapper/utils/cdpClient.ts
```diff
@@ -17,7 +17,11 @@ import { ServerBinding, AbstractServer } from './iServer';
 import { log } from './log';
 const logCdp = log('cdp');
 
-export class CdpServer extends AbstractServer {
+// TODO: Remove. This is a quick test to verify we can import a node module.
+import { version } from 'devtools-protocol/json/browser_protocol.json';
+logCdp(`Using CDP version: ${version}`);
```
sadym-chromium, line 22: When should this be removed?

sadym-chromium (approved):
LGTM with a small comment

src/bidiMapper/utils/cdpClient.ts
bwalderman: I'll remove on the next PR. I just wanted something in here to use the module to verify it is included in the build output.

PR #20, opened 2021-07-16
Unit testing using Sinon

sadym-chromium:
Unit tests using Sinon's stubs + CI running unit tests


src/bidiMapper/utils/Tests/stub.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { assert, createStubInstance, SinonStubbedInstance } from 'sinon';
+
+export class Stub {
```
bwalderman, line 18: Is a class with all static members being used to just so that everything is easier to import all at once? If so, I think it would be more idiomatic for consumers to use `import * as stub from './stub'`. I'd prefer not to define a class unless we really need one (i.e. it carries state and will be instantiated one or more times).

src/bidiMapper/utils/Tests/stub.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+  // Returns mock object of a given type.
+  static stub<T>(constructor: new (...args) => T): StubbedClass<T> {
```
bwalderman, line 20: What are the advantages to stub/StubbedClass<T> over using createStubInstance directly?

src/bidiMapper/utils/Tests/stub.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+  >(mock: StubbedClass<T>): (string) => void {
+    Stub.assert.calledOnce(mock.setOnMessage);
+    const onMessage = mock.setOnMessage.getCall(0).args[0] as any as (
+      string
+    ) => void;
+    return onMessage;
+  }
+
+  // Export `assert` to simplify imports.
+  static assert = assert;
```
bwalderman, line 38: I'm not sure I like this idea. Making imports easier is nice, but as someone unfamiliar reading this code, if I see an `assert` function being imported from a different location than 'sinon', I would wonder if it's a different function than sinon's `assert`. I think it would be better for readability if we import things we don't change directly from 'sinon'.

bwalderman (changes requested):
Generally LGTM. Excited to have unit tests up and running! I just left some questions/comments about how Sinon is being used.

src/bidiMapper/utils/Tests/stub.spec.ts
sadym-chromium: The advantage is in casting the result to `StubbedClass`, which is simultaneously `SinonStubbedInstance<T>`  and `T` itself, which allows to avoid disturbing casting in each stub usage.

sadym-chromium: The idea was to make a stubbed instance of type `SinonStubbedInstance<T> | T`, which would allow to both use it as a parameter for downstream classes and to verify stubs. For now `StubServer` solves the issue by making it's methods both stubs and methods.

src/bidiMapper/utils/Tests/stub.spec.ts
sadym-chromium: Having one `import` shows other developers the common way of developing tests. But I don't insist on this approach and removed it in current revision.

src/bidiMapper/utils/Tests/stub.spec.ts
sadym-chromium: Simplified and removed redundancy

.github/workflows/unit.js.yml
```diff
@@ -0,0 +1,33 @@
...
+name: Unit tests
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  build:
```
foolip, line 13: This shows up as the name of the check. I would suggest renaming it "test", "unittests" or something like that.

.github/workflows/unit.js.yml
```diff
@@ -0,0 +1,33 @@
...
+    - name: Set up Node.js
+      uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node }}
+    - name: Install dependencies
+      run: npm install
+    - name: Build
+      run: npm run build
+    - name: Run unit tests
+      run: npm run unit
```
foolip, line 33: I would suggest appending this to the end of the file:
```
env:
  FORCE_COLOR: 3
```

Then we will get colors in the output on GitHub, making it easier to see green checkmarks and red errors.

src/bidiMapper/mapper.ts
```diff
@@ -68,18 +68,24 @@ const _waitSelfTargetIdPromise = _waitSelfTargetId();
...
-        if (this.onmessage) {
-          this.onmessage.call(null, message);
+        if (this._onMessage) {
+          this._onMessage.call(null, message);
         }
       };
     }
 
-    sendMessage(message: string): void {
+    async setOnMessage(
```
foolip, line 82: Isn't this entirely synchronous?

src/bidiMapper/mapper.ts
```diff
@@ -68,18 +68,24 @@ const _waitSelfTargetIdPromise = _waitSelfTargetId();
...
+      onMessage: (messageObj: any) => Promise<void>
+    ): Promise<void> {
+      this._onMessage = onMessage;
+    }
+
+    async sendMessage(message: string): Promise<void> {
```
foolip, line 88: You added `async` here, but the implementation doesn't `await`. Is TypeScript different from JS, or is that just missing?

src/bidiMapper/utils/cdpServer.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { connectCdp } from './cdpClient';
+import { StubServer } from './tests/stubServer.spec';
+import { assert as chai_assert } from 'chai';
+import { assert as sinon_assert } from 'sinon';
+import { Protocol } from 'devtools-protocol';
+
+describe('CdpClient tests.', async () => {
```
foolip, line 22:
```suggestion
describe('CdpClient tests.', () => {
```

Looks like this doesn't need to be async, and IIRC, mocha doesn't support async here, if tests are defined async one needs to pass some command line argument.

src/bidiMapper/utils/cdpServer.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  it(`given some command is called, when CDP command is done, then
+        'sendMessage' promise is resolved`, async () => {
+    const mockCdpServer = new StubServer();
+    const cdpClient = connectCdp(mockCdpServer);
+
+    // Get handler 'onMessage' to notify 'cdpClient' about new CDP messages.
+    const onMessage = mockCdpServer.getOnMessage();
+
+    // Send CDP command and store result promise.
```
foolip, line 52:
```suggestion
    // Send CDP command and store returned promise.
```

src/bidiMapper/utils/cdpServer.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+    });
+
+    // Verify CDP command was sent.
+    sinon_assert.calledOnce(mockCdpServer.sendMessage);
+
+    // Notify 'cdpClient' the CDP command is finished.
+    onMessage(JSON.stringify({ id: 0, result: {} }));
+
+    // Assert 'cdpClient' resolved message promise.
+    await commandPromise;
```
foolip, line 64: Is there nothing to assert about what this resolves to?

src/bidiMapper/utils/cdpServer.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+  });
+
+  it(`given some command is called 2 times, when CDP commands are done, then
+        each command promise is resolved with proper results`, async () => {
+    const mockCdpServer = new StubServer();
+    const cdpClient = connectCdp(mockCdpServer);
+    const expectedResult1 = {
+      someResult: 1,
+    } as any as Protocol.Target.AttachToTargetResponse;
```
foolip, line 73: This `as` stuff is TypeScript I don't understand. I guess it's the type?

(No change needed, I just don't understand.)

src/bidiMapper/utils/cdpServer.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+    const expectedResult2 = {
+      anotherResult: 2,
+    } as any as Protocol.Target.AttachToTargetResponse;
+    const commandResult1 = { id: 0, result: expectedResult1 };
+    const commandResult2 = { id: 1, result: expectedResult2 };
+
+    // Get handler 'onMessage' to notify 'cdpClient' about new CDP messages.
+    const onMessage = mockCdpServer.getOnMessage();
+
+    // Send 2 CDP commands and store result promise.
```
foolip, line 83:
```suggestion
    // Send 2 CDP commands and store returned promises.
```

src/bidiMapper/utils/cdpServer.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+      targetId: 'Target_2',
+    });
+
+    // Verify CDP command was sent.
+    sinon_assert.calledTwice(mockCdpServer.sendMessage);
+
+    // Notify 'cdpClient' the command2 is finished.
+    onMessage(JSON.stringify(commandResult2));
+    // Assert second message promise is resolved.
+    const actualResult2 = await commandPromise2;
```
foolip, line 97: The `await` here does mean that the test is verifying that commands aren't in a blocking queue, which is great.

Could we also verify that the order that the promises resolve is correct? This test could also pass if `commandPromise1` is incorrectly resolved at this point.

src/bidiMapper/utils/log.ts
```diff
@@ -15,16 +15,18 @@
  */
 export function log(type: string): (...message: any[]) => void {
   return (...messages: any[]) => {
-    const elementId = type + '_log';
+    // If run in browser, add debug message to the page.
+    if (typeof window !== 'undefined' && window?.document?.documentElement) {
```
foolip, line 19: Since this will only run in recent browsers, maybe this simplification?

```suggestion
    if (globalThis?.document?.documentElement) {
```

You could also skip the `window.` bits below. Just accessing `document` directly is more common in my experience.

src/bidiMapper/utils/log.ts
```diff
@@ -15,16 +15,18 @@
...
 
-    if (!window.document.getElementById(elementId)) {
-      window.document.documentElement.innerHTML += `<h3>${type}:</h3><pre id='${elementId}'></pre>`;
-      window.document.getElementById(elementId);
-    }
+      const elementId = type + '_log';
 
-    const element = window.document.getElementById(elementId);
+      if (!window.document.getElementById(elementId)) {
+        window.document.documentElement.innerHTML += `<h3>${type}:</h3><pre id='${elementId}'></pre>`;
```
foolip, line 25: It looks like this doesn't support more than one `type`, because all the other content will get removed by this line if there's a new `type`.

If multiple `type`s should be supported, maybe rework this to create and add a new section to the document?

src/bidiMapper/utils/log.ts
```diff
@@ -15,16 +15,18 @@
...
+      }
 
-    console.log.apply(null, [type].concat(messages));
-    element.innerText += messages.join(', ') + '\n';
+      const element = window.document.getElementById(elementId);
+      element.innerText += messages.join(', ') + '\n';
```
foolip, line 29: This will get slower and slower with more logs as each time everything needs to be parsed again. Perhaps create and append a `<div>`?

.github/workflows/unit.js.yml
sadym-chromium: done

.github/workflows/unit.js.yml
sadym-chromium: done

src/bidiMapper/mapper.ts
sadym-chromium: the implementation of `window.cdp.send` is synchronous, but the generic method `sendMessage` of `IServer` can be async, so instead of returning resolved promise, I marked it as async.

src/bidiMapper/utils/cdpServer.spec.ts
sadym-chromium: It's implicit by the fact the `commandPromise1` should be resolved with the `commandResult1`, which, in order, was introduced after the `commandPromise2` resolved. 

If you know an easy way to verify the promise is not resolved, I'd be happy to add it. But al the ways I can think of are quite bulky.

src/bidiMapper/mapper.ts
sadym-chromium: fixed.

src/bidiMapper/utils/cdpServer.spec.ts
sadym-chromium: there is another test verifies the resolve result

src/bidiMapper/utils/cdpServer.spec.ts
sadym-chromium: It's a trick for TypeScript to cast any type to any. I consider it acceptable for testing purposes.

src/bidiMapper/utils/log.ts
sadym-chromium: fixed

src/bidiMapper/utils/cdpServer.spec.ts
bwalderman: Mocha supports async test cases but needs the `async function() {}` syntax instead of the `async () => {}` syntax due to how `this` is used: https://mochajs.org/#using-async-await

src/bidiMapper/utils/cdpServer.spec.ts
```diff
@@ -0,0 +1,107 @@
+/**
```
bwalderman, line 1: Please rename file to cdpClient.spec.ts.

src/bidiMapper/utils/cdpServer.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+    await commandPromise;
+  });
+
+  it(`given some command is called 2 times, when CDP commands are done, then
+        each command promise is resolved with proper results`, async () => {
+    const mockCdpServer = new StubServer();
+    const cdpClient = connectCdp(mockCdpServer);
+    const expectedResult1 = {
+      someResult: 1,
+    } as any as Protocol.Target.AttachToTargetResponse;
```
bwalderman, line 73: CdpClient still has a raw sendCommand(method, params) method that is useful here if you don't want to force cast to an AttachToTargetResponse.

src/bidiMapper/utils/cdpClient.ts
```diff
@@ -17,18 +17,14 @@ import { log } from './log';
 const logCdp = log('cdp');
 
 import { EventEmitter } from './events';
+import { IServer } from './iServer';
```
bwalderman, line 20: What do you think about changing the name of IServer to either IBinding or ITransport? I don't think "IServer" accurately describes what this class does. It represents that channel that client and server use to communicate. It's not the server itself.

bwalderman (changes requested):

src/bidiMapper/utils/cdpServer.spec.ts
sadym-chromium: `CdpClientImpl` has the `sendCommand(method, params)` method, but `CdpClient` doesn't. If we need it for a real scenario, we can expose `sendCommand(method, params)` to `CdpClient`, but I'm not sure if we should do it for test purposes.

src/bidiMapper/utils/cdpClient.ts
sadym-chromium: I find `server` is a bad naming as well, and `transport` suites better. But I don't want to do it in current PR, as the scope of this one grows too fast, and we end up with a huge un-reviewable PR.

src/bidiMapper/utils/cdpServer.spec.ts
sadym-chromium: done

src/bidiMapper/utils/cdpServer.spec.ts
sadym-chromium: Another approach would be to test `CdpClientImpl` itself. But we will need to use `rewire` to test not-exported classes, and it wouldn't test wiring "CdpApi" -> "IServer calls". So I'd keep testing `CdpClient` as atomic.

src/bidiMapper/utils/log.ts
foolip: Oh wait, I misunderstood the original code, `+=` here is key.

src/bidiMapper/utils/log.ts
```diff
@@ -16,17 +16,27 @@
 export function log(type: string): (...message: any[]) => void {
   return (...messages: any[]) => {
     // If run in browser, add debug message to the page.
-    if (typeof window !== 'undefined' && window?.document?.documentElement) {
+    if (globalThis?.document?.documentElement) {
```
foolip, line 19:
```suggestion
    if (globalThis.document?.documentElement) {
```

I guess the first `?.` is unnecessary, since `globalThis` can't be null. My bad :)

src/bidiMapper/utils/log.ts
```diff
@@ -16,17 +16,27 @@
...
+      const typeLogContainer = findOrCreateTypeLogContainer(type);
 
-      if (!window.document.getElementById(elementId)) {
-        window.document.documentElement.innerHTML += `<h3>${type}:</h3><pre id='${elementId}'></pre>`;
-      }
-
-      const element = window.document.getElementById(elementId);
-      element.innerText += messages.join(', ') + '\n';
+      const pre = document.createElement('pre');
+      pre.innerText = messages.join(', ');
```
foolip, line 25:
```suggestion
      pre.textContent = messages.join(', ');
```

I would avoid `innerText`, it has very peculiar behavior. `textContent` simply created/updates a single `Text` node, which is much more likely what you want.

src/bidiMapper/utils/log.ts
```diff
@@ -16,17 +16,27 @@
...
+  const elementId = type + '_log';
+  if (!document.getElementById(elementId)) {
+    const typeLogContainer = document.createElement('div');
+    typeLogContainer.id = elementId;
+    typeLogContainer.innerHTML = `<h3>${type}:</h3>`;
+    document.body.appendChild(typeLogContainer);
+    return typeLogContainer;
+  }
+
+  return document.getElementById(elementId);
```
foolip, line 41: Optional nit: do just a single `document.getElementById(elementId)` call and return it if true:
```js
const something = document.getElementById(elementId);
if (something) {
  return something;
}
```

foolip (approved):
LGTM with a few suggestions

src/bidiMapper/utils/log.ts
sadym-chromium: done

.github/workflows/unit.js.yml
```diff
@@ -0,0 +1,34 @@
...
+    branches: [main]
+
+jobs:
+  unit_test:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [14.x]
+        # See supported Node.js release schedule at https://nodejs.org/en/about/releases/
```
mathiasbynens, line 19: Node.js is only used as a build dependency, right? In that case we don’t need to support multiple Node.js versions —  we can always just use the one we want (i.e. the most recent stable). So let’s simplify by dropping `matrix` and just setting `node-version: 14` below?

src/bidiMapper/utils/cdpClient.ts
mathiasbynens: Improving the name as a follow-up patch sounds great!

foolip (approved):

.github/workflows/unit.js.yml
sadym-chromium: done

.github/workflows/unit.js.yml
```diff
@@ -0,0 +1,29 @@
...
+jobs:
+  unit_test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: 14.x
```
mathiasbynens, line 21:
```suggestion
          node-version: 14
```

mathiasbynens (approved):

bwalderman (approved):

PR #22, opened 2021-07-17
Typed CDP Bindings - Part 2

bwalderman:
This changes implements a strongly-typed CdpClient object that implements the interface provided in devtools-protocol module. CdpClient callers have been updated to use the new style.

Before:
```js
await cdpServer.sendMessage({
  method: 'Target.setAutoAttach',
  params: {
    autoAttach: true,
    waitForDebuggerOnStart: false,
    flatten: true,
  },
});
```

After:
```js
await cdpClient.Target.setAutoAttach({
  autoAttach: true,
  waitForDebuggerOnStart: false,
  flatten: true,
});
```

Events:
```js
this._cdpClient.Target.on('attachedToTarget', (params) => {
  this._contextProcessor.handleAttachedToTargetEvent(params);
});
```

Command and event parameters are fully typed. TypeScript interfaces for CDP types are also available from the `Protocol` namespace. e.g.: `Protocol.Target.TargetInfo`. Some functions that accept protocol types like TargetInfo as a parameter have been updated to use these.

The devtools-protocol module only provides types, and no implementation, so the implementation is up to us. The new code in cdpClient.ts reads the protocol .json files and uses this information at runtime to generate classes that implement the expected interface. When a generated method on a generate class is called, it will send a raw CDP message using a CdpTransport object and await a response.

Follow-up:
- There are still some manually-written CDP calls in mapperServer.ts that can be updated to use the new CdpClient. Needs a CdpTransport for websocket.
- Unit test coverage for CdpClient. Waiting for #12

### DIFF
--- a/src/bidiMapper/commandProcessor.ts
+++ b/src/bidiMapper/commandProcessor.ts
@@ -14,18 +14,26 @@
  * limitations under the License.
  */
 import { IServer } from './utils/iServer';
+import { CdpClient } from './utils/cdpClient';
 import { BrowsingContextProcessor } from './domains/context/browsingContextProcessor';
 import { Context } from './domains/context/context';
+import { Protocol } from 'devtools-protocol';
+
+interface BidiCommandMessage {
+  id: number;
+  method: string;
+  params?: any;
+}
 
 export class CommandProcessor {
-  private _cdpServer: IServer;
+  private _cdpClient: CdpClient;
   private _bidiServer: IServer;
   private _selfTargetId: string;
   private _contextProcessor: BrowsingContextProcessor;
 
-  static run(cdpServer: IServer, bidiServer: IServer, selfTargetId: string) {
+  static run(cdpClient: CdpClient, bidiServer: IServer, selfTargetId: string) {
     const commandProcessor = new CommandProcessor(
-      cdpServer,
+      cdpClient,
       bidiServer,
       selfTargetId
     );
@@ -34,18 +42,18 @@ export class CommandProcessor {
   }
 
   private constructor(
-    cdpServer: IServer,
+    cdpClient: CdpClient,
     bidiServer: IServer,
     selfTargetId: string
   ) {
     this._bidiServer = bidiServer;
-    this._cdpServer = cdpServer;
+    this._cdpClient = cdpClient;
     this._selfTargetId = selfTargetId;
   }
 
   private run() {
     this._contextProcessor = new BrowsingContextProcessor(
-      this._cdpServer,
+      this._cdpClient,
       this._selfTargetId,
       (t: Context) => {
         return this._onContextCreated(t);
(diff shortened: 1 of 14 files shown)


src/bidiMapper/commandProcessor.ts
```diff
@@ -55,21 +63,32 @@ export class CommandProcessor {
       }
     );
 
-    this._cdpServer.setOnMessage((messageObj) => {
-      return this._onCdpMessage(messageObj);
```
sadym-chromium, line 59: This method was intended to inject all the message listeners logic and later to be split between different domains.

src/bidiMapper/mapper.ts
```diff
@@ -15,79 +15,85 @@
...
-  const cdpBinding = new ServerBinding(
-    (message) => {
-      globalObj.cdp.send(message);
-    },
-    (handler) => {
-      globalObj.cdp.onmessage = handler;
+function _createCdpClient() {
+  // A CdpTransport implementation that uses the window.cdp bindings
+  // injected by Target.exposeDevToolsProtocol.
+  class WindowGlobalCdpTransport implements CdpTransport {
```
sadym-chromium, line 71: `WindowGlobalCdpTransport` -> `WindowCdpTransport `
```suggestion
  class WindowCdpTransport implements CdpTransport {
```

src/bidiMapper/mapper.ts
```diff
@@ -15,79 +15,85 @@
...
     }
-  );
-  return new CdpClient(cdpBinding);
+
+    sendMessage(message: string): void {
+      window.cdp.send(message);
+    }
+  }
+
+  return connectCdp(new WindowGlobalCdpTransport());
```
sadym-chromium, line 87:
```suggestion
  return connectCdp(new WindowCdpTransport());
```

src/bidiMapper/tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "compilerOptions": {
     "target": "ESNext",
     "moduleResolution": "node",
-    "resolveJsonModule": true
+    "resolveJsonModule": true,
+    "noImplicitAny": true
```
sadym-chromium, line 6: +1

src/bidiMapper/utils/cdpClient.ts
```diff
@@ -13,50 +13,150 @@
...
 
-export class CdpClient extends AbstractServer {
-  private _commandCallbacks: Map<number, (messageObj: any) => void> = new Map();
+import * as browserProtocol from 'devtools-protocol/json/browser_protocol.json';
+import * as jsProtocol from 'devtools-protocol/json/js_protocol.json';
+import ProtocolProxyApi from 'devtools-protocol/types/protocol-proxy-api';
 
-  constructor(cdpBinding: ServerBinding) {
-    super(cdpBinding);
+export type CdpClient = EventEmitter & ProtocolProxyApi.ProtocolApi;
```
sadym-chromium, line 25: I like the approach with dynamic generating methods, but it can confuse IDE. Alternative I can think of is generating a TS file based on JSON. Not sure if we need it now though. WDYT?

src/browserLauncher.ts
```diff
@@ -105,9 +107,7 @@ async function _waitForWSEndpoint(browserProcess): Promise<string> {
   ];
   return new Promise((resolve, reject) => {
     const rl = readline.createInterface({ input: browserProcess.stderr });
-    _addEventListener(rl, 'line', onLine);
-
-    function onLine(line) {
+    rl.on('line', (line) => {
```
sadym-chromium, line 110: +1

sadym-chromium (approved):

sadym-chromium:
I'll wait with merging #20, as it can involve lots of conflicts. So please go ahead and merge this one, and I rebase and merge #20 afterwards (after getting LGTM, of cause).

src/bidiMapper/utils/cdpClient.ts
bwalderman: Right now, the IDE uses the pre-existing types from https://github.com/ChromeDevTools/devtools-protocol/tree/master/types. The nice thing about using pre-built type definitions is that all of these types are known at compile time. The drawback is there is no implementation and we need to generate the implementation at runtime. If there's a bug in how we generate the implementation, then the IDE's understanding of the types won't match reality. I think this code is correct though :)

If we generate our own typescript files that include an implementation, it will ensure the "compile-time" and "runtime" types always match, but we'll need an extra pre-build step to generate the files. The IDE won't recognize any CDP types until the developer builds at least once.

It might be useful to do this in the future if we need more control or want to have a different API than the one defined in `devtools-protocol`. For now though, I think dynamically generating the methods is good enough.

src/bidiMapper/commandProcessor.ts
bwalderman: I think we can still organize this so that different domains are separate. I'm imagining `BrowsingContextProcessor` can have a public method like `registerEventListeners(cdpClient)` that we call from here. The `registerEventListeners` method can register for whatever events the `BrowsingContextProcessor` needs, and the `CommandProcessor` doesn't need to know or care.

sadym-chromium:
I'm merging the change to unblock myself with the #20

PR #26, opened 2021-07-21
Add `unit test` and `e2e` sections into readme

sadym-chromium:


mathiasbynens (commented):
LGTM. Consider writing a more concrete commit message (“Update readme” doesn’t really describe the change, just the file that happens to be changed — how about something like “Document how to run unit and e2e tests”)

mathiasbynens (approved):

Issue #14, opened 2021-07-13
Support both a "standalone" and "embedded" build


bwalderman:
Idea: Support building/running the bidi mapper in two configurations:

- "embedded" - This is the current configuration. Mapper is injected into a browser tab and communicates with the rest of the browser through `exposeDevToolsProtocol`.
- "standalone" - New config. Mapper runs as a node module outside the browser and communicates with the browser through remote-debugging-port/pipe.

Motivation:

Standalone mode would simplify development and testing of mapper code. The bidi server code and mapper code would all run in the same node process which simplifies debugging.

foolip:
This sounds reasonable if it doesn't come with any complications to support both contexts. Can the same API that `exposeDevToolsProtocol` exposes be used in the "standalone" config as well, so that extra code for achieving the same API in both is only needed when running as a node module? That would minimize binary size cost in the hoped-for final configuration of being bundled with the browser.

sadym-chromium:
Unlocked by https://github.com/GoogleChromeLabs/chromium-bidi/commit/0d2b05c871d76607b678621ae0b43ebde1bd99e8

PR #29, opened 2021-07-23
Make CDP implementation usable on both server and browser

bwalderman:
This change moves the CdpClient class and some related files from src/bidiMapper/utils/cdp to src/cdp. I've also moved files from src/bidiMapper/utils that cdp/ depends on. These are now under src/utils.

The end result is that code common to the server and mapper can be placed in src/ and both server and mapper code should be able to import it normally. This is to prepare to use the CdpClient class on the server side, and to help make the "standalone" configuration in #14 possible.

To get the build and unit tests working again after the move, I had to make some changes to the package.json and tsconfig files...

Rollup was having trouble recognizing mapper.ts as a typescript file and so would hit a syntax error when trying to build. I was able to solve this by removing the tsconfig.json file in src/bidiMapper. There is only one tsconfig.json file now (the one in src). This appears to let both the rollup build and server build work as expected. Having a single tsconfig should be okay because it enables code sharing and rollup will only pull in code that it actually needs.

There was also an issue due to the project being defined as an ESM module. Now that there is only one tsconfig, the unit tests files are being built with `"module": "ESNext"` like everything else. However, ts-node doesn't yet fully support ESM module loading (see https://github.com/TypeStrong/ts-node/issues/1007). I spent some time trying to get mocha to run with ts-node's experimental ESM support but I wasn't able to find a quick solution. Instead, I removed the `"type": "module"` property from the project's package.json and set the tsconfig to output "commonjs" modules. This lets the mocha unit tests and the actual product work as expected. Using "commonjs" as the target for Node projects appears to be recommended anyway: https://github.com/tsconfig/bases/blob/main/bases/node16.json


bwalderman:
This change means we can probably revert #27 but I left that out to keep this PR smaller.

sadym-chromium (commented):
In general I like the idea of #14. But I would keep the mapper logic as independent as possible, as in future we would need to have the mapper JS code in ChromeDriver to implement the magic with running mapper in Chrome.

bwalderman:
@sadym-chromium agreed, the mapper logic should be always buildable as an independent target. As long as we only share code that will work in both node.js and browser environments then this should be fine.

sadym-chromium:
ok, let's make it

sadym-chromium (approved):
Ok, let's go this way for now, and will think how to inject it into ChromeDriver later

bwalderman:
@sadym-chromium thanks for the signoff. The mapper does still build as an single injectable file even with this PR. That aspect hasn't change.

PR #30, opened 2021-07-23
Make domain objects extend DomainImpl

bwalderman:
Rename DomainBase to DomainImpl to match name of CdpClientImpl.

Update the type of CdpClient so that all of its domain properties also extend DomainImpl. This lets callers access the EventEmitter methods like removeListener which will be useful later for implementing things like unsubscribe.


sadym-chromium (commented):
It looks like the PR is targeted not to main but to `cdp_refactor_2`. Was is intended to be so?

bwalderman:
> It looks like the PR is targeted not to main but to `cdp_refactor_2`. Was is intended to be so?

Yes. I wanted to merge cdp_refactor_2 first and then I'll rebase this PR onto main.

sadym-chromium (approved):

PR #35, opened 2021-07-26
*experimental* Enable strict typescript compiler

bwalderman:
Adds the [`strict`](https://www.typescriptlang.org/tsconfig#strict) compiler option to the typescript build and fixes the resulting build errors. Enabling strict mode results in stronger guarantees of program correctness. 

Summary of changes:
- Where possible, errors were fixed by ensuring a class's fields are initialized in the constructor so that we are not required to make them nullable. MapperServer was updated so that the mapper is launched and the CDP client is created before the class is constructed.
- In cases where this was not possible, class fields that are not initialized in the constructor are explicitly marked nullable.
- Some Map usages were edited so that the compiler can infer whether the value from Map.get is valid.


mathiasbynens (approved):

sadym-chromium (approved):

sadym-chromium:
like it!

PR #36, opened 2021-07-28
Update typings for server and message-handling classes and functions

bwalderman:
This is a pass through the server and message-handling code to make the types as explicit as possible. Also renaming some interfaces to better describe how they are used.

- `IServer` is now `ITransport`.

- All of the methods and callbacks that handle messages are typed with either `string` or `object` instead of `any`. This makes it clear what type of message is being handled at a given place in the code.

- Uses of the `{}` type have been replaced with `object` as well, because apparently a string is acceptable as a `{}`. The `object` type is more explicit and prevents confusion between string and object parameters.

- A `IBidiServer` interface has been added, which contains the subset of BidiServer methods that external callers need to access. Related interfaces have been added to the describe various Bidi message types.

- Since BidiServer was the only remaining subclass of AbstractServer, I've merged these classes together. Now that EventEmitter is available in the mapper code, BidiServer uses EventEmitter for firing 'message' events instead.

- Found a way to simplify the typings for Sinon spies in stubTransport.spec.ts.


src/bidiMapper/utils/bidiServer.ts
```diff
@@ -14,36 +14,81 @@
...
+    this._transport.sendMessage(messageStr);
   }
 
-  close(): void {}
+  close(): void {
+    this._transport.close();
+  }
 
-  private _onBidiMessage(messageStr: string): void {
+  private _onBidiMessage = async (messageStr: string) => {
```
sadym-chromium, line 91: nit: let's stick to one declaration style

src/index.ts
```diff
@@ -80,7 +80,7 @@ function parseArguments() {
 async function _onNewBidiConnectionOpen(
   browserExecutablePath: string,
   headless: boolean,
-  bidiServer: IServer
+  bidiConnection: ITransport
```
sadym-chromium, line 83: `bidiTransport`?

src/tests/stubTransport.spec.ts
```diff
@@ -14,26 +14,30 @@
...
 
   constructor() {
-    this.sendMessage = spy() as any as ((messageObj: any) => Promise<any>) &
-      SinonSpy<any[], any>;
-    this.setOnMessage = spy() as any as ((messageObj: any) => Promise<void>) &
-      SinonSpy<any[], any>;
-    this.close = spy() as any as (() => void) & SinonSpy<any[], any>;
+    this.sendMessage = typedSpy();
+    this.setOnMessage = typedSpy();
+    this.close = typedSpy();
```
sadym-chromium, line 41: +1

sadym-chromium (approved):

PR #38, opened 2021-08-03
Browser fetcher + e2e github workflow

sadym-chromium:
In the PR I copied the browser fetcher logic from Puppeteer with small simplifications. Having browser downloaded locally simplifies working with BiDi server + allows to run e2e tests in workflows.


install/install.js
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+/**
+ * This file is part of public API.
+ *
+ * By default, the package runs this script during the installation
+ * process unless one of the env flags is provided.
+ */
+
+// Copied form Puppeteer
```
mathiasbynens, line 24: typo: "from"

Have you considered using `puppeteer.createBrowserFetcher(options)` instead of duplicating the underlying Browser Fetcher code (+ the additional `typescript-if-required` complexity) here? Not having to maintain this logic across multiple projects would be nice.

Another option is to write a simple standalone script that just fetches the latest binary.

mathiasbynens (changes requested):

install/install.js
sadym-chromium: fixed typo.

If you mean bring Puppeteer as a dependency, it seems like an overkill.

The `BrowserFetcher` is a relatively simple script injecting retrieving binary path for different platforms logic, which used both during setup and server running. From that perspective it seems to make sense to share it between `install` and `server` workflows.

mathiasbynens: > The `BrowserFetcher` is a relatively simple script injecting retrieving binary path for different platforms logic, which used both during setup and server running. From that perspective it seems to make sense to share it between `install` and `server` workflows.

We’re in agreement here — I’m not arguing against using `BrowserFetcher` in both workflows.

> If you mean bring Puppeteer as a dependency, it seems like an overkill.

Hmm, which criteria are you using to determine whether something is “overkill” or not?

This PR currently introduces 6 new `dependencies` + 3 new `devDependencies`, and duplicates the `BrowserFetcher` code from Puppeteer, which includes logic that doesn’t seem necessary here (e.g. dealing with proxy env vars). 

I’m suggesting it could instead be 1 new dependency with no code duplication/additional maintenance overhead.

The latter option sounds simpler, more maintainable, and less “overkill” to me — can you help me understand what you think is more complex about it?

sadym-chromium: let me try it (puppeteer dependency) out

foolip:
I like the approach of factoring out browser fetcher logic

foolip:
Oops, didn't mean to close and reopen :)

sadym-chromium:
Implemented in the other way in #39

PR #39, opened 2021-08-04
Download and launch Chromium using Puppeteer

sadym-chromium:
* Use Puppeteer to download on install and launch browser.
* Remove `--browser` CLI parameter.
* Disable not implemented `e2e` tests.
* Add `e2e` workflow.


src/index.ts
```diff
@@ -103,11 +93,9 @@ async function _onNewBidiConnectionOpen(
...
-}
-
-function _onBidiConnectionClosed(browserProcess: BrowserProcess): void {
-  // Client disconnected. Close browser.
-  browserProcess.closeBrowser();
+  // Return delegate to be called when the connection is closed.
+  return async () => {
+    // Client disconnected. Close browser.
+    await browser.close();
+  };
```
mathiasbynens, line 100: `return browser.close()`?

src/index.ts
```diff
@@ -55,41 +50,36 @@ function parseArguments() {
...
+ */
 async function _onNewBidiConnectionOpen(
-  browserExecutablePath: string,
   headless: boolean,
   bidiServer: IServer
-): Promise<BrowserProcess> {
-  // Launch browser.
-  const browserProcess = await launchBrowser(browserExecutablePath, headless);
+): Promise<() => void> {
+  // Hijack Puppeteer's implementation of fetching and launching browser.
```
mathiasbynens, line 72: What do you mean by “hijack”? What does “fetching” refer to here? (`launch()` doesn’t do fetching.)

If we need more control about which Chromium revision to download we could use `puppeteer.createBrowserFetcher()` in the future, but for now using the bundled-with-Puppeteer Chromium seems fine 👍 .

src/index.ts
```diff
@@ -15,12 +15,13 @@
  */
 `use strict`;
```
mathiasbynens, line 16: Unrelated to this patch, but we should fix it: this line has no effect. It should be `'use strict'` or `"use strict"`.

Compare e.g.

```
(function() { 'use strict'; return this; })();
// → undefined, i.e. strict mode is enabled

(function() { `use strict`; return this; })();
// → globalThis, strict mode is not enabled
```

mathiasbynens:
Are the changes to `tests/test_bidi.py` really related to this patch? If not, please separate them.

src/index.ts
sadym-chromium: `return browser.close`, without brackets, but it can be confusing, and can be read as `close browser and return`, while it should be `return delegate, calling which will close the browser`.

src/index.ts
sadym-chromium: I wasn't aware of that difference. Thanks!

sadym-chromium:
> Are the changes to `tests/test_bidi.py` really related to this patch? If not, please separate them.

Changes in the tests made the `e2e` git workflow possible. And `e2e` is needed to verify implemented logic is valid.

src/index.ts
```diff
@@ -55,44 +50,48 @@ function parseArguments() {
...
   headless: boolean,
   bidiServer: IServer
-): Promise<BrowserProcess> {
-  // Launch browser.
-  const browserProcess = await launchBrowser(browserExecutablePath, headless);
-  // Get BiDi Mapper script.
+): Promise<() => void> {
+  // 1. Launch Chromium (using Puppeteer for now).
+  // Puppeteer should have downloaded Chromium during the installation.
+  // Use Puppetter's logic of launching browser as well.
```
mathiasbynens, line 80: typos

src/index.ts
```diff
@@ -55,44 +50,48 @@ function parseArguments() {
...
+  const browser = await (puppeteer as any as PuppeteerNode).launch({
```
mathiasbynens, line 81: For the record, my actual suggestion was to use [`puppeteer.createBrowserFetcher()`](https://github.com/puppeteer/puppeteer/blob/main/docs/api.md#puppeteercreatebrowserfetcheroptions). This patch is a little more indirect since it just uses whatever Chromium revision Puppeteer happens to bundle. If we need/want to control which exact revision to use for chromium-bidi in the future, we could still move towards using it.

src/mapperReader.ts
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-`use strict`;
+"use strict";
```
mathiasbynens, line 16: (is Prettier happy with the double quotes? we use single quotes elsewhere)

README.md
```diff
@@ -10,42 +10,30 @@ This is a Node.js project, so install dependencies as usual:
...
 
-    npm run bidi-server -- --headless=false
+    npm run server -- --headless=false
 
 ### Starting on Linux and Mac
 
 TODO: verify if it works on Windows.
 
-You can alson run the Server by using script `./runBiDiServer.sh`. It will write output to the file `log.txt`:
+You can alson run the server by using script `./runBiDiServer.sh`. It will write output to the file `log.txt`:
```
mathiasbynens, line 34: typo: alson

mathiasbynens (approved):
LGTM with comments. Please write a descriptive commit message.

README.md
sadym-chromium: done, thanks!

src/mapperReader.ts
sadym-chromium: fixed

mathiasbynens (approved):
Thank you!

PR #42, opened 2021-09-01
Basic JavaScript value serialization

bwalderman:
This is an initial implementation of JS value serialization which will be needed to provide JS parameters with log.entryAdded events. To enable us to support deep (i.e. maxDepth > 1) serialization of objects without round-tripping, I've implemented a helper script that does the work on the page itself. This is inspired by ChromeDriver classic's design which is described in https://source.chromium.org/chromium/chromium/src/+/main:chrome/test/chromedriver/docs/run_javascript.md

The new helper script can be found in src/bidiMapper/scripts. The first time it runs, it creates a Serializer that has a map of weak references to objects. At the moment, only basic serialization is implemented. Serialization for objects (including recursion) should be fairly easy to add in follow-up changes.

Most of the rest of the change is plumbing for the PROTO.page.evaluate command to enable test coverage for the serialization script. Notable changes include fixing a race condition during mapper initialization where the BrowsingContextProcessor was missing some attachedToTarget events, and passing CdpClients to Contexts so they can be used for sending Runtime.evaluate calls.


src/bidiMapper/scripts/eval.es
```diff
@@ -0,0 +1,115 @@
...
+        }
+        default:
+          throw new Error('not yet implemented');
+      }
+    }
+  }
+
+  function uuid() {
+    // TODO: Generate UUIDs.
+    return '';
```
mathiasbynens, line 93: Since this gets evaluated in the page context, we could use the new `crypto.randomUUID()` API here — it shipped in Chrome 92 IIRC.

package.json
```diff
@@ -20,19 +22,21 @@
     "debug": "^4.3.1",
     "puppeteer": "^10.2.0",
     "ts-node": "^9.1.1",
-    "tslib": "^2.2.0",
+    "uuid": "^8.3.2",
```
mathiasbynens, line 25: I think we might be able to get by without this dependency by leveraging `crypto.randomUUID()` in the page context. See my comment below.

src/bidiMapper/scripts/eval.es
```diff
@@ -0,0 +1,115 @@
...
+        if (!id) {
+          id = uuid();
+          this._objectToId.set(value, id);
+          this._idToObject.set(id, new WeakRef(value));
+        }
+      } else {
+        throw new Error('not yet implemented');
+      }
+      return { type: 'object', objectId: uuid() };
+    }
```
sadym-chromium, line 43:
```suggestion
        return { type: 'object', objectId: id };
      }
      throw new Error('not yet implemented');
    }
```
otherwise the `objectId` is never returned

src/bidiMapper/scripts/eval.es
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  function uuid() {
+    // TODO: Generate UUIDs.
+    return '';
+  }
+
+  return function evaluate(script, args) {
+    if (window.__webdriver_js_serializer === undefined) {
```
sadym-chromium, line 97: What if some malicious page change the serialisation function? They could mimic the eval results to anything and make BiDi working unpredictably/unstable on such pages.

sadym-chromium (commented):
Couple of concerns. One minor, and another is about keeping the serialisation logic in the client's window object.

bwalderman:
@googlebot I consent.

src/bidiMapper/scripts/eval.es
bwalderman: Agreed this is not ideal. Leaving it as-is for now per our meeting discussions and can look at more robust solutions such as updating CDP in the future. This will at least provide us will spec-compliant values for well-behaved pages which will be useful for testing.

sadym-chromium:
@googlebot I consent.

sadym-chromium (approved):

sadym-chromium:
LGTM

src/bidiMapper/domains/context/context.ts
```diff
@@ -29,4 +45,45 @@ export class Context {
...
+        expression,
+        returnByValue: true,
+      }
+    );
+
+    // TODO: Should we let exceptions thrown from the user script bubble up to here,
+    // or catch them and send an { error: .. } value?
+    if (exceptionDetails) {
+      throw new Error(exceptionDetails.text);
+    }
```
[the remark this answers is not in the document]
bwalderman: I've updated the code to return protocol success with an `exceptionDetails` property when the user script throws. This is what https://github.com/w3c/webdriver-bidi/pull/136 currently does.

PR #44, opened 2021-09-23
Initial implementation of `script.evaluate`

sadym-chromium:
This approach splits `script.evaluate` into 2 CDP calls:
1. Evaluate script and get a `CDP objectId` as a result;
2. Serialise the result `CDP object` in the page context, as implemented in #42.

The ratio behind it is to keep stack-trace of the execution, and at the same time don't make multiple, but only 1 extra request for the object property-by-property serialisation.


bwalderman (changes requested):

sadym-chromium:
Abandon, as rebasing wouldn't make sense after merging #48

PR #48, opened 2021-09-28
unit tests on `eval.es`

sadym-chromium:
* Adding unit tests for serialization/deserialization logic.

Referencing `eval.es` from `eval.spec.ts` wouldn't work, because `eval.es` is excluded from the compilation. So the test reads the script code and eval it, as Mapper will do in the page context.

### DIFF
--- a/src/bidiMapper/scripts/eval.es
+++ b/src/bidiMapper/scripts/eval.es
@@ -2,7 +2,7 @@
  * Copyright 2021 Google LLC.
  * Copyright (c) Microsoft Corporation.
  *
- * Licensed under the Apache License, Version 2.0 (the "License");
+ * Licensed under the Apache License, Version 2.0 (the 'License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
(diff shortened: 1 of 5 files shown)


src/bidiMapper/scripts/eval.es
```diff
@@ -16,117 +16,147 @@
...
       }
+      default:
+        throw new Error('not yet implemented');
+
     }
   }
 
   function uuid() {
-    return crypto.randomUUID();
+    // TODO sadym: `crypto.randomUUID()` works only in secure context.
```
bwalderman, line 134: I tried configuring rollup and typescript to build this file so that the `uuid` node module could be imported. I ran into various problems trying to get this to work. Rollup outputs an IIFE with a semicolon at the end, and the semicolon causes a syntax error when wrapping the script text. This might be worth another look if we can't use crypto.randomUUID.

src/bidiMapper/scripts/eval.es
```diff
@@ -16,117 +16,147 @@
  */
 
 (() => {
-  class Serializer {
+  class SerializationMapper {
```
bwalderman, line 19:
```suggestion
  class ObjectCache {
```

src/bidiMapper/scripts/eval.es
```diff
@@ -2,7 +2,7 @@
  * Copyright 2021 Google LLC.
  * Copyright (c) Microsoft Corporation.
  *
- * Licensed under the Apache License, Version 2.0 (the "License");
+ * Licensed under the Apache License, Version 2.0 (the 'License");
```
bwalderman, line 5:
```suggestion
 * Licensed under the Apache License, Version 2.0 (the "License");
```

src/bidiMapper/scripts/eval.es
```diff
@@ -16,117 +16,147 @@
...
-        // TODO: Recursive serialize.
-        let objectId = this._objectToId.get(value);
-        if (!objectId) {
-          objectId = uuid();
-          this._objectToId.set(value, objectId);
-          this._idToObject.set(objectId, new WeakRef(value));
-        }
-        return { type: 'object', objectId };
+  function getSerializationMapper() {
+    // No `window` can be in case of unit tests.
```
bwalderman, line 25: Can we use `globalThis` instead of `window`?

src/bidiMapper/scripts/eval.spec.ts
```diff
@@ -0,0 +1,162 @@
+import fs from 'fs/promises';
```
bwalderman, line 1: This needs the copyright text.

src/bidiMapper/scripts/eval.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+import path from 'path';
+import * as chai from 'chai';
+import chaiExclude from 'chai-exclude';
+
+chai.use(chaiExclude);
+
+describe('Evaluator', function () {
+    let EVALUATOR: any;
```
bwalderman, line 9: I'd prefer to avoid using `any` when the type is actually known. Could you define a small interface type with the evaluate/serialize/deserialize methods?

src/bidiMapper/scripts/eval.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+    before(async function () {
+        const eval_text = await fs.readFile(
+            path.join(__dirname, './eval.es'),
+            'utf8'
+        );
+
+        EVALUATOR = eval(eval_text);
+    });
+
+    function checkSerializeAndDeserialize(originalObject: any, expectedSerializedObj: any, excluding: string[] = []) {
```
bwalderman, line 21: Consider using `unknown` instead of `any` where possible. I think it will work here. It's similar to `any` but forces you to type check the variable before you can operate on it: https://dmitripavlutin.com/typescript-unknown-vs-any/

bwalderman, line 21: Minor: Once #44 is merged and `RemoteValue` is fully defined, we can use it for `expectedSerializedObj`.

bwalderman (changes requested):

src/bidiMapper/scripts/eval.es
sadym-chromium: I created a ticket for that: #49

src/bidiMapper/scripts/eval.es
sadym-chromium: done

src/bidiMapper/scripts/eval.spec.ts
sadym-chromium: done

src/bidiMapper/scripts/eval.spec.ts
sadym-chromium: Good suggestion. I tried it out, but definition of `deepEqualExcludingEvery` uses `keyof T` as an ignored properties type. Using `unknown` makes properties of type `never[]`, which forces to use a cast in the call `excluding as any as never[]`, which looks a bit ugly for my taste.
```
 deepEqualExcludingEvery<T>(actual: T | T[], expected: T | T[], props: keyof T | (keyof T)[], message?: string): void;
```

src/bidiMapper/scripts/eval.spec.ts
sadym-chromium: done

src/bidiMapper/scripts/eval.spec.ts
sadym-chromium: Done in #50

sadym-chromium:
@googlebot I fixed it.

src/bidiMapper/scripts/eval.spec.ts
bwalderman: Sounds good, thanks for giving this a try.

bwalderman (approved):

PR #51, opened 2021-09-29
Split `script.evaluate` into `eval` and `serialize`

sadym-chromium:
To provide a stacktrace in case of not-`Error` thrown, the exception should not be caught in the evaluated script itself.

Another advantage of splitting `eval` and `serialize` is there is no need in parsing text stacktrace from JS `Error` object, and no need in correction stacktrace positions because of the script wrapping.

### DIFF
--- a/src/bidiMapper/domains/context/context.ts
+++ b/src/bidiMapper/domains/context/context.ts
@@ -96,8 +96,38 @@ export class Context {
     return response.result.value;
   }
 
+  private async _serializeCdpExceptionDetails(
+    cdpExceptionDetails: Protocol.Runtime.ExceptionDetails
+  ) {
+    const callFrames = cdpExceptionDetails.stackTrace?.callFrames.map(
+      (frame) => ({
+        url: frame.url,
+        functionName: frame.functionName,
+        lineNumber: frame.lineNumber,
+        columnNumber: frame.columnNumber,
+      })
+    );
+    const exception = await this._serializeCdpObject(
+      // Exception should always be there.
+      cdpExceptionDetails.exception!
+    );
+
+    return {
+      exceptionDetails: {
+        exception,
+        columnNumber: cdpExceptionDetails.columnNumber,
+        lineNumber: cdpExceptionDetails.lineNumber,
+        stackTrace: {
+          callFrames: callFrames || [],
+        },
+        text: cdpExceptionDetails.text,
+      },
+    };
+  }
+
   public async evaluateScript(
-    script: string
+    script: string,
+    awaitPromise: boolean
   ): Promise<Script.ScriptEvaluateResult> {
     // Evaluate works with 2 DP calls:
     // 1. Evaluates the script;
@@ -107,35 +137,14 @@ export class Context {
     const expression = script;
     const cdpEvaluateResult = await this._cdpClient.Runtime.evaluate({
       expression,
+      awaitPromise,
       returnByValue: false,
     });
-    // Serialize exception details.
     if (cdpEvaluateResult.exceptionDetails) {
-      const callFrames =
-        cdpEvaluateResult.exceptionDetails.stackTrace?.callFrames.map(
-          (frame) => ({
-            url: frame.url,
-            functionName: frame.functionName,
-            lineNumber: frame.lineNumber,
-            columnNumber: frame.columnNumber,
-          })
-        );
-      const exception = await this._serializeCdpObject(
-        // Exception should always be there.
-        cdpEvaluateResult.exceptionDetails.exception!
+      // Serialize exception details.
+      return await this._serializeCdpExceptionDetails(
+        cdpEvaluateResult.exceptionDetails
       );
-
-      return {
-        exceptionDetails: {
-          exception,
-          columnNumber: cdpEvaluateResult.exceptionDetails.columnNumber,
-          lineNumber: cdpEvaluateResult.exceptionDetails.lineNumber,
-          stackTrace: {
-            callFrames: callFrames || [],
-          },
-          text: cdpEvaluateResult.exceptionDetails.text,
-        },
-      };
     }
 
     return {
--- a/src/bidiMapper/domains/context/context.ts
+++ b/src/bidiMapper/domains/context/context.ts
@@ -25,6 +25,8 @@ export class Context {
   _targetInfo?: Protocol.Target.TargetInfo;
   _sessionId?: string;
 
+  private _dummyContextObjectId: string = '';
+
   private constructor(
     private _contextId: string,
     private _cdpClient: CdpClient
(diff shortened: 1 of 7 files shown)


sadym-chromium:
@googlebot I fixed it.

bwalderman:
I like the approach of splitting eval and serialize. I think having a separate serialize function will be useful later when implementing log.entryAdded. We can use it to convert CDP RemoteObjects from Runtime.consoleAPICalled into serialized Bidi values.

I do have a suggestion. If I understand correctly, this PR currently does the following when script.evaluate is called:
1. Call CDP Runtime.evaluate with `script` and `returnByValue`=false
2. Receive a CDP RemoteObject representing the real (not serialized) result value, or an exceptionDetails.
3. Pass the result or exception RemoteObject to the eval.es `serialize` function to obtain a serialized Bidi value.

This lets us have deeply-nested result values and stacktraces even for non-Error types, but there is a race condition between steps 2 and 3. The result RemoteObject from step 2 is a handle to a live JS object in the page, so the page might change it before step 3 can be completed.

My suggestion is to do the following instead:
First, ensure the eval.es `evaluate` function calls `serialize` on its result. Then,
1. Use Runtime.evaluate to call eval.es `evaluate` with `script` and `returnByValue`=false.
2. Because we set `returnByValue`=false above, we receive a CDP RemoteObject representing the *serialized* result value, or an exceptionDetails.
3. If we got an exceptionDetails, pass exception RemoteObject to the eval.es `serialize` function to obtain a serialized Bidi value.
4. If we got a result RemoteObject, pass it to Runtime.callFunctionOn with functionDeclaration `(val) => val` and `returnByValue`=true to unwrap it into JSON.

The difference is that in step 2, we receive a handle to a serialized snapshot (which is immutable), instead of a handle to a live object.

bwalderman (changes requested):

src/bidiMapper/domains/context/context.ts
```diff
@@ -64,41 +64,91 @@ export class Context {
...
-    }
+  /**
+   * Serializes a given CDP object into BiDi, keeping references in the
+   * target's `globalThis`.
+   * @param cdpObject CDP remote object to be serialized.
+   */
+  private async _serializeCdpObject(
+    cdpObject: Protocol.Runtime.RemoteObject
+  ): Promise<CommonDataTypes.RemoteValue> {
+    // TODO sadym: `dummyContextObject` needed for the running context.
```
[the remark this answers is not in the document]
sadym-chromium: CDP requires "Either ObjectId or executionContextId must be specified".

src/bidiMapper/domains/context/context.ts
```diff
@@ -64,41 +64,91 @@ export class Context {
...
+    // Evaluate works with 2 DP calls:
+    // 1. Evaluates the script;
+    // 2. serializes the result or exception.
+    // This needed to provide a detailed stacktrace in case of not `Error` but
+    // anything else wihtout`stacktrace` is thrown. To provide the stacktrace,
+    // CDP domains `Debugger` and`Runtime` should be enabled.
+    await this._cdpClient.Debugger.enable({});
+    await this._cdpClient.Runtime.enable();
+
+    const expression = script;
```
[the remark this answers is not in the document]
sadym-chromium: fixed

sadym-chromium:
> I like the approach of splitting eval and serialize. I think having a separate serialize function will be useful later when implementing log.entryAdded. We can use it to convert CDP RemoteObjects from Runtime.consoleAPICalled into serialized Bidi values.
> 
> I do have a suggestion. If I understand correctly, this PR currently does the following when script.evaluate is called:
> 
> 1. Call CDP Runtime.evaluate with `script` and `returnByValue`=false
> 2. Receive a CDP RemoteObject representing the real (not serialized) result value, or an exceptionDetails.
> 3. Pass the result or exception RemoteObject to the eval.es `serialize` function to obtain a serialized Bidi value.
> 
> This lets us have deeply-nested result values and stacktraces even for non-Error types, but there is a race condition between steps 2 and 3. The result RemoteObject from step 2 is a handle to a live JS object in the page, so the page might change it before step 3 can be completed.
> 
> My suggestion is to do the following instead: First, ensure the eval.es `evaluate` function calls `serialize` on its result. Then,
> 
> 1. Use Runtime.evaluate to call eval.es `evaluate` with `script` and `returnByValue`=false.
> 2. Because we set `returnByValue`=false above, we receive a CDP RemoteObject representing the _serialized_ result value, or an exceptionDetails.
> 3. If we got an exceptionDetails, pass exception RemoteObject to the eval.es `serialize` function to obtain a serialized Bidi value.
> 4. If we got a result RemoteObject, pass it to Runtime.callFunctionOn with functionDeclaration `(val) => val` and `returnByValue`=true to unwrap it into JSON.
> 
> The difference is that in step 2, we receive a handle to a serialized snapshot (which is immutable), instead of a handle to a live object.

I impleneted the suggestion in the next PR https://github.com/GoogleChromeLabs/chromium-bidi/pull/53

src/bidiMapper/scripts/eval.es
```diff
@@ -153,19 +159,11 @@
   }
 
   function evaluate(script, args) {
-    try {
       const deserializedArgs = args.map((arg) => deserialize(arg));
-      const func = new Function(`return (${script})`);
+      const func = new Function(`return (\n${script}\n)`);
```
foolip, line 163: I was wondering why the added newlines here. @sadym-chromium says the answer is to preserve the column number. There's also an offset (in another PR) to adjust the numbers. In any case it's here for a reason, not accidental.

tests/test_bidi.py
```diff
@@ -26,6 +26,12 @@ async def websocket():
     async with websockets.connect(url) as connection:
         yield connection
 
+@pytest.fixture(autouse=True)
+async def before_each_test(websocket):
```
foolip, line 30: This was needed because `_getOrCreateContext` was made async, and an event is always emitted. This will probably change again when event filtering is added.

tests/test_bidi.py
```diff
@@ -745,6 +751,105 @@ async def _ignore_test_consoleError_logEntryWithMethodErrorEmmited(websocket):
...
+            "awaitPromise": False,
+            "target": {"context": contextID}}})
+
+    # Assert command done.
+    resp = await read_JSON_message(websocket)
+    assert resp["id"] == 46
+
+    # Compare ignoring `objectId`.
+    recursiveCompare({
+            "type":"promise",
```
foolip, line 826: Right now there's no way to do anything with this promise in a subsequent BiDi command, but that's coming. It needs a way to pass arguments to script evaluation.

foolip (approved):

PR #53, opened 2021-10-01
Evaluate and serialize in the same call.

sadym-chromium:
I make this in separate PR to keep previous PR reviewable.

In this PR the [suggestion to avoid race condition](https://github.com/GoogleChromeLabs/chromium-bidi/pull/51#issuecomment-931737752) is implemented. Evaluation and serialization are done in the same call, and serialization results retrieved by another call.

`test_scriptEvaluateChangingObject_resultObjectDidNotChange` fails on the #51 but passes in this PR.

### DIFF
--- a/src/bidiMapper/domains/context/context.ts
+++ b/src/bidiMapper/domains/context/context.ts
@@ -102,28 +102,6 @@ export class Context {
 
     return response.result.value;
   }
-  /**
-   * Get value of CDP object.
-   * @param cdpObject CDP remote object to be returned.
-   */
-  private async _getCdpObjectValue(
-    cdpObject: Protocol.Runtime.RemoteObject
-  ): Promise<CommonDataTypes.RemoteValue> {
-    const cdpValueResult = await this._cdpClient.Runtime.callFunctionOn({
-      functionDeclaration: `(a)=>{return a;}`,
-      objectId: this._dummyContextObjectId,
-      arguments: [cdpObject],
-      returnByValue: true,
-    });
-
-    if (cdpValueResult.exceptionDetails) {
-      throw new Error(
-        'Cannot get CDP value: ' + cdpValueResult.exceptionDetails!.text
-      );
-    }
-
-    return cdpValueResult.result.value;
-  }
 
   private async _serializeCdpExceptionDetails(
     cdpExceptionDetails: Protocol.Runtime.ExceptionDetails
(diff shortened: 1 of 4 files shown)


tests/test_bidi.py
```diff
@@ -850,6 +850,34 @@ async def test_scriptEvaluateWaitPromise_resultReturned(websocket):
...
+@pytest.mark.asyncio
+async def test_scriptEvaluateChangingObject_resultObjectDidNotChange(websocket):
+    contextID = await get_open_context_id(websocket)
+
+    # Send command.
+    await send_JSON_command(websocket, {
+        "id": 47,
+        "method": "script.evaluate",
+        "params": {
+            "expression": "(()=>{const a={i:0};const f=()=>{ setTimeout(()=>{ a.i++; f(); }, 0); };f();return a;})()",
```
foolip, line 862: A comment explaining why the timing of this is guaranteed to be correct would be great. It has to be impossible for any subsequent CDP command to be handled before the `setTimeout` callback runs for it to be valid.

When there's an invoke method or something to serialize the value (given by objectId) again we could also check that the `value` becomes non-zero directly afterwards. (Nice-to-have.)

src/bidiMapper/domains/context/context.ts
```diff
@@ -26,6 +26,9 @@ export class Context {
   _sessionId?: string;
 
   private _dummyContextObjectId: string = '';
+  // As `script.evaluate` wraps call into serialisation script, `lineNumber`
+  // should be adjusted.
+  private _evaluateStacktraceLineOffset = 1;
```
foolip, line 31: This is compensating for the added newline in the previous PR.

src/bidiMapper/domains/context/context.ts
```diff
@@ -133,14 +163,29 @@ export class Context {
...
+        return (${EVALUATOR_SCRIPT})
+          .serialize.apply(null, [${awaitPromise ? 'await' : ''} value])
+      }`;
+
     const cdpEvaluateResult = await this._cdpClient.Runtime.evaluate({
-      expression,
-      awaitPromise,
+      expression: evalAndSerialiseScript,
+      // Always wait for the result of `_serialize`. Wait or not for the user's
+      // expression is handled in the`_serialize` function.
```
foolip, line 187:
```suggestion
      // expression is handled in the `_serialize` function.
```

foolip (changes requested):

src/bidiMapper/domains/context/context.ts
```diff
@@ -133,14 +163,29 @@ export class Context {
...
-    // 1. Evaluates the `script`;
-    // 2. serializes the result or exception.
+    // Evaluate works with 2 CDP calls:
+    // 1. Evaluates the `script` + serializes the result into CDP object.
+    // 2. Retrieves the serialization result.
     // This needed to provide a detailed stacktrace in case of not `Error` but
-    // anything else wihtout`stacktrace` is thrown.
+    // anything else wihtout a `stacktrace` is thrown. And in the same time to
+    // avoid a race condition in case of the result object is changed between
+    // those 2 CDP calls.
```
sadym-chromium, line 172: Actually, this can be simplified and made in 1 call.

sadym-chromium: Which is done

PR #55, opened 2021-10-05
`create` and `navigate.browsingContext`

sadym-chromium:
* Implement [`browsingContext.navigate`](https://w3c.github.io/webdriver-bidi/#command-browsingContext-navigate) command.
* Add `browsingContext.create` -> `params.type`.
* Add unit test for different event orders in `browsingContext.create`.
* Made `EVALUATOR_SCRIPT` a property to allow unit testing.
* `BrowsingContextProcessor` is responsible for handling events for testability.

### DIFF
--- a/src/bidiMapper/bidiProtocolTypes.ts
+++ b/src/bidiMapper/bidiProtocolTypes.ts
@@ -95,6 +95,7 @@ export namespace Script {
 // https://w3c.github.io/webdriver-bidi/#module-browsingContext
 export namespace BrowsingContext {
   export type BrowsingContext = string;
+  export type Navigation = string;
 
   export type BrowsingContextGetTreeCommand = {
     method: 'browsingContext.getTree';
@@ -128,10 +129,29 @@ export namespace BrowsingContext {
   export type BrowsingContextCreateType = 'tab' | 'window';
 
   export type BrowsingContextCreateParameters = {
-    type: BrowsingContextCreateType;
+    type?: BrowsingContextCreateType;
   };
+
   export type BrowsingContextCreateResult = {
     context: BrowsingContext;
+  };
+
+  export type BrowsingContextNavigateCommand = {
+    method: 'browsingContext.navigate';
+    params: BrowsingContextNavigateParameters;
+  };
+
+  export type BrowsingContextNavigateParameters = {
+    context: BrowsingContext;
+    url: string;
+    wait?: ReadinessState;
+  };
+
+  export type ReadinessState = 'none';
+  // TODO sadym: implement 'interactive' and 'complete' states.
+  export type BrowsingContextNavigateResult = {
+    navigation?: Navigation;
+    url: string;
   };
 }
 
--- a/src/bidiMapper/bidiProtocolTypes.ts
+++ b/src/bidiMapper/bidiProtocolTypes.ts
@@ -136,23 +136,25 @@ export namespace BrowsingContext {
     context: BrowsingContext;
   };
 
-  export type BrowsingContextNavigateCommand = {
-    method: 'browsingContext.navigate';
-    params: BrowsingContextNavigateParameters;
-  };
+  export namespace PROTO {
+    export type BrowsingContextNavigateCommand = {
+      method: 'PROTO.browsingContext.navigate';
+      params: BrowsingContextNavigateParameters;
+    };
 
-  export type BrowsingContextNavigateParameters = {
-    context: BrowsingContext;
-    url: string;
-    wait?: ReadinessState;
-  };
+    export type BrowsingContextNavigateParameters = {
+      context: BrowsingContext;
+      url: string;
+      wait?: ReadinessState;
+    };
 
-  export type ReadinessState = 'none';
-  // TODO sadym: implement 'interactive' and 'complete' states.
-  export type BrowsingContextNavigateResult = {
-    navigation?: Navigation;
-    url: string;
-  };
+    export type ReadinessState = 'none';
+    // TODO sadym: implement 'interactive' and 'complete' states.
+    export type BrowsingContextNavigateResult = {
+      navigation?: Navigation;
+      url: string;
+    };
+  }
 }
 
 export namespace Session {
--- a/src/bidiMapper/bidiProtocolTypes.ts
+++ b/src/bidiMapper/bidiProtocolTypes.ts
@@ -120,22 +120,6 @@ export namespace BrowsingContext {
     children: BrowsingContextInfoList;
   };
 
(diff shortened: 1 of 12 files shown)


src/bidiMapper/bidiProtocolTypes.ts
```diff
@@ -128,11 +129,32 @@ export namespace BrowsingContext {
...
+    type?: BrowsingContextCreateType;
   };
+
   export type BrowsingContextCreateResult = {
     context: BrowsingContext;
   };
+
+  export namespace PROTO {
+    export type BrowsingContextNavigateCommand = {
+      method: 'PROTO.browsingContext.navigate';
```
foolip, line 141: Since https://w3c.github.io/webdriver-bidi/#command-browsingContext-navigate exists now, does this need to be under PROTO?

src/bidiMapper/domains/context/browsingContextProcessor.spec.ts
```diff
@@ -0,0 +1,152 @@
...
+    cdpConnection = new CdpConnection(mockCdpServer);
+    browsingContextProcessor = new BrowsingContextProcessor(
+      cdpConnection,
+      'SELF_TARGET_ID',
+      sinon.fake as unknown as (t: Context) => Promise<void>,
+      sinon.fake as unknown as (t: Context) => Promise<void>,
+      EVALUATOR_SCRIPT
+    );
+
+    // Actual `Context.create` logic involves several CDP calls, so mock it to avoid all the simulations.
```
foolip, line 66: This is a helpful comment, thanks!

src/bidiMapper/domains/context/browsingContextProcessor.ts
```diff
@@ -38,38 +38,63 @@ export class BrowsingContextProcessor {
...
+    browserCdpClient.Target.on('targetInfoChanged', (params) => {
+      this._handleInfoChangedEvent(params);
+    });
+    browserCdpClient.Target.on('detachedFromTarget', (params) => {
+      this._handleDetachedFromTargetEvent(params);
+    });
+  }
+
+  // Creation of `Context` can take quite a while. To avoid race condition, in
+  // the map kept Pmomise, eventualy resolved with the`Context`.
```
foolip, line 65:
```suggestion
  // Creation of `Context` can take quite a while. To avoid race condition, keep
  // a Promise in the map, eventually resolved with the `Context`.
```

src/bidiMapper/bidiProtocolTypes.ts
sadym-chromium: Actually I meant `PROTO.browsingContext.create`.

src/bidiMapper/domains/context/browsingContextProcessor.spec.ts
```diff
@@ -0,0 +1,152 @@
...
+    },
+  };
+
+  beforeEach(async function () {
+    mockCdpServer = new StubTransport();
+    cdpConnection = new CdpConnection(mockCdpServer);
+    browsingContextProcessor = new BrowsingContextProcessor(
+      cdpConnection,
+      'SELF_TARGET_ID',
+      sinon.fake as unknown as (t: Context) => Promise<void>,
```
bwalderman, line 61: Can you wrap the `sinon.fake as unknown as (t: Context) => Promise<void>` code into a `typedFake` function similar to `typedSpy`?

bwalderman: Also, shouldn't this call `fake` (e.g. `fake()`)?

src/bidiMapper/domains/context/browsingContextProcessor.ts
```diff
@@ -82,37 +107,40 @@ export class BrowsingContextProcessor {
...
+  //   "params": {
+  //     "sessionId": "7EFBFB2A4942A8989B3EADC561BC46E9",
+  //     "targetId": "19416886405CBA4E03DBB59FA67FF4E8" } }
+  private async _handleDetachedFromTargetEvent(
     params: Protocol.Target.DetachedFromTargetEvent
   ) {
     logContext('detachedFromTarget event received', params);
 
-    // TODO: params.targetId is deprecated. Update this class to track using params.sessionId instead.
+    // TODO: params.targetId is deprecated. Update this class to track using
```
bwalderman, line 140: Filed #60 to track this.

src/bidiMapper/domains/context/browsingContextProcessor.ts
```diff
@@ -122,45 +150,62 @@ export class BrowsingContextProcessor {
...
+      const result = await browserCdpClient.Target.createTarget({
+        url: 'about:blank',
+        newWindow: params.type === 'window',
+      });
+
+      const targetId = result.targetId;
+
+      const existingContext = await this._tryGetContext(targetId);
+      if (existingContext) {
+        resolve(existingContext.toBidi());
```
bwalderman, line 170: I think awaiting here might introduce a potential race condition. Once Target.createTarget returns, we need to register the onAttachToTarget handler synchronously, or there's a chance the event will come and we'll miss it.

src/bidiMapper/domains/context/context.ts
```diff
@@ -80,6 +87,23 @@ export class Context {
     };
   }
 
+  public async navigate(
+    url: string,
+    wait: BrowsingContext.ReadinessState = 'none'
```
bwalderman, line 92: A default parameter makes sense at the API level, but for clarity, I'd prefer to avoid them in internal/implementation classes.

sadym-chromium: done

src/bidiMapper/domains/context/browsingContextProcessor.ts
sadym-chromium: Agree. Added synchronous check, if the context is already known, in which case `attachedToTarget` is not needed.

src/bidiMapper/domains/context/browsingContextProcessor.ts
sadym-chromium: Added a comment

src/bidiMapper/domains/context/browsingContextProcessor.ts
bwalderman: Thank you, that makes more sense. I think this is safe then. Even though `await this._tryGetContext(targetId)` will yield to the runtime, any `attachedToTarget` event should be handled by `_handleAttachedToTargetEvent` and the context will be put in the map before `_tryGetContext` actually executes.

bwalderman (approved):

PR #58, opened 2021-10-07
Typification

sadym-chromium:
Commands typification.


bwalderman (approved):
LGTM. Just a question (not blocking sign-off). Is `type` preferred instead of `interface` to make everything more uniform because some things must be `type`?

PR #62, opened 2021-11-17
Add `this` parameter to `callFunction` command.

sadym-chromium:
* Added `this` parameter to the command `script.callFunction`.
* Specified types (Primitive, Remote, Local).
* Added e2e tests to async functions.


foolip:
Does this implement things that will be added to https://github.com/w3c/webdriver-bidi/pull/142? The proposed name there was `thisObject`.

sadym-chromium:
I added parameter `this` in the spec and use it here as well. If you think `thisObject` is a better name, could you please comment in [w3c/webdriver-bidi#142 ](https://github.com/w3c/webdriver-bidi/pull/142) ?

foolip:
I see, I thought the difference was a mistake. I'm fine with `this`.

bwalderman (approved):

Issue #65, opened 2021-11-19
Move deserialization into CDP


sadym-chromium:
After #64 is implemented, there is no reason to keep the deserialization on [the Mapper level](https://github.com/GoogleChromeLabs/chromium-bidi/blob/e2d92766a227b1e2256383a8cb33173e6c6bbce8/src/bidiMapper/domains/context/context.ts#L255).

CDP method [`Runtime.callFunctionOn`](https://chromedevtools.github.io/devtools-protocol/tot/Runtime/#method-callFunctionOn) accepts [`Runtime.CallArgument`](https://chromedevtools.github.io/devtools-protocol/tot/Runtime/#type-CallArgument) as an argument, which can be extended with a field `bidiValue`. 

This will reduce the overhead and will make deserialization and serialization unified.

sadym-chromium:
Done in #123

Issue #52, opened 2021-10-01
Store `executionContextId` in each context


sadym-chromium:
Each time [`Runtime.executionContextCreated`](https://chromedevtools.github.io/devtools-protocol/tot/Runtime/#event-executionContextCreated) event raised, [`Context`](https://github.com/GoogleChromeLabs/chromium-bidi/blob/main/src/bidiMapper/domains/context/context.ts) should keep the `executionContextId` and use it to run [`Runtime. callFunctionOn ](https://chromedevtools.github.io/devtools-protocol/tot/Runtime/#method-callFunctionOn).

sadym-chromium:
Implemented by https://github.com/GoogleChromeLabs/chromium-bidi/blob/09d0239f60bbba2e263f1ece6218b6e521e4f87f/src/bidiMapper/domains/script/realm.ts

PR #68, opened 2021-11-19
Example scenario

sadym-chromium:
Example scenario based on https://github.com/puppeteer/puppeteer/blob/main/examples/cross-browser.js


tests/example.py
```diff
@@ -0,0 +1,143 @@
...
+    # const page = await browser.newPage();
+
+    # Part 1. Execute BiDi command.
+    # ... page.goto('https://news.ycombinator.com/');
+    # https://github.com/puppeteer/puppeteer/blob/4c3caaa3f99f0c31333a749ec50f56180507a374/examples/cross-browser.js#L34
+    command = {
+        "id": 1001,
+        "method": "browsingContext.navigate",
+        "params": {
+            "url": "https://news.ycombinator.com/",
```
foolip, line 87: Note that hitting https://news.ycombinator.com/ in the test will likely make the test flaky, as it probably doesn't have 100% uptime.

sadym-chromium: it's not a test, but an example script. Probably should be moved to another folder.

sadym-chromium: which I just did

foolip: Oh :)

src/bidiMapper/domains/context/context.ts
```diff
@@ -28,7 +28,6 @@ export class Context {
   _targetInfo?: Protocol.Target.TargetInfo;
   _sessionId?: string;
 
-  private _dummyContextObjectId: string = '';
```
sadym-chromium, line 31: The was a problem: after navigation the world is recreated, and the `dummyContextObject` was destroyed. So it should be either created after each navigation or on each evaluation call. Anyway, it should be removed as part of #52 and #65

examples/cross-browser.py
```diff
@@ -0,0 +1,143 @@
...
+    return json.loads(await websocket.recv())
+
+async def main():
+    # const browser = await puppeteer.launch();
+    # https://github.com/puppeteer/puppeteer/blob/4c3caaa3f99f0c31333a749ec50f56180507a374/examples/cross-browser.js#L29
+    websocket = await get_websocket()
+
+    # Receive initial `browsingContext.contextCreated` event.
+    resp = await read_JSON_message(websocket)
+    assert resp['method'] == 'browsingContext.contextCreated'
```
foolip, line 44: Leaving the asserts makes this look a lot like a test, and not like what a real client might do. What do you think about adding a simple helper that waits for the "browsingContext.contextCreated" event but doesn't fail if some other event comes first?

Similarly for commands, waiting for the command response and not failing if there's another message that comes first.

I guess what this has to be is one bit of code that reads messages and dispatches, and it probably has a list of `asyncio.Future` objects, calling `set_result` or `set_exception`.

sadym-chromium: Agree. Removed assertions and extracted waiting for a command logic to a dedicated function.

examples/cross-browser.py
```diff
@@ -0,0 +1,152 @@
...
+
+async def send_JSON_command(command, websocket):
+    await websocket.send(json.dumps(command))
+
+
+async def read_JSON_message(websocket):
+    return json.loads(await websocket.recv())
+
+
+async def run_and_wait_command(command, websocket):
```
foolip, line 39: Since it's async it's sort of implied that it will wait for something. I suggest simplifying:

```suggestion
async def run_command(command, websocket):
```

examples/cross-browser.py
```diff
@@ -0,0 +1,152 @@
...
+    command_id = command["id"]
+    await send_JSON_command(command, websocket)
+
+    # Read messages until the sent command is done.
+    while True:
+        message = await read_JSON_message(websocket)
+        if "id" in message and message["id"] == command_id:
```
foolip, line 46: Can you detect errors here, and throw if we get one?

examples/cross-browser.py
```diff
@@ -0,0 +1,152 @@
...
+    # https://github.com/puppeteer/puppeteer/blob/4c3caaa3f99f0c31333a749ec50f56180507a374/examples/cross-browser.js#L32
+
+    # Puppeteer:
+    # const page = await browser.newPage();
+
+    # Part 1. Execute BiDi command and wait for result.
+    # ... await browser.newPage();
+    # https://github.com/puppeteer/puppeteer/blob/4c3caaa3f99f0c31333a749ec50f56180507a374/examples/cross-browser.js#L31
+    command_result = await run_and_wait_command({
+        "id": 1000,
```
foolip, line 66: Can you hide the ID inside of `run_command`? The code shouldn't have to pick an ID I think.

examples/cross-browser.py
```diff
@@ -0,0 +1,152 @@
...
+        "method": "PROTO.script.callFunction",
+        "params": {
+            "functionDeclaration": """(resultsSelector) => {
+                const anchors = Array.from(document.querySelectorAll(resultsSelector));
+                return anchors.map((anchor) => {
+                    const title = anchor.textContent.trim();
+                    return `${title} - ${anchor.href}`;
+                });
+            }""",
+            "args": [results_selector],
```
foolip, line 118: It would be good if `run_and_wait_command` (or another helper) takes care of turning arguments into the right shape. Then, this could be simpler:

```suggestion
            "args": [".titlelink"],
```

Alternatively, ".titlelink" can be baked right into the function declaration since it's a constant. Maybe another example can have a more realistic use case for arguments?

examples/cross-browser.py
```diff
@@ -0,0 +1,152 @@
...
+
+async def run_and_wait_command(command, websocket):
+    command_id = command["id"]
+    await send_JSON_command(command, websocket)
+
+    # Read messages until the sent command is done.
+    while True:
+        message = await read_JSON_message(websocket)
+        if "id" in message and message["id"] == command_id:
+            return message
```
foolip, line 47: This should simplify call sites:

```suggestion
            return message["result"]
```

examples/cross-browser.py
```diff
@@ -0,0 +1,152 @@
...
+                    return `${title} - ${anchor.href}`;
+                });
+            }""",
+            "args": [results_selector],
+            "target": {"context": context_id}}}, websocket)
+
+    # Part 2. Get the command result.
+    # const links = ...
+    # https://github.com/puppeteer/puppeteer/blob/4c3caaa3f99f0c31333a749ec50f56180507a374/examples/cross-browser.js#L38
+    # The `command_result` should be like this:
```
foolip, line 124: How about a simple deserialization helper that does at least the minimum for this example, so that this comment isn't needed, and the code looks simpler?

examples/cross-browser.py
sadym-chromium: `run_command_and_wait_it_is_complite` looks quite massive. I meant we not only send a command (which will be async), but wait for the command to be finished.

examples/cross-browser.py
sadym-chromium: The script is not intended to implement all the protocol with events and error handling. It is to show the low-level interactions on the protocol level. For this reason I don't want to hide the command ID and serialization/deserialization logic in helpers for now.

examples/cross-browser.py
sadym-chromium: I don't want to hide the ID logic in a helper, as reading the protocol from the example would be harder.
https://github.com/GoogleChromeLabs/chromium-bidi/pull/68#discussion_r757911067

examples/cross-browser.py
sadym-chromium: arguments above

examples/cross-browser.py
sadym-chromium: arguments above

examples/cross-browser.py
sadym-chromium: arguments above

examples/cross-browser.py
foolip: OK, I guess there is a tradeoff between helpers &ndash; which are ergonomic and deal with error handling and message order being unreliable &ndash; and code that's obvious about what's happening at the protocol level.

Let's go with your preferences :)

foolip (approved):

PR #76, opened 2021-12-03
`browsingContext` events + `browsingContext.navigate` `wait` param.

sadym-chromium:
1. Emit `browsingContext.load` when `Page.lifecycleEvent` fired with `name="load"`.
2. Emit `browsingContext.domContentLoaded` when `Page.lifecycleEvent` fire with `name="DOMContentLoaded"`.
3. Added and implement `interactive` and `complete` params to `browsingContext.navigate` command.

### DIFF
--- a/src/bidiMapper/domains/context/context.ts
+++ b/src/bidiMapper/domains/context/context.ts
@@ -74,23 +74,27 @@ export class Context {
     this._cdpClient.Page.setLifecycleEventsEnabled({ enabled: true });
 
     this._cdpClient.Page.on('lifecycleEvent', async (params) => {
-      if (params.name === 'DOMContentLoaded')
-        await this._bidiServer.sendMessage({
-          method: 'browsingContext.domContentLoaded',
-          params: {
-            context: this._contextId,
-            navigation: params.loaderId,
-          },
-        });
+      switch (params.name) {
+        case 'DOMContentLoaded':
+          await this._bidiServer.sendMessage({
+            method: 'browsingContext.domContentLoaded',
+            params: {
+              context: this._contextId,
+              navigation: params.loaderId,
+            },
+          });
+          break;
 
-      if (params.name === 'load')
-        await this._bidiServer.sendMessage({
-          method: 'browsingContext.load',
-          params: {
-            context: this._contextId,
-            navigation: params.loaderId,
-          },
-        });
+        case 'load':
+          await this._bidiServer.sendMessage({
+            method: 'browsingContext.load',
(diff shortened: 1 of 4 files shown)


src/bidiMapper/domains/context/context.ts
```diff
@@ -63,6 +63,22 @@ export class Context {
...
+  }
+
+  private _initializeEventListeners() {
+    this._handleDomContentEvent();
+  }
+
+  private _handleDomContentEvent() {
+    this._cdpClient.Page.on('domContentEventFired', async (params) => {
+      await this._bidiServer.sendMessage({
+        method: 'browsingContext.domContentLoaded',
```
bwalderman, line 76: CDP `Page.lifecycleEvent` fires both "DOMContentLoaded" and "load" events. We could possibly use the `frameId` and/or `loaderId` properties to implement the BiDi `navigation` info.

sadym-chromium: Good point, thanks!
Fixed and implemented `browsingContext.load` event.

src/bidiMapper/domains/context/context.ts
```diff
@@ -63,6 +63,35 @@ export class Context {
...
+      if (params.name === 'DOMContentLoaded')
+        await this._bidiServer.sendMessage({
+          method: 'browsingContext.domContentLoaded',
+          params: {
+            context: this._contextId,
+            navigation: params.loaderId,
+          },
+        });
+
+      if (params.name === 'load')
```
bwalderman, line 86: Nit: This can be an `else if`.

src/bidiMapper/domains/context/context.ts
```diff
@@ -121,29 +146,48 @@ export class Context {
...
+        cdpNavigateResult.loaderId!
+      );
+
+      return {
+        navigation: cdpNavigateResult.loaderId,
+        url: url,
+      };
+    }
+
+    if (wait === 'complete') {
```
bwalderman, line 161: This can also be `else if`.

bwalderman (approved):
LGTM with minor suggestions.

src/bidiMapper/domains/context/context.ts
sadym-chromium: replaced with `switch - case`

src/bidiMapper/domains/context/context.ts
sadym-chromium: replaced with `switch - case`

Issue #79, opened 2021-12-07
Prototype `binding` type of `LocalValue`


sadym-chromium:
https://github.com/w3c/webdriver-bidi/issues/157

> Instead of adding global bindings, pass bindings as deserializable arguments to `script.callFunction`. 
...
The argument can be something like:
> 
> ```
> {
>   type: "binding", 
>   value: "SOME_BINDING_NAME"
> }
> ```
> 
> , which will be deserialized to a callback. Calling that callback with any arguments will cause the BiDi event with those arguments like:
> 
> ```
> {
>   method: "script.bindingCalled",
>   params: {
>     arguments: [... remote values ...],
>     name: "SOME_BINDING_NAME"
>   }
> }
> ```

sadym-chromium:
Implemented in https://github.com/GoogleChromeLabs/chromium-bidi/commit/0e627a8153eaae96b8ff5320de1bddaac3a1afc1, but was removed during [switching to CDP serialization](https://github.com/GoogleChromeLabs/chromium-bidi/commit/46d82734760dbf1d7e0a29721874e3fe065f4a31).

sadym-chromium:
Closing in favour of #294

PR #80, opened 2021-12-07
Add `PROTO.binding` type to Local Value.

sadym-chromium:
https://github.com/GoogleChromeLabs/chromium-bidi/issues/79

Based on discussion https://github.com/w3c/webdriver-bidi/issues/157

> Instead of adding global bindings, pass bindings as deserializable arguments to `script.callFunction`. 
...
The argument can be something like:
> 
> ```
> {
>   type: "binding", 
>   value: "SOME_BINDING_NAME"
> }
> ```
> 
> , which will be deserialized to a callback. Calling that callback with any arguments will cause the BiDi event with those arguments like:
> 
> ```
> {
>   method: "script.bindingCalled",
>   params: {
>     arguments: [... remote values ...],
>     name: "SOME_BINDING_NAME"
>   }
> }
> ```

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,8 +125,9 @@ python3 examples/cross-browser.py
 
 ## WPT
 
-WPT is added as a [git submodule](https://git-scm.com/book/en/v2/Git-Tools-Submodules).
-To get run WPT tests:
+WPT is added as
+a [git submodule](https://git-scm.com/book/en/v2/Git-Tools-Submodules). To get
+run WPT tests:
 
 ### Check out WPT and setup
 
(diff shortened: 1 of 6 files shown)


sadym-chromium:
@bwalderman PTAL

bwalderman:
@sadym-chromium, I'm sorry for the delay in reviewing this. I'd like to have a little more consensus on https://github.com/w3c/webdriver-bidi/issues/157 first. I'm about to be on vacation for a few days, but I will update that thread on Monday.

sadym-chromium:
> @sadym-chromium, I'm sorry for the delay in reviewing this. I'd like to have a little more consensus on [w3c/webdriver-bidi#157](https://github.com/w3c/webdriver-bidi/issues/157) first. I'm about to be on vacation for a few days, but I will update that thread on Monday.

Prefix `PROTO` doesn't require us to support the functionality at all. It is there as an illustration of principe suggestion.

bwalderman (approved):

bwalderman:
Approved. Sorry again for the long delay. From the issue discussion, it sounds like there's general agreement on the shape of the API, if not the name.

PR #87, opened 2022-01-11
Run BiDi tests from wpt in CI

foolip:


foolip:
@sadym-chromium this PR now is now in a state where you could try to add the command line arguments to make `./wpt run` use the chromium-bidi implementation. The path to anything in this repo should be ../chromium-bidi/bla when references from the `./wpt run` step.

.github/workflows/wpt.yml
```diff
@@ -0,0 +1,46 @@
...
+        run: ./wpt make-hosts-file | sudo tee -a /etc/hosts
+        working-directory: wpt
+      # TODO: install certs
+      - name: Install Chrome
+        run: ./wpt install chrome browser
+        working-directory: wpt
+      - name: Run tests
+        run: ./wpt run --webdriver-binary=../runBiDiServer.sh --binary _venv3/browsers/nightly/chrome-linux/chrome --log-mach - --log-mach-level info chrome webdriver/tests/bidi/
+        working-directory: wpt
+env:
```
sadym-chromium, line 45: Can we add ` cat log.txt` after the tests? It could help investigate the failures.

foolip: Many servers will be started and they'll clobber each others logs, but yes, I'll add a job to print the log on failure.

runBiDiServer.sh
```diff
@@ -5,4 +5,4 @@ cd "$(dirname $0)/"
 
 NODE_OPTIONS="--unhandled-rejections=strict" \
 DEBUG=* \
-npm run server -- "$@" 2>&1 | tee log.txt
+npm run server-no-build -- "$@" 2>&1 | tee -a log.txt
```
sadym-chromium, line 8: @foolip I added `-a` argument to `tee`. WDYT?

metadata/README.md
```diff
@@ -0,0 +1 @@
+This directory contains expectation metadata for web-platform-tests
```
sadym-chromium, line 1: Minor: maybe move it to `wtp/metadata`?

.github/workflows/wpt.yml
```diff
@@ -0,0 +1,68 @@
...
+        with:
+          name: metadata
+          path: metadata
+      - name: Upload logs
+        if: ${{ always() }}
+        uses: actions/upload-artifact@v2
+        with:
+          name: logs
+          # Note: runBiDiServer.sh will be run multiple times and log.txt will be
+          # overwritten. The problem might not be in the log.txt written last.
```
sadym-chromium, line 64: I added `-a` to `tee`, so the logs should be appended one by one.

foolip: Since multiple instances could be running at the same time I think that might still end up interleaving things. I would suggest to instead have a logs/ directory and have one log per invocation, with a file name using the process ID. That's `$$` in Bash. WDYT?

sadym-chromium (approved):
Great job. Thanks!

PR #92, opened 2022-01-25
`PROTO.cdp.sendCommand`

sadym-chromium:
Implementation of the `BiDi+` extension described in the [BiDi in Puppeteer design doc](https://docs.google.com/document/d/1Vv_EdyryaMQ2RLlMugPQwHf7GfBqh9ESqTHKrWnZmjo).
* `PROTO.cdp.sendCommand`
* `PROTO.cdp.getSession`
* `PROTO.cdp.eventReceived`

Precise CDP events filtering is out of scope https://github.com/GoogleChromeLabs/chromium-bidi/issues/98


src/bidiMapper/commandProcessor.ts
```diff
@@ -174,6 +180,15 @@ export class CommandProcessor {
           commandData as BrowsingContext.PROTO.CloseCommand
         );
 
+      case 'PROTO.cdp.sendCommand':
```
OrKoN, line 183: what does the PROTO prefix indicate? prototype?

src/bidiMapper/domains/context/context.ts
```diff
@@ -86,9 +86,26 @@ export class Context {
 
   private _initializeEventListeners() {
     this._initializePageLifecycleEventListener();
+    this._initializeCdpEventListeners();
     this._handleBindingCalledEvent();
   }
 
+  private _initializeCdpEventListeners() {
+    this._cdpClient.on('event', async (method, params) => {
+      await this._eventManager.sendEvent(
```
OrKoN, line 95: is the event is already handled by the bidi mapper will it be still sent as a CDP command?

sadym-chromium: Yes, CDP events are always sent to client in case of client subscribed to them. It doesn't depend on if BiDi processed it or not.

OrKoN (approved):

PR #99, opened 2022-02-23
Implement `session.unsubscribe`

sadym-chromium:
Implement `session.unsubscribe` to unblock WPT merge.


src/bidiMapper/domains/events/EventManager.ts
```diff
@@ -79,6 +79,24 @@ export class EventManager implements IEventManager {
...
+    }
+  }
+
+  private _unsubscribe(
+    event: string,
+    contextId: BrowsingContext.BrowsingContext | null
+  ): void {
+    if (!this._subscriptions.has(contextId)) return;
+
+    this._subscriptions.get(contextId)!.delete(event);
```
OrKoN, line 98: it's better to use `?` instead of !. `!` is unsafe typecasting whereas would be checked at runtime.

src/bidiMapper/domains/events/EventManager.ts
```diff
@@ -79,6 +79,24 @@ export class EventManager implements IEventManager {
...
+    if (this._subscriptions.get(contextId)!.size === 0)
+      this._subscriptions.delete(contextId);
```
OrKoN, line 100: E.g., smth like this might be better:

```suggestion
    const subscription = this._subscriptions.get(contextId);
    subscription?.delete(event);
    if (subscription?.size === 0) {
      this._subscriptions.delete(contextId);
    }
```

OrKoN (approved):

PR #108, opened 2022-04-25
fix example

sadym-chromium:
#83 removed extra "result" layer, which wasn't reflected in the example. Created and issue to add example to the CI as well: #107. This PR fixes the example script.


mathiasbynens (approved):

mathiasbynens:
Can we get this test to run on CI as well?

sadym-chromium:
> Can we get this test to run on CI as well?
I created an issue for that: #107.

PR #109, opened 2022-04-26
Implement log.entryAdded for javascript error

bwalderman:
Implement the steps for sending a log.entryAdded event when an uncaught javascript error is thrown. Described in the spec under the *error reporting steps* in https://w3c.github.io/webdriver-bidi/#event-log-entryAdded

This is implemented using the CDP equivalent Runtime.exceptionThrown.

Although it's not included in the spec at the moment, I added a `realm` property to the log event so the client can know where the exception came from. This is tracked by https://github.com/w3c/webdriver-bidi/issues/146


sadym-chromium (approved):

sadym-chromium:
Probably WPT expectations should be updated: [Update WPT expectations if needed](https://github.com/GoogleChromeLabs/chromium-bidi#update-wpt-expectations-if-needed)

bwalderman:
The tests results match expectations when I run locally. I'll try updating the expectations to match what's happening in CI.

bwalderman:
@sadym-chromium, please take another look. I updated the implementation to use the CDP `exception` object to get a better string message that matches the WPT tests.

sadym-chromium (approved):

bwalderman:
@whimboo thanks. I do believe the exception text in the WPT tests (a string representation of the thrown exception) is what we should aim for. The `text` field of the CDP exceptionThrown event just seems to be the static text `"Uncaught"` which is not as useful.

PR #112, opened 2022-04-27
Run `examples/cross-browser.py` on CI

mathiasbynens:
Issue: #107


.github/workflows/e2e.js.yml
```diff
@@ -28,5 +28,8 @@ jobs:
         run: python -m pip install -r tests/requirements.txt
       - name: Run e2e tests
         run: npm run e2e
+      - name: Run examples/cross-browser.py
+        run: python3 examples/cross-browser.py
```
sadym-chromium, line 32: LGTM, if you are sure the strep would fail in case of the broken script.

sadym-chromium (approved):

PR #113, opened 2022-04-27
Make examples/cross-browser.py hermetic

mathiasbynens:
Since we want to run this on CI, it’s better to have a fixed target page that is available locally than to have an ever-changing one that requires a network request to access.

Issue: #107

### DIFF
--- a/examples/cross-browser.py
+++ b/examples/cross-browser.py
@@ -85,8 +85,8 @@ async def main():
     # Puppeteer:
     # await page.goto('https://news.ycombinator.com/');
     # https://github.com/puppeteer/puppeteer/blob/4c3caaa3f99f0c31333a749ec50f56180507a374/examples/cross-browser.js#L34
-    appUrl = f'file://{Path(__file__).parent.resolve()}/app.html'
-    print(appUrl)
+    requestedPagePath = f'file://{Path(__file__).parent.resolve()}/app.html'
+    print(requestedPagePath)
     await run_and_wait_command({
         "id": 1001,
         "method": "browsingContext.navigate",
--- a/examples/cross-browser.py
+++ b/examples/cross-browser.py
@@ -20,6 +20,7 @@ import asyncio
 import json
 import os
 import websockets
+from pathlib import Path
 
 
 async def get_websocket():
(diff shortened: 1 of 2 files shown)


examples/cross-browser.py
```diff
@@ -84,11 +85,13 @@ async def main():
     # Puppeteer:
     # await page.goto('https://news.ycombinator.com/');
```
sadym-chromium, line 86: please update the comment. And maybe add clarification where the page came from.

examples/cross-browser.py
```diff
@@ -84,11 +85,13 @@ async def main():
...
     # https://github.com/puppeteer/puppeteer/blob/4c3caaa3f99f0c31333a749ec50f56180507a374/examples/cross-browser.js#L34
+    appUrl = f'file://{Path(__file__).parent.resolve()}/app.html'
+    print(appUrl)
```
sadym-chromium, line 89:
```suggestion
    requestedPagePath = f'file://{Path(__file__).parent.resolve()}/app.html'
    print(requestedPagePath)
```

examples/cross-browser.py
```diff
@@ -84,11 +85,13 @@ async def main():
     # Puppeteer:
     # await page.goto('https://news.ycombinator.com/');
     # https://github.com/puppeteer/puppeteer/blob/4c3caaa3f99f0c31333a749ec50f56180507a374/examples/cross-browser.js#L34
+    requestedPagePath = f'file://{Path(__file__).parent.resolve()}/app.html'
+    print(requestedPagePath)
```
mathiasbynens, line 89:
```suggestion
```

examples/cross-browser.py
```diff
@@ -84,11 +85,12 @@ async def main():
     # Puppeteer:
     # await page.goto('https://news.ycombinator.com/');
     # https://github.com/puppeteer/puppeteer/blob/4c3caaa3f99f0c31333a749ec50f56180507a374/examples/cross-browser.js#L34
+    requestedPagePath = f'file://{Path(__file__).parent.resolve()}/app.html'
```
mathiasbynens, line 88:
```suggestion
    # To avoid the network dependency in this test, we use a local, static copy.
    requestedPagePath = f'file://{Path(__file__).parent.resolve()}/app.html'
```

examples/cross-browser.py
mathiasbynens: Done down below

Issue #118, opened 2022-04-29
Implement `session.new` command


mathiasbynens:
https://w3c.github.io/webdriver-bidi/#command-session-new

mathiasbynens:
Also likely blocking #121.

sadym-chromium:
As the `session` should be implemented by the embedder, our NodeJS server simply accepts all the session commands: https://github.com/GoogleChromeLabs/chromium-bidi/blob/ca2e4ba66e94b5250f86c544ecfe3e5911aab6f4/src/bidiServerRunner.ts#L48

OrKoN:
We need this for Puppeteer. Firefox impl supports it and it's part of the spec.

sadym-chromium:
The best the `mapper` can do is to send a error "500 | session not created", as the session is already exists at the point mapper is created. And the transport implementation (`bidiServerRunner`) should implement the proper request handling.

sadym-chromium:
Closing, as session handling cannot be handled by the Mapper.

OrKoN:
Re-opening this based on the latest understanding that mapper should support multiple sessions and that it should provide handling for capabilities implementable via CDP such as `ignoreHTTPSErrors`. This capability requires calling `Security.setIgnoreCertificateErrors` on every target as there is no other way to implement it in Chrome.

Issue #110, opened 2022-04-27
Get `examples/cross-browser.py` to run on top of Firefox’s BiDi implementation


mathiasbynens:
With Firefox Nighty:

```
$ ./firefox --remote-debugging-port=9222
WebDriver BiDi listening on ws://localhost:9222
DevTools listening on ws://localhost:9222/devtools/browser/6cf65b2b-54e9-444e-a36e-0c04f12b47c9
…

$ # in another terminal / tab

$ PORT=9222 python3 examples/cross-browser.py
Traceback (most recent call last):
  File "~/projects/chromium-bidi/examples/cross-browser.py", line 150, in <module>
    result = loop.run_until_complete(main())
  File "~/homebrew/Cellar/python@3.9/3.9.9/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/base_events.py", line 642, in run_until_complete
    return future.result()
  File "~/projects/chromium-bidi/examples/cross-browser.py", line 53, in main
    websocket = await get_websocket()
  File "~/projects/chromium-bidi/examples/cross-browser.py", line 28, in get_websocket
    return await websockets.connect(url)
  File "~/Library/Python/3.9/lib/python/site-packages/websockets/client.py", line 542, in __await_impl__
    await protocol.handshake(
  File "~/Library/Python/3.9/lib/python/site-packages/websockets/client.py", line 296, in handshake
    raise InvalidStatusCode(status_code)
websockets.exceptions.InvalidStatusCode: server rejected WebSocket connection: HTTP 200
```

We should figure out what’s going on here.

whimboo:
@mathiasbynens this might be related to the host and origin header restrictions which have been discussed on the WebDriver BiDi issue tracker. There is https://github.com/w3c/webdriver-bidi/pull/155 for getting it merged.

Basically WebSocket clients should send an empty `Origin` header. If that's not done by the client then the connection will be denied. Maybe enable `DUMPIO=1` if it's available here as well similar to Puppeteer to see the Firefox stdout logging.

In case of a non-empty `Origin` header there is the `--remote-allow-origins` argument for Firefox which can be used to add specific origins to the allow list. Maybe your client sends a `null` origin? In such a case make use of `--remote-allow-origins=null` when starting Firefox.

Please let me know if this is the issue. If not the stdout logging from Firefox will be very helpful. Maybe you could even set the `remote.log.level` preference to `Trace` when pre-creating the Firefox profile.

[...]

PR #119, opened 2022-04-29
Send `session.new` command in cross-browser example

mathiasbynens:
This is required per spec, and also necessary for the Firefox implementation to accept any non-static commands.

Issue: #110, #118


mathiasbynens:
Closing in favor of #126.

PR #123, opened 2022-05-06
Cdp serialization / deserialization

sadym-chromium:
Use CDP `generateWebDriverValue` for serialization.

The PR happened to be quite huge, as it involved lots of refactoring.

* Move serialization to CDP-side by using `generateWebDriverValue` in CallFunction, Evaluate and when serialization is needed.
* Remove in-realm serialization / deserialization.
* Reformatted some files.
* Implemented CDP-serialization.
* Serialization / deserialization tests.
* Make e2e and WPT tests run on chrome-dev channel (as serialization is not in the stable Chrome yet).
* Added e2e timeout 15 minutes to prevent hanging the step forever.


.github/workflows/e2e.js.yml
```diff
@@ -12,25 +12,31 @@ jobs:
...
-        uses: actions/setup-python@v2
-        with:
-          python-version: 3.8
-      - name: Install Python dependencies
-        run: python -m pip install -r tests/requirements.txt
-      - name: Run e2e tests
-        run: npm run e2e
-      - name: Run examples/cross-browser.py
-        run: python3 examples/cross-browser.py
+    - uses: actions/checkout@v2
```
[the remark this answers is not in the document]
sadym-chromium: done

src/bidiMapper/domains/context/context.ts
```diff
@@ -219,187 +136,127 @@ export class Context {
...
-    }
+    });
+  }
 
-    return findElementCommandResult.result
-      .value as CommonDataTypes.NodeRemoteValue;
+  #initializePageLifecycleEventListener() {
+    this.#cdpClient.Page.setLifecycleEventsEnabled({ enabled: true });
+
+    this.#cdpClient.Page.on('lifecycleEvent', async (params) => {
```
OrKoN, line 206: should there be .off calls for every .on call somewhere?

src/bidiMapper/domains/context/scriptEvaluator.ts
```diff
@@ -0,0 +1,394 @@
...
+    serializedValue: CommonDataTypes.LocalValue
+  ): Promise<Protocol.Runtime.CallArgument> {
+    if (`objectId` in serializedValue) {
+      return { objectId: serializedValue.objectId };
+    }
+    switch (serializedValue.type) {
+      // Primitive Protocol Value
+      // https://w3c.github.io/webdriver-bidi/#data-types-protocolValue-primitiveProtocolValue
+      case 'undefined': {
+        return { unserializableValue: 'undefined' };
```
OrKoN, line 214: what does unserializableValue means? a value that cannot be deserialized?

OrKoN:
LGTM in general but I am not very familiar with the bidi spec to see if serialisation is correct.

src/bidiMapper/domains/context/scriptEvaluator.ts
sadym-chromium: [`Runtime.CallArgument. unserializableValue`](https://chromedevtools.github.io/devtools-protocol/tot/Runtime/#type-CallArgument) is basically a string that is evaluated instead of being JSON-deserialized.

sadym-chromium:
> I am not very familiar with the bidi spec to see if serialisation is correct.

We delegate the serialization logic to CDP, but verify it in e2e tests. Deserialization is on the Mapper side, but it is tested by e2e as well.

src/bidiMapper/domains/context/context.ts
sadym-chromium: Let's keep it out of scope of this PR. Ideally, we should subscribe to events only when user subscribed to them, and unsubscribe when user unsubscribed. But for now Mapper (as well as Puppeteer, AFAIK) just listens to all event and sends the needed to user.

OrKoN (approved):

PR #167, opened 2022-05-20
Fix Evaluate and CallFunction types

sadym-chromium:
Align CallFunction and Evaluate results with the [W3C BiDi spec](https://w3c.github.io/webdriver-bidi/#type-script-Result):
```
{
    result: {
        result | exceptionDetails: {...}
    }
}
```

### DIFF
--- a/examples/cross-browser.py
+++ b/examples/cross-browser.py
@@ -143,18 +143,20 @@ async def main():
     # {
     #     "id": 1002,
     #     "result": {
-    #         "objectId": "__SOME_OBJECT_ID__",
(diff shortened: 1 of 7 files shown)


OrKoN (approved):
LGTM but is it possible to now have result.result on the spec level?

sadym-chromium:
The spec requires having `result.result` or `result.exceptionDetails`. The change in the spec was after our implementation is done, so it wasn't reflected in the Mapper. This PR fixes the issue and aligns Mapper with the spec.

PR #205, opened 2022-06-27
Add simplified example script

sadym-chromium:
1. Add  `wait: complete` to `browsingContext.navigate`. Otherwise there are no elements by the time command is finished.
2. Added simplified (and not aligned with Puppeteer) script version, using `script.evaluate` instead of `script.callFunction`.

### DIFF
--- a/examples/cross-browser-simplified.py
+++ b/examples/cross-browser-simplified.py
@@ -67,7 +67,7 @@ async def main():
         "params": {}}, websocket)
 
     # Not implemented:
-    # await browser.version());
+    # await browser.version();
     # https://github.com/puppeteer/puppeteer/blob/4c3caaa3f99f0c31333a749ec50f56180507a374/examples/cross-browser.js#L32
 
     # Puppeteer:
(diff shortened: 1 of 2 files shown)


examples/cross-browser-simplified.py
```diff
@@ -0,0 +1,168 @@
...
+    # const browser = await puppeteer.launch();
+    # https://github.com/puppeteer/puppeteer/blob/4c3caaa3f99f0c31333a749ec50f56180507a374/examples/cross-browser.js#L29
+    websocket = await get_websocket()
+    await run_and_wait_command({
+        "id": 0,
+        "method": "session.new",
+        "params": {}}, websocket)
+
+    # Not implemented:
+    # await browser.version());
```
thiagowfx, line 70: del extra closing parenthesis in comment

thiagowfx (approved):

examples/cross-browser-simplified.py
sadym-chromium: nice catch!

Issue #89, opened 2022-01-17
Optimize Mapper compiled code


sadym-chromium:
**Actual:** the rolled-up Mapper script `/src/.build/bidiMapper/mapper.js` has a size of **`748K`**, and the vast majority of the file is CDP string literals describing CDP methods, which are added for development convenience and are required for the functionality.

Removing those definitions reduces the size of the plain `/src/.build/bidiMapper/mapper.js` to **`68K`**, and minimizing afterwards leads to only **`28K`** source size.

**Expected:** the rolled-up Mapper script `/src/.build/bidiMapper/mapper.js` doesn't have unnecessary string literals and has a size of ~ **`30K`**.

thiagowfx:
Maksim:

> Try to compile mapper, and check its size. The vast majority of the space is taken by the CDP method definitions.

thiagowfx:
cc @vladimir-nechaev @mathiasbynens

The first commit that integrates terser has been already merged (c.f. #221). To close this bug I need more clarification from @sadym-chromium w.r.t.:

> Removing those definitions [CDP string literals describing CDP methods, which are added for development convenience and are required for the functionality]

Let me know how we can accomplish that.

OrKoN:
I think it could be achieved by removing the following imports in the cdp client:

```
import * as browserProtocol from 'devtools-protocol/json/browser_protocol.json';
import * as jsProtocol from 'devtools-protocol/json/js_protocol.json';
```

Perhaps instead of generating classes at runtime based on JSON definitions, we could employ a Proxy?

sadym-chromium:
> I think it could be achieved by removing the following imports in the cdp client:
> 
> ```
> import * as browserProtocol from 'devtools-protocol/json/browser_protocol.json';
> import * as jsProtocol from 'devtools-protocol/json/js_protocol.json';
> ```
> 
> Perhaps instead of generating classes at runtime based on JSON definitions, we could employ a Proxy?

@thiagowfx IDK how exactly to reach that, and the issue involves some research. Alex's approach looks promising.

sadym-chromium:
Additionally I'd consider adding a step verifying the compiled size in the given diapason.

thiagowfx:
> Additionally I'd consider adding a step verifying the compiled size in the given diapason.

That was a good idea! Done on https://github.com/GoogleChromeLabs/chromium-bidi/pull/224

thiagowfx:
[...]

PR #224, opened 2022-07-22
github workflow: add step to print the mapper file size

thiagowfx:
This adds a straightforward way to measure the effectiveness of #89
wherein we intend to optimize mapper compiled code.


thiagowfx:
c.f. https://github.com/GoogleChromeLabs/chromium-bidi/runs/7466757420?check_suite_focus=true

sadym-chromium (approved):

PR #227, opened 2022-07-25
reduce mapper protocol domains size with jq by filtering out unneeded fields

thiagowfx:


thiagowfx:
Just a bit of refactoring. Real changes will come later.

thiagowfx:
Sure, I can! I was hoping to first hear if you have any concerns with the current approach, before we take the effort to document it.

sadym-chromium (approved):
LGTM if the description of the new optimisation is added

PR #232, opened 2022-08-05
Deferred

sadym-chromium:
Step 1 towards Out Of Process iFrames (OOPiF):
* Reformat with `prettier`
* Introduced `Deferred`
* Switch Context to Deferred, unties the waiter from the specific CDP session, and which make sit possible to transparently switch sessions.

### DIFF
--- a/src/bidiMapper/domains/context/context.ts
+++ b/src/bidiMapper/domains/context/context.ts
@@ -97,14 +97,17 @@ export abstract class Context {
     this.#sessionId = sessionId;
     this.cdpClient = cdpClient;
 
-    this.setCDPListeners();
+    this.#setCdpListeners();
   }
 
-  setCDPListeners() {
+  #setCdpListeners() {
     this.cdpClient.Page.on(
       'lifecycleEvent',
-      async (params: Protocol.Page.LifecycleEventEvent) => {
-        let map = undefined;
+      (params: Protocol.Page.LifecycleEventEvent) => {
+        let map: Map<
+          string,
+          Deferred<Protocol.Page.LifecycleEventEvent>
+        > | null = null;
         if (params.name === 'DOMContentLoaded') {
           map = this.#targetDeferres.Page.lifecycleEvent.DOMContentLoaded;
         }
@@ -118,7 +121,7 @@ export abstract class Context {
 
     this.cdpClient.Page.on(
       'navigatedWithinDocument',
-      async (params: Protocol.Page.NavigatedWithinDocumentEvent) => {
+      (params: Protocol.Page.NavigatedWithinDocumentEvent) => {
         if (params.frameId !== this.getContextId()) {
           return;
         }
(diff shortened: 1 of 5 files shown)


src/bidiMapper/utils/deferred.ts
```diff
@@ -0,0 +1,65 @@
+/**
+ * Copyright 2021 Google LLC.
```
thiagowfx, line 2:
```suggestion
 * Copyright 2022 Google LLC.
```

src/bidiMapper/utils/deferred.ts
```diff
@@ -0,0 +1,65 @@
...
+    this.#promise = new Promise((resolve, reject) => {
+      this.#resolve = resolve;
+      this.#reject = reject;
+    });
+  }
+
+  then<TResult1 = T, TResult2 = never>(
+    onFulfilled?:
+      | ((value: T) => TResult1 | PromiseLike<TResult1>)
+      | undefined
```
thiagowfx, line 33: Since `onFulfilled?` is optional, you should not need `| undefined` here. Same applies below (3 more times, including `catch` and `finally`).

src/bidiMapper/domains/context/context.ts
```diff
@@ -82,6 +96,36 @@ export abstract class Context {
     this.#parentId = parent;
     this.#sessionId = sessionId;
     this.cdpClient = cdpClient;
+
+    this.setCDPListeners();
```
thiagowfx, line 100: In order to follow http://go/jsstyle (open source: https://google.github.io/styleguide/jsguide.html), this should be `setCdpListeners` c.f. https://google.github.io/styleguide/jsguide.html#naming-rules-common-to-all-identifiers

src/bidiMapper/domains/context/context.ts
```diff
@@ -82,6 +96,36 @@ export abstract class Context {
...
+  }
+
+  setCDPListeners() {
+    this.cdpClient.Page.on(
+      'lifecycleEvent',
+      async (params: Protocol.Page.LifecycleEventEvent) => {
+        let map = undefined;
```
thiagowfx, line 107: Could you type annotate `map`?

src/bidiMapper/domains/context/context.ts
```diff
@@ -82,6 +96,36 @@ export abstract class Context {
...
     this.#sessionId = sessionId;
     this.cdpClient = cdpClient;
+
+    this.setCDPListeners();
+  }
+
+  setCDPListeners() {
+    this.cdpClient.Page.on(
+      'lifecycleEvent',
+      async (params: Protocol.Page.LifecycleEventEvent) => {
```
thiagowfx, line 106: If there's no `await`, you should not need `async`. Same below.

src/bidiMapper/utils/deferred.ts
sadym-chromium: The original interface was copied from the TypeScript Promise interface definition. But it can be simplified. Done.

src/bidiMapper/domains/context/context.ts
sadym-chromium: done

src/bidiMapper/domains/context/context.ts
sadym-chromium: done

src/bidiMapper/domains/context/context.ts
sadym-chromium: done

thiagowfx (approved):

PR #235, opened 2022-08-12
Add tests for strict mode

sadym-chromium:
Add test for strict mode in `evaluate` and `callFunction`

### DIFF
--- a/tests/_helpers.py
+++ b/tests/_helpers.py
@@ -110,16 +110,33 @@ def recursive_compare(expected, actual):
     assert expected == actual
 
 
-def any_string(expected):
-    assert isinstance(expected, str), \
-        f"'{expected}' should be string, " \
-        f"but is {type(expected)} instead."
+def error_response(exception, exception_text):
+    return lambda actual: recursive_compare({"exceptionDetails": {
+        "exception": exception,
+        "columnNumber": any_int,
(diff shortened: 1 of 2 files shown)


tests/_helpers.py
```diff
@@ -110,16 +110,33 @@ def recursive_compare(expected, actual):
...
+        f"but is {type(actual)} instead."
+
+
+def any_int(actual):
+    assert isinstance(actual, int), \
+        f"'{actual}' should be string, " \
+        f"but is {type(actual)} instead."
+
+
+def any_timestamp(actual):
```
thiagowfx, line 136: How about `any_number` instead?

thiagowfx (approved):

mathiasbynens:
Why are these not WPT tests?

sadym-chromium:
> Why are these not WPT tests?

There are such tests in WPT as well: https://github.com/web-platform-tests/wpt/pull/35448

Some of our e2e tests intersect with WPT, and this is an example. Do you think this tests are too niche, and should not be in e2e tests? I'm inclined to agree with that position.

sadym-chromium:
The tests are too niche, and we can keep them only in WPT: https://github.com/web-platform-tests/wpt/pull/35448

Issue #241, opened 2022-08-18
Sandbox support


sadym-chromium:
- [x] Mapper implements execution sandboxes.
- [x] WPT tests are passing ([`webdriver/tests/bidi/script/get_realms/sandbox.py.ini` is removed](https://github.com/GoogleChromeLabs/chromium-bidi/blob/main/wpt-metadata/webdriver/tests/bidi/script/get_realms/sandbox.py.ini)).

PR #242, opened 2022-08-18
Support Sandboxes

sadym-chromium:
Addressing #241 :
* Added `sandbox` logic.
* Added e2e tests.
* Throwing "not implemented error" in case of `realm` parameter.
* Updating WPT expectations (after exception is thrown).

TBD in another PR:
* Adding `realm` support.

### DIFF
--- a/src/bidiMapper/domains/context/browsingContextImpl.ts
+++ b/src/bidiMapper/domains/context/browsingContextImpl.ts
@@ -51,7 +51,10 @@ export class BrowsingContextImpl {
   #eventManager: IEventManager;
   readonly #children: BrowsingContextImpl[] = [];
   #scriptEvaluator: ScriptEvaluator;
-  #executionContext: number | null = null;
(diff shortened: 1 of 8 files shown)


src/bidiMapper/domains/context/browsingContextImpl.ts
```diff
@@ -51,7 +51,10 @@ export class BrowsingContextImpl {
   #eventManager: IEventManager;
   readonly #children: BrowsingContextImpl[] = [];
   #scriptEvaluator: ScriptEvaluator;
-  #executionContext: number | null = null;
+  readonly #sandboxToExecutionContextIdMap: Map<
+    string | null,
```
thiagowfx, line 55: What does `null` mean in this context? We should probably add a doc comment here.

sadym-chromium: Added comments describing. I use `null` for default execution context.

thiagowfx (approved):

PR #245, opened 2022-08-24
Align `session.status` with spec

sadym-chromium:
As we don't support creating additional sessions, the `ready` state should be `false`:
> The readiness state of a [remote end](https://w3c.github.io/webdriver/#dfn-remote-ends) indicates whether it is free to accept new connections.


thiagowfx:
I am not familiar with this part, but I assume you know you're doing so I am approving it.

sadym-chromium:
> I am not familiar with this part, but I assume you know you're doing so I am approving it.

I updated the description

PR #253, opened 2022-09-14
Avoid race condition in events when during `browsingContext.create`

sadym-chromium:
Avoid race condition in the load events when during `browsingContext.create`.

Motivation and discussion are here: https://github.com/web-platform-tests/wpt/issues/35846


thiagowfx (approved):

sadym-chromium:
Waiting for the final solution in https://github.com/web-platform-tests/wpt/issues/35846

PR #266, opened 2022-09-24
Add realm support

sadym-chromium:
* Add `script.getRealms` command.
* Track `executionContextId` creating and destroying.
* Use [`Runtime.ExecutionContextDescription. uniqueId`](https://chromedevtools.github.io/devtools-protocol/tot/Runtime/#type-ExecutionContextDescription) as a BiDi realm.
* Added e2e tests.
* Don't allow empty string as a `sandbox`.

Follow up: refactoring of the BrowsingContexts #269


src/bidiMapper/domains/protocol/bidiProtocolTypes.ts
```diff
@@ -402,6 +408,86 @@ export namespace Script {
...
+  export type WorkletRealmInfo = BaseRealmInfo & {
+    type: 'worklet';
+  };
+
+  const RealmTypeSchema = zod.enum([
+    'window',
+    'dedicated-worker',
+    'shared-worker',
+    'service-worker',
+    'worker',
```
thiagowfx, line 465: if the order does not matter, can we sort this list?

src/bidiMapper/domains/context/browsingContextProcessor.ts
```diff
@@ -312,36 +313,66 @@ export class BrowsingContextProcessor {
...
+
   async process_script_callFunction(
     params: Script.CallFunctionParameters
   ): Promise<Script.CallFunctionResult> {
-    if ('realm' in params.target) {
-      throw new Error('Realm target is not implemented yet.');
-    }
-    const target: Script.ContextTarget = params.target;
+    const scriptTarget = BrowsingContextProcessor.#getScriptTarget(
+      params.target
```
thiagowfx, line 366:
```suggestion
      params.target,
```

src/bidiMapper/domains/context/browsingContextImpl.ts
```diff
@@ -308,8 +319,56 @@ export class BrowsingContextImpl {
           // Default execution context is set with key `null`.
           this.#sandboxToExecutionContextIdMap.set(null, params.context.id);
         }
+        if (params.context.name !== undefined) {
+          this.#sandboxToExecutionContextIdMap.set(
+            params.context.name,
+            params.context.id
```
thiagowfx, line 325:
```suggestion
            params.context.id,
```

tests/_helpers.py
```diff
@@ -107,25 +107,41 @@ def recursive_compare(expected, actual):
...
-        f"but is {type(expected)} instead."
+def not_one_of(not_expected_list):
+    def _not_one_of(actual):
+        for not_expected in not_expected_list:
+            assert actual != not_expected
+
+    return _not_one_of
+
+
+def comppare_sorted(key_name, expected):
```
thiagowfx, line 124: typo: compare_sorted

thiagowfx (approved):

src/bidiMapper/domains/context/browsingContextProcessor.ts
sadym-chromium: Prettier doesn't think so )

src/bidiMapper/domains/context/browsingContextImpl.ts
sadym-chromium: Same. Prettier should be updated instead, but it's out of scope.

src/bidiMapper/domains/protocol/bidiProtocolTypes.ts
sadym-chromium: The order is aligned with [the order in spec](https://w3c.github.io/webdriver-bidi/#type-script-RealmInfo).

src/bidiMapper/domains/context/browsingContextProcessor.ts
mathiasbynens: This is a style discussion. If we want trailing commas for function params across multiple lines we could set `trailingComma: 'all'` in the Prettier config. But I agree that we should discuss that separately from the code changes in this PR 👍

PR #307, opened 2022-11-07
Synchronous exception text retrieval

sadym-chromium:
WPT log tests are racy and sometimes fail: [[1]](https://github.com/GoogleChromeLabs/chromium-bidi/actions/runs/3409317669/jobs/5670936087), [[2]](https://github.com/GoogleChromeLabs/chromium-bidi/actions/runs/3409460492). 

Log analysis showed the root cause is in the extra round-trip `realm.stringifyObject`. To avoid the race condition and make the CDP-BiDi event forwarding synchronous, some heuristic is implemented:
1. If exception has a description, use it. Date, Error with stacktrace, non-trivial objects.
2. If exception has value, use it. Primitives like number, string.
3. If exception has an objectId, use a CDP round-trip to make `toString` of it.
4. Return the whole exceptionDetails data if none of the above worked.

[BiDi Spec says](https://w3c.github.io/webdriver-bidi/#event-log-entryAdded) the `text` for non-primitives is implementation-defined.

WPT tests relies on the specific text format, and has to be updated.

### DIFF
--- a/src/bidiMapper/domains/log/logManager.ts
+++ b/src/bidiMapper/domains/log/logManager.ts
@@ -16,11 +16,12 @@
  */
 
 import { CdpClient } from '../../../cdp';
-import { Log, Script } from '../protocol/bidiProtocolTypes';
+import { Log } from '../protocol/bidiProtocolTypes';
 import { getRemoteValuesText } from './logHelper';
 import { Protocol } from 'devtools-protocol';
 import { Realm } from '../script/realm';
 import { IEventManager } from '../events/EventManager';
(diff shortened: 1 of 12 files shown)


thiagowfx (approved):

sadym-chromium:
Pause the PR to try another approach. Converted to draft.

sadym-chromium:
Closed in favour of https://github.com/GoogleChromeLabs/chromium-bidi/pull/310. Motivation: http://go/webdriver-non-racy-serialization-proposal.

PR #310, opened 2022-11-09
Non-racy serialization with message queue

sadym-chromium:
Implementation of the "processing queue with promises": http://go/webdriver-non-racy-serialization-proposal.

To provide serialization of the events, additional CDP call is required. When such a call is done, the order of the events can be changed. This PR implements `ProcessingQueue`, which allows adding Promises instead of the real values, and guarantee the order of the messages are kept.

The promise queue is for consoleAPI events.

Extended e2e tests to cover result, exception and log serialization.

### DIFF
--- a/src/utils/processingQueue.ts
+++ b/src/utils/processingQueue.ts
@@ -53,7 +53,7 @@ export class ProcessingQueue<T> {
         await entryPromise
           .then((entry) => this.#processor(entry))
           .catch((e) => {
-            logSystem('Event was not processed! ' + e);
+            logSystem('Event was not processed:' + e);
             this.#catch(e);
           })
           .finally();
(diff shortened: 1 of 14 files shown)


src/utils/processingQueue.ts
```diff
@@ -48,8 +52,10 @@ export class ProcessingQueue<T> {
       if (entryPromise !== undefined) {
         await entryPromise
           .then((entry) => this.#processor(entry))
-          .catch((e) => this.#catch(e))
-          // Continue processing queue.
+          .catch((e) => {
+            logSystem('Event was not processed! ' + e);
```
thiagowfx, line 56:
```suggestion
            logSystem('Event was not processed:' + e);
```

thiagowfx (approved):

PR #313, opened 2022-11-16
Perform miscellaneous ergonomic changes

thiagowfx:
These changes are trivial and not worth to be documented individually. Things like: sorting, consistency, removal of redundant comments/docs, removal of deprecated docs, upgrade of python packages, and minor styles adjustments. I go through these as I try to get familiar with every file in the codebase, then make these small tweaks and adjustments.


.gitignore
```diff
@@ -5,3 +5,4 @@ __pycache__
 logs/
 chromium-bidi.iml
 node_modules/
+package-lock.json
```
thiagowfx, line 8: this file is created when we do some `npm` operations, so it's worth to explicitly exclude it

examples/README.md
```diff
@@ -28,22 +26,16 @@ npm run server
...
 ```
 
 Firefox can be run by:
 
 ```sh
-$FIREFOX_BINARY_PATH --remote-debugging-port=8080
+$FIREFOX_BINARY_PATH [--remote-debugging-port=8080]
 ```
 
-CDP-BiDi Mapper can be run from the project root:
```
thiagowfx, line 40: removed because it was stated twice in this file

examples/cross-browser.py
```diff
@@ -141,14 +144,11 @@ async def main():
         "type": "string",
         "value": ".titlelink"}
 
-    # `script.callFunction` is not implemented by Firefox yet:
-    # https://bugzilla.mozilla.org/show_bug.cgi?id=1750541
```
thiagowfx, line 145: This is now implemented upstream

.gitignore
mathiasbynens: We could also check in a `.npmrc` with `package-lock=false` in it to prevent npm from even touching the file.

tests/requirements.txt
```diff
@@ -1,3 +1,3 @@
 pytest===6.1.2
-pytest-asyncio==0.14.0
-websockets==8.1
+pytest-asyncio==0.20.2
+websockets==10.4
```
sadym-chromium, line 3: Looks like this change broke e2e tests.

thiagowfx: I'll revert the websockets change only. Let's see if the `pytest-asyncio` one is fine.

.gitignore
thiagowfx: Is that your preference? I am happy to do either. We currently don't have a `.npmrc` file.

mathiasbynens: The `.npmrc` feels strictly better since it avoids doing the unneeded busywork of reading/updating the `package-lock.json` altogether, but I guess the difference doesn't matter much in practice. We could even do both. No strong opinions here :)

thiagowfx: Thanks for the explanation. Done.

sadym-chromium (approved):

Issue #294, opened 2022-10-13
Implement `channel` mechanism


sadym-chromium:
https://github.com/w3c/webdriver-bidi/issues/157

Prototype the mechanism using long-poll. Implementation details: http://go/webdriver-binding-implementation.

## Status

- [x] Prefactoring: https://github.com/GoogleChromeLabs/chromium-bidi/pull/542
- [x] Introduce `Script.Message`: https://github.com/GoogleChromeLabs/chromium-bidi/pull/541
- [x] Implement binding mechanism: https://github.com/GoogleChromeLabs/chromium-bidi/pull/319

thiagowfx:
Spec work seems to be in progress in https://github.com/w3c/webdriver-bidi/pull/361/files

thiagowfx:
@sadym-chromium can this be closed?

PR #319, opened 2022-11-21
Prototype channel mechanism (`script.MessageEvent`)

thiagowfx:
Provide a way to communicate from scripts evaluated on the page side to
the BiDi client.

This is a safe (non-observable by the page script) analog of the `Runtime.addBinding` CDP command.

Prototype the mechanism using long-poll. Implementation details: http://go/webdriver-binding-implementation.

Reference: https://chromedevtools.github.io/devtools-protocol/tot/Runtime/#method-callFunctionOn

Bug: https://github.com/GoogleChromeLabs/chromium-bidi/issues/294

## Aligning with spec:

- [x] BindingValue -> Channel
- [x] ScriptBindingCalledParams -> MessageParams

## Status

- [x] Introduce `Script.Message`: forked to https://github.com/GoogleChromeLabs/chromium-bidi/pull/541
- [x] Finish passing argument to the event in the deserialization method.
- [x] Rename `PROTO.binding` to channel to match spec.

### DIFF
--- a/src/bidiMapper/domains/script/scriptEvaluator.ts
+++ b/src/bidiMapper/domains/script/scriptEvaluator.ts
@@ -17,6 +17,7 @@
 import {Protocol} from 'devtools-protocol';
 
 import {CommonDataTypes, Message, Script} from '../../../protocol/protocol.js';
+import {IEventManager} from '../events/EventManager.js';
 
 import {Realm} from './realm.js';
 
@@ -27,6 +28,12 @@ const EVALUATE_STACKTRACE_LINE_OFFSET = 0;
 export const SHARED_ID_DIVIDER = '_element_';
 
 export class ScriptEvaluator {
+  readonly #eventManager: IEventManager;
+
+  constructor(eventManager: IEventManager) {
+    this.#eventManager = eventManager;
+  }
+
   /**
    * Gets the string representation of an object. This is equivalent to
    * calling toString() on the object value.
@@ -392,6 +399,118 @@ export class ScriptEvaluator {
         );
         // TODO(#375): Release `argEvalResult.result.objectId` after using.
         return {objectId: argEvalResult.result.objectId};
+      }
+
+      case 'channel': {
+        const channelName = argumentValue.value.channel;
+
+        const argEvalResult = await realm.cdpClient.sendCommand(
+          'Runtime.callFunctionOn',
+          {
+            functionDeclaration: String(() => {
+              const queue: unknown[] = [];
+              let resolver: null | (() => void) = null;
+
+              return {
+                /**
+                 * Gets a promise, which is resolved as soon as a message occurs
+                 * in the queue.
+                 */
+                async getMessage(): Promise<unknown> {
+                  const onMessage =
+                    queue.length > 0
+                      ? Promise.resolve()
+                      : new Promise<void>((resolve) => {
+                          resolver = resolve;
+                        });
+                  await onMessage;
+                  return queue.shift();
+                },
+
+                /**
+                 * Adds a message to the queue.
+                 * If the queue was initially empty, resolve the pending promise.
+                 */
+                sendMessage(message: string) {
+                  queue.push(message);
+                  if (resolver !== null) {
+                    resolver();
+                    resolver = null;
+                  }
+                },
+              };
+            }),
+            returnByValue: false,
+            executionContextId: realm.executionContextId,
+            generateWebDriverValue: false,
+          }
+        );
+        const channelHandle = argEvalResult.result.objectId;
+
+        // Long-poll the message queue asynchronously.
+        (async () => {
+          // TODO: Remove this loop after the realm is destroyed.
+          // Rely on the CDP throwing exception in such a case.
+          for (;;) {
+            const message = await realm.cdpClient.sendCommand(
+              'Runtime.callFunctionOn',
+              {
+                functionDeclaration: String(
+                  async (channelHandle: {
+                    getMessage: () => Promise<unknown>;
+                    sendMessage: (message: string) => void;
+                  }) => channelHandle.getMessage()
+                ),
+                arguments: [
+                  {
+                    objectId: channelHandle,
+                  },
+                ],
+                awaitPromise: true,
+                executionContextId: realm.executionContextId,
+                generateWebDriverValue: true,
+              }
+            );
+
+            this.#eventManager.registerEvent(
+              {
+                method: Script.EventNames.MessageEvent,
+                params: {
+                  channel: channelName,
+                  data: await realm.cdpToBidiValue(
+                    message,
+                    argumentValue.value.ownership ?? 'none'
+                  ),
+                  source: {
+                    realm: realm.realmId,
+                    context: realm.browsingContextId,
+                  },
+                },
+              },
+              realm.browsingContextId
+            );
+          }
+        })();
+
+        const sendMessageArgResult = await realm.cdpClient.sendCommand(
+          'Runtime.callFunctionOn',
+          {
+            functionDeclaration: String(
+              (channelHandle: {sendMessage: (message: string) => void}) => {
+                return channelHandle.sendMessage;
+              }
+            ),
+            arguments: [
+              {
+                objectId: channelHandle,
+              },
+            ],
+            returnByValue: false,
+            executionContextId: realm.executionContextId,
+            generateWebDriverValue: false,
+          }
+        );
+        return {objectId: sendMessageArgResult.result.objectId};
       }
 
       // TODO(#375): Dispose of nested objects.
--- a/src/bidiMapper/domains/script/scriptEvaluator.ts
+++ b/src/bidiMapper/domains/script/scriptEvaluator.ts
@@ -21,6 +21,55 @@ import {IEventManager} from '../events/EventManager.js';
 
 import {Realm} from './realm.js';
 
+async function initChannelListener(
+  channel: Script.Channel,
+  channelHandle: string | undefined,
+  eventManager: IEventManager,
+  realm: Realm
+) {
+  const channelId = channel.value.channel;
+
+  // TODO(#294): Remove this loop after the realm is destroyed.
+  // Rely on the CDP throwing exception in such a case.
+  for (;;) {
+    const message = await realm.cdpClient.sendCommand(
+      'Runtime.callFunctionOn',
+      {
+        functionDeclaration: String(
+          async (channelHandle: {getMessage: () => Promise<unknown>}) =>
+            channelHandle.getMessage()
+        ),
+        arguments: [
+          {
+            objectId: channelHandle,
+          },
+        ],
+        awaitPromise: true,
+        executionContextId: realm.executionContextId,
+        generateWebDriverValue: true,
+      }
+    );
+
+    eventManager.registerPromiseEvent(
+      realm
+        .cdpToBidiValue(message, channel.value.ownership ?? 'none')
+        .then((data) => ({
+          method: Script.EventNames.MessageEvent,
+          params: {
+            channel: channelId,
+            data,
+            source: {
+              realm: realm.realmId,
+              context: realm.browsingContextId,
+            },
+          },
+        })),
+      realm.browsingContextId,
+      Script.EventNames.MessageEvent
+    );
+  }
+}
+
 // As `script.evaluate` wraps call into serialization script, `lineNumber`
 // should be adjusted.
 const CALL_FUNCTION_STACKTRACE_LINE_OFFSET = 1;
@@ -402,14 +451,12 @@ export class ScriptEvaluator {
       }
 
       case 'channel': {
-        const channelName = argumentValue.value.channel;
-
-        const argEvalResult = await realm.cdpClient.sendCommand(
+        const createChannelHandleResult = await realm.cdpClient.sendCommand(
           'Runtime.callFunctionOn',
           {
             functionDeclaration: String(() => {
               const queue: unknown[] = [];
-              let resolver: null | (() => void) = null;
+              let queueNonEmptyResolver: null | (() => void) = null;
 
               return {
                 /**
@@ -421,7 +468,7 @@ export class ScriptEvaluator {
                     queue.length > 0
                       ? Promise.resolve()
                       : new Promise<void>((resolve) => {
-                          resolver = resolve;
+                          queueNonEmptyResolver = resolve;
                         });
                   await onMessage;
                   return queue.shift();
@@ -429,13 +476,13 @@ export class ScriptEvaluator {
 
                 /**
                  * Adds a message to the queue.
-                 * If the queue was initially empty, resolve the pending promise.
+                 * Resolves the pending promise if needed.
                  */
                 sendMessage(message: string) {
                   queue.push(message);
-                  if (resolver !== null) {
-                    resolver();
-                    resolver = null;
+                  if (queueNonEmptyResolver !== null) {
+                    queueNonEmptyResolver();
+                    queueNonEmptyResolver = null;
                   }
                 },
               };
@@ -445,52 +492,15 @@ export class ScriptEvaluator {
             generateWebDriverValue: false,
           }
         );
-        const channelHandle = argEvalResult.result.objectId;
+        const channelHandle = createChannelHandleResult.result.objectId;
 
         // Long-poll the message queue asynchronously.
-        (async () => {
-          // TODO: Remove this loop after the realm is destroyed.
-          // Rely on the CDP throwing exception in such a case.
-          for (;;) {
-            const message = await realm.cdpClient.sendCommand(
-              'Runtime.callFunctionOn',
-              {
-                functionDeclaration: String(
-                  async (channelHandle: {
-                    getMessage: () => Promise<unknown>;
-                    sendMessage: (message: string) => void;
-                  }) => channelHandle.getMessage()
-                ),
-                arguments: [
-                  {
-                    objectId: channelHandle,
-                  },
-                ],
-                awaitPromise: true,
-                executionContextId: realm.executionContextId,
-                generateWebDriverValue: true,
-              }
-            );
-
-            this.#eventManager.registerEvent(
-              {
-                method: Script.EventNames.MessageEvent,
-                params: {
-                  channel: channelName,
-                  data: await realm.cdpToBidiValue(
-                    message,
-                    argumentValue.value.ownership ?? 'none'
-                  ),
-                  source: {
-                    realm: realm.realmId,
-                    context: realm.browsingContextId,
-                  },
-                },
-              },
-              realm.browsingContextId
-            );
-          }
-        })();
+        initChannelListener(
+          argumentValue,
+          channelHandle,
+          this.#eventManager,
+          realm
+        );
 
         const sendMessageArgResult = await realm.cdpClient.sendCommand(
           'Runtime.callFunctionOn',
--- a/src/protocol-parser/protocol-parser.ts
+++ b/src/protocol-parser/protocol-parser.ts
@@ -217,6 +217,8 @@ export namespace CommonDataTypes {
 
   // BrowsingContext = text;
   export const BrowsingContextSchema = zod.string();
+
+  export const MaxDepthSchema = zod.number().int().nonnegative().max(MAX_INT);
 }
 
 /** @see https://w3c.github.io/webdriver-bidi/#module-script */
@@ -306,7 +308,9 @@ export namespace Script {
 
   const ChannelPropertiesSchema = zod.object({
     channel: ChannelIdSchema,
-    maxDepth: zod.number().int().nonnegative().max(MAX_INT).optional(),
+    // TODO(#294): maxDepth: CommonDataTypes.MaxDepthSchema.optional(),
+    // See: https://github.com/w3c/webdriver-bidi/pull/361/files#r1141961142
+    maxDepth: zod.number().int().min(1).max(1).optional(),
     ownership: ResultOwnershipSchema.optional(),
   });
 
@@ -358,9 +362,7 @@ export namespace BrowsingContext {
   //   ?root: browsingContext.BrowsingContext,
   // }
   const GetTreeParametersSchema = zod.object({
-    // TODO(##294): maxDepth: zod.number().int().nonnegative().max(MAX_INT).optional(),
-    // See: https://github.com/w3c/webdriver-bidi/pull/361/files#r1141961142
-    maxDepth: zod.number().int().min(1).max(1).optional(),
+    maxDepth: CommonDataTypes.MaxDepthSchema.optional(),
     root: CommonDataTypes.BrowsingContextSchema.optional(),
   });
 
--- a/src/protocol-parser/protocol-parser.ts
+++ b/src/protocol-parser/protocol-parser.ts
@@ -302,14 +302,29 @@ export namespace Script {
     return parseObject(params, DisownParametersSchema);
   }
 
+  const ChannelIdSchema = zod.string();
+
+  const ChannelPropertiesSchema = zod.object({
+    channel: ChannelIdSchema,
+    maxDepth: zod.number().int().nonnegative().max(MAX_INT).optional(),
+    ownership: ResultOwnershipSchema.optional(),
+  });
+
+  export const ChannelSchema = zod.object({
+    type: zod.literal('channel'),
+    value: ChannelPropertiesSchema,
+  });
+
   // ArgumentValue = (
   //   RemoteReference //
-  //   LocalValue
+  //   LocalValue //
+  //   script.Channel
   // );
   const ArgumentValueSchema = zod.union([
     CommonDataTypes.RemoteReferenceSchema,
     CommonDataTypes.SharedReferenceSchema,
     CommonDataTypes.LocalValueSchema,
+    Script.ChannelSchema,
   ]);
 
   // CallFunctionParameters = {
@@ -334,19 +349,6 @@ export namespace Script {
   ): ScriptTypes.CallFunctionParameters {
     return parseObject(params, CallFunctionParametersSchema);
   }
-
-  const ChannelIdSchema = zod.string();
-
-  const ChannelPropertiesSchema = zod.object({
-    channel: ChannelIdSchema,
-    maxDepth: zod.number().int().nonnegative().max(MAX_INT).optional(),
-    ownership: ResultOwnershipSchema.optional(),
-  });
-
-  export const ChannelSchema = zod.object({
-    type: zod.literal('channel'),
-    value: ChannelPropertiesSchema,
-  });
 }
 
 /** @see https://w3c.github.io/webdriver-bidi/#module-browsingContext */
@@ -356,7 +358,9 @@ export namespace BrowsingContext {
   //   ?root: browsingContext.BrowsingContext,
   // }
   const GetTreeParametersSchema = zod.object({
-    maxDepth: zod.number().int().nonnegative().max(MAX_INT).optional(),
+    // TODO(##294): maxDepth: zod.number().int().nonnegative().max(MAX_INT).optional(),
+    // See: https://github.com/w3c/webdriver-bidi/pull/361/files#r1141961142
+    maxDepth: zod.number().int().min(1).max(1).optional(),
     root: CommonDataTypes.BrowsingContextSchema.optional(),
   });
 
--- a/tests/test_channel.py
+++ b/tests/test_channel.py
@@ -15,15 +15,14 @@
 
 import pytest
 from anys import ANY_STR
-from test_helpers import execute_command, read_JSON_message, subscribe
+from test_helpers import (execute_command, read_JSON_message,
+                          send_JSON_command, subscribe)
 
 
-# TODO(#294): Implement E2E tests.
 @pytest.mark.asyncio
-async def test_channel(websocket, context_id):
+async def test_channel_twoMessageEvents(websocket, context_id):
     await subscribe(websocket, "script.message")
 
-    # The channel was successfully created if there's no thrown exception.
     await execute_command(
         websocket,
         {
@@ -52,7 +51,6 @@ async def test_channel(websocket, context_id):
         })
 
     resp = await read_JSON_message(websocket)
-
     assert resp == {
         "method": "script.message",
         "params": {
@@ -69,7 +67,6 @@ async def test_channel(websocket, context_id):
     }
 
     resp = await read_JSON_message(websocket)
-
     assert resp == {
         "method": "script.message",
         "params": {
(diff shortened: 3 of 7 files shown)


src/bidiMapper/domains/protocol/bidiProtocolTypes.ts
```diff
@@ -97,12 +97,20 @@ export namespace CommonDataTypes {
   });
   export type RemoteReference = zod.infer<typeof RemoteReferenceSchema>;
 
+  // BindingValue = {
+  //   type: "binding",
```
sadym-chromium, line 101: PROTO.binding

thiagowfx:
Planning to resume this PR this week.

thiagowfx:
Q: What is the difference between protocol-parser.ts and protocol.ts?

Q: Some definitions have both schema + event type with `zod.infer`. Others don't have schema. What's the difference?

sadym-chromium:
> Q: What is the difference between protocol-parser.ts and protocol.ts?

https://github.com/GoogleChromeLabs/chromium-bidi/pull/350

> Q: Some definitions have both schema + event type with `zod.infer`. Others don't have schema. What's the difference?

Can you give a specific example?

thiagowfx:
Side note: discussed off-line.

src/bidiMapper/domains/protocol/bidiProtocolTypes.ts
thiagowfx: Done

thiagowfx:
Reference: https://chromedevtools.github.io/devtools-protocol/tot/Runtime/#method-callFunctionOn

sadym-chromium:
> @sadym-chromium I continued the design doc but now I am stuck. More specifically:
> 

> 2. What should `sendMessage` do?

`sendMessage` should add message to the Binding's message queue.

> 1. What should `getMessage` do?

`getMessage` is used for long-poll messages to Mapper from the page. It returns either a value from the message queue, if it's non-empty, or a promise, which will be resolved after a message appears in the queue.

> 3. What should the deserialization (the `case`) do?

During deserialization, you should create a binding object on the page side, store it's `objectId` (will be required to poll messages), and provide its `sendMessage` to the user script as an argument.

> 4. Do we have a definition of "Binding" anywhere?  https://w3c.github.io/webdriver-bidi/ contains nothing. I'd like to document it in the newly added class.

It's a CDP term: https://chromedevtools.github.io/devtools-protocol/tot/Runtime/#method-addBinding. In BiDi it will probably called in another way:
* https://github.com/w3c/webdriver-bidi/issues/157
* https://github.com/w3c/webdriver-bidi/pull/361

> 5. When/where should I emit the `script.bindingCalled` event?

Script Evaluator should start long-polling `getMessage()`, and send `script.bindingCalled` event for each message.

> 6. What should the E2E test look like? Should it be a call to `script.evaluate`? If yes, that should it contain?

There is no way to pass argument to `script.evaluate`. The simplest e2e test should call `script.callFunction` with binding argument and with some script invoking that binding. Then you should verify the events are received by the Mapper. Additionally I can think of events order verification, and serialization.

sadym-chromium (commented):
comments after 1:1

thiagowfx:
@sadym-chromium PTAL. I wrote 3 `TODO(BLOCKER)` comments.

sadym-chromium:
Hope you don't mind me changing the PR name, as the original was misleading.

src/bidiMapper/domains/script/scriptEvaluator.ts
```diff
@@ -316,180 +388,175 @@ export async function stringifyObject(
   return stringifyResult.result.value;
 }
 
-export class ScriptEvaluator {
```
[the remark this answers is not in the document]
thiagowfx: I needed to make one method global within this file, that's why I eliminated the class. It was outside of the scope where I needed to call it otherwise.

src/bidiMapper/domains/script/scriptEvaluator.ts
```diff
@@ -242,6 +242,78 @@ async function deserializeToCdpArg(
...
+            {
+              value: bindingName,
+            },
+          ],
+          returnByValue: false,
+          executionContextId: realm.executionContextId,
+          generateWebDriverValue: false,
+        }
+      );
+      const bindingHandle = argEvalResult.result.objectId;
```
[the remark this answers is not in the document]
thiagowfx: Note: Long-poll should live here.

src/bidiMapper/domains/script/scriptEvaluator.ts
```diff
@@ -242,6 +242,78 @@ async function deserializeToCdpArg(
...
+                  // No other cases in scriptEvaluator are currently sending any event.
+                  // sendBiDiEvent(
+                  //   'script.bindingCalled',
+                  //   serializeCdpObject(message.result, 'root', realm)
+                  // );
+                }
+              },
+              /** Adds a message to the queue. */
+              sendMessage(message: string) {
+                queue.push(message);
```
sadym-chromium, line 278: It should add message to queue, and resolve Promise returned by `getMessage`

thiagowfx: Promise is only needed/used iff the queue is empty.

thiagowfx: `message.result` -> `{name: bindingName, value: message.result}`.

tests/test_binding.py
```diff
@@ -0,0 +1,47 @@
...
+                }],
+                "target": {
+                    "context": context_id
+                },
+                "awaitPromise": False,
+                "resultOwnership": "root"
+            }
+        })
+
+    assert {
```
thiagowfx, line 41: (binding)=> <<<< setTimeout(binding('SOME_MESSAGE')) >>> (racy / flaky --> delay by ~1ms for now)

src/bidiMapper/domains/script/scriptEvaluator.ts
```diff
@@ -242,6 +242,78 @@ async function deserializeToCdpArg(
...
+
+      const sendMessageArgResult = await realm.cdpClient.sendCommand(
+        'Runtime.callFunctionOn',
+        {
+          functionDeclaration: String(
+            (bindingHandle: {sendMessage: (message: string) => void}) => {
+              return bindingHandle.sendMessage;
+            }
+          ),
+          awaitPromise: true,
```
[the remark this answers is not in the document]
thiagowfx: Then I'll just delete all of them.

src/bidiMapper/domains/script/scriptEvaluator.ts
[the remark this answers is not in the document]
thiagowfx: This is not applicable anymore given the recent changes. Resolving.

thiagowfx:
@sadym-chromium Left two more blockers in the code. PTAL.

src/bidiMapper/domains/script/scriptEvaluator.ts
```diff
@@ -317,29 +63,317 @@ export async function stringifyObject(
...
+              }),
+              arguments: [
+                {
+                  value: bindingHandle,
+                },
+              ],
+              generateWebDriverValue: true,
+            }
+          );
+          // TODO(BLOCKER): I am not sure how to inject the eventManager here. This is starting to turn into spaghetti code.
```
thiagowfx, line 334: We should inject the eventManager from realm.ts, but even `Realm` does not contain eventManager.

How do we inject the eventManager into realm?

src/bidiMapper/domains/script/scriptEvaluator.ts
```diff
@@ -317,29 +63,317 @@ export async function stringifyObject(
...
+            }),
+            returnByValue: false,
+            executionContextId: realm.executionContextId,
+            generateWebDriverValue: false,
+          }
+        );
+        const bindingHandle = argEvalResult.result.objectId;
+
+        // Long-poll the message queue.
+        // TODO(BLOCKER): When do we break from the loop?
```
thiagowfx, line 318: We will never return `sendMessage` if we don't break here.

src/bidiMapper/domains/script/scriptEvaluator.ts
sadym-chromium: Realm is created by BrowsingContextImpl, so you can add EventManager to Realm's constructor.

src/bidiMapper/domains/script/scriptEvaluator.ts
sadym-chromium: you can run async code without awaiting for the result:
```suggestion
     (async ()=>{
       while(true) {
         await cdp.callFunctionOn({
             functionDeclaration: "this.getMessage()", 
             objectId: bindingHandle,
             awaitPromise: true
         });
       }
     })()
```

thiagowfx: Done

src/bidiMapper/domains/script/scriptEvaluator.ts
thiagowfx: Yup, done.

tests/test_binding.py
thiagowfx: Done

src/bidiMapper/domains/script/scriptEvaluator.ts
```diff
@@ -317,29 +69,332 @@ export async function stringifyObject(
...
+              'Runtime.callFunctionOn',
+              {
+                functionDeclaration: String(async (bindingHandle: any) => {
+                  return bindingHandle.getMessage();
+                }),
+                arguments: [
+                  {
+                    value: bindingHandle,
+                  },
+                ],
```
sadym-chromium, line 342: awaitPromise: true

thiagowfx:
The last remaining AI/blocker is to pass the bidi event with the proper arguments.

thiagowfx:
From Maksim, off-line:

> Event Manager accepts an object, which represents the event. To receive it on the bidi client, you should subscribe to it
> To pass arguments do the same thing script.evaluate does
> CDP to bidi value or something like that
> And add it to property "payload" or "value" of the object you pass to event manager

I don't understand this @sadym-chromium. EventManager has exactly 4 public methods:

- subscribe (N/A)
- unsubscribe (N/A)
- registerEvent
- registerPromiseEvent

I suposse I need to use the last two. The first argument is `EventMessage` --> the `script.bindingCalled` event. The second argument is the context id. There are no more arguments.

**How can I pass an argument to the event**?

sadym-chromium:
> From Maksim, off-line:
> 
> > Event Manager accepts an object, which represents the event. To receive it on the bidi client, you should subscribe to it
> > To pass arguments do the same thing script.evaluate does
> > CDP to bidi value or something like that
> > And add it to property "payload" or "value" of the object you pass to event manager
> 
> I don't understand this @sadym-chromium. EventManager has exactly 4 public methods:
> 
> * subscribe (N/A)
> * unsubscribe (N/A)
> * registerEvent
> * registerPromiseEvent
> 
> I suposse I need to use the last two. The first argument is `EventMessage` --> the `script.bindingCalled` event. The second argument is the context id. There are no more arguments.
> 
> **How can I pass an argument to the event**?

1. Extend the [`type EventMessage`](https://github.com/GoogleChromeLabs/chromium-bidi/blob/8c1d7b000f39af4fe62e2e4c0fcecb51bfa1a50c/src/protocol/protocol.ts#L59) with `Script.message` (same way it as [`LogEntryAddedEvent`](https://github.com/GoogleChromeLabs/chromium-bidi/blob/8c1d7b000f39af4fe62e2e4c0fcecb51bfa1a50c/src/protocol/protocol.ts#L796)). You can use [this PR](https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/361.html#event-script-message) as a reference for data structure. I don't believe it would change drastically.
2. Create an instance of `Script.message` and add the payload (translated from CDP to BiDi) into field `data`.
3. Call [`EventManager.registerEvent()`](https://github.com/GoogleChromeLabs/chromium-bidi/blob/8c1d7b000f39af4fe62e2e4c0fcecb51bfa1a50c/src/bidiMapper/domains/events/EventManager.ts#L125) with this message.

thiagowfx:
To double check: `Script.Message` does not currently exist in our codebase, correct? Should it be introduced in this PR as well?

thiagowfx:
```
 script.Message = {
  method: "script.message",
  params: script.MessageParameters
}

script.MessageParameters = {
  channel: script.Channel,
  data: RemoteValue,
  source: script.Source,
}
```

What should `channel` and `source` be set to?

sadym-chromium:
> To double check: `Script.Message` does not currently exist in our codebase, correct? Should it be introduced in this PR as well?

Yes. `Script.Message` event should be a part of this PR.

sadym-chromium:
> What should `channel` and `source` be set to?

`channel` is a `binding.value`.

`source` is the [source from where the message came from](https://w3c.github.io/webdriver-bidi/#type-script-Source).

thiagowfx:
This PR is getting too complicated and hitting too many rebase conflicts as time passes. I will start to split it into smaller ones. Start with https://github.com/GoogleChromeLabs/chromium-bidi/pull/541 to define `Script.Message`.

thiagowfx:
With https://github.com/GoogleChromeLabs/chromium-bidi/pull/542 the diff should be less noisier now.

thiagowfx:
PTAL.

In particular, I am always getting this error (e.g. when running E2E test, and when running `HEADLESS=true ./runWPT.sh webdriver/tests/bidi/script/call_function/invalid.py` WPT test):

```
 webdriver.bidi.error.UnknownErrorException: unknown error (Cannot read properties of undefined (reading '_parseSync'))
```

src/protocol-parser/protocol-parser.ts
```diff
@@ -310,6 +310,7 @@ export namespace Script {
     CommonDataTypes.RemoteReferenceSchema,
     CommonDataTypes.SharedReferenceSchema,
     CommonDataTypes.LocalValueSchema,
+    Script.ChannelSchema,
```
sadym-chromium, line 313: `Script.ChannelSchema` is not set at this point. Either move `Script.ChannelSchema` above this declaration, or use `zod.lazy(() => ...`.

sadym-chromium:
> PTAL.
> 
> In particular, I am always getting this error (e.g. when running E2E test, and when running `HEADLESS=true ./runWPT.sh webdriver/tests/bidi/script/call_function/invalid.py` WPT test):
> 
> ```
>  webdriver.bidi.error.UnknownErrorException: unknown error (Cannot read properties of undefined (reading '_parseSync'))
> ```
> 
> Minimal repro: The error happens just with the two-line change in `protocol.ts` and `protocol-parser.ts`.

I pushed the fix. [Reason described here](https://github.com/GoogleChromeLabs/chromium-bidi/pull/319#discussion_r1139967444).

src/protocol-parser/protocol-parser.ts
thiagowfx: Done

thiagowfx:
> I pushed the fix. https://github.com/GoogleChromeLabs/chromium-bidi/pull/319#discussion_r1139967444.

Indeed, this fixes the WPT tests. Now we just need to fix the E2E tests.

thiagowfx:
E2E test is now passing. Please deeply review the implementation.

Is there some WPT test associated with this feature?

src/bidiMapper/domains/script/scriptEvaluator.ts
```diff
@@ -251,154 +397,129 @@ export class ScriptEvaluator {
...
+                 * in the queue.
+                 */
+                async getMessage(): Promise<string> {
+                  if (queue.length > 0) {
+                    return queue.shift()!;
+                  }
+                  const promise = new Promise<string>((resolve) => {
+                    resolver = resolve;
+                  });
+                  promise.then(() => queue.shift());
```
sadym-chromium, line 426:
```suggestion
                  promise.then(() => {
                      resolver = null;
                      return queue.shift();
                  );
```

src/bidiMapper/domains/script/scriptEvaluator.ts
```diff
@@ -251,154 +397,129 @@ export class ScriptEvaluator {
...
 
-      default:
-        throw new Error(
-          `Value ${JSON.stringify(argumentValue)} is not deserializable.`
+        const argEvalResult = await realm.cdpClient.sendCommand(
+          'Runtime.callFunctionOn',
+          {
+            functionDeclaration: String(() => {
+              const queue: string[] = [];
+              let resolver: (message: string | PromiseLike<string>) => void;
```
sadym-chromium, line 412:
```suggestion
              let resolver: ((message: string | PromiseLike<string>) => void) | null = null;
```

src/bidiMapper/domains/script/scriptEvaluator.ts
```diff
@@ -251,154 +397,129 @@ export class ScriptEvaluator {
...
+                /**
+                 * Adds a message to the queue.
+                 * If the queue is initially empty, resolve the pending promise.
+                 */
+                sendMessage(message: string) {
+                  queue.push(message);
+                  if (queue.length === 1) {
+                    resolver(message);
+                  }
+                },
```
sadym-chromium, line 439:
```suggestion
                sendMessage(message: string) {
                  queue.push(message);
                  if (resolver !== null) {
                    resolver();
                  }
                },
```

src/bidiMapper/domains/script/scriptEvaluator.ts
```diff
@@ -251,154 +397,129 @@ export class ScriptEvaluator {
...
+                async getMessage(): Promise<string> {
+                  if (queue.length > 0) {
+                    return queue.shift()!;
+                  }
+                  const promise = new Promise<string>((resolve) => {
+                    resolver = resolve;
+                  });
+                  promise.then(() => queue.shift());
+                  return promise;
+                },
```
sadym-chromium, line 428: This change is required for option **1** _`sendMessage` should send message via `queue`_.
```suggestion
```

```suggestion
                async getMessage(): Promise<string> {
                  if (queue.length > 0) {
                    resolver = null;
                    return queue.shift()!;
                  }
                  const promise = new Promise<string>((resolve) => {
                    resolver = resolve;
                  });
                  promise.then(getMessage);
                  return promise;
                },
```

src/bidiMapper/domains/script/scriptEvaluator.ts
```diff
@@ -251,154 +397,129 @@ export class ScriptEvaluator {
...
-        executionContextId: realm.executionContextId,
-      });
-    return realm.cdpToBidiValue(cdpWebDriverValue, resultOwnership);
-  }
+        const channelHandle = argEvalResult.result.objectId;
+
+        // Long-poll the message queue asynchronously.
+        (async () => {
+          for (;;) {
+            const message = await realm.cdpClient.sendCommand(
```
sadym-chromium, line 452: Please add TODO to remove this loop after the realm is destroyed. You can rely on the CDP throwing exception in such a case.

src/bidiMapper/domains/script/scriptEvaluator.ts
```diff
@@ -251,154 +397,129 @@ export class ScriptEvaluator {
...
+              'Runtime.callFunctionOn',
+              {
+                functionDeclaration: String(async (channelHandle: any) => {
+                  return channelHandle.getMessage();
+                }),
+                arguments: [
+                  {
+                    value: channelHandle,
```
sadym-chromium, line 460:
```suggestion
                    objectId: channelHandle,
```

src/bidiMapper/domains/script/scriptEvaluator.ts
```diff
@@ -251,154 +397,129 @@ export class ScriptEvaluator {
...
-        arguments: [arg],
-        generateWebDriverValue: true,
-        executionContextId: realm.executionContextId,
-      });
-    return realm.cdpToBidiValue(cdpWebDriverValue, resultOwnership);
-  }
+        const channelHandle = argEvalResult.result.objectId;
+
+        // Long-poll the message queue asynchronously.
+        (async () => {
```
sadym-chromium, line 450: Please move to a dedicated private method

src/bidiMapper/domains/script/scriptEvaluator.ts
```diff
@@ -251,154 +397,129 @@ export class ScriptEvaluator {
...
+                  channel: {
+                    type: 'channel',
+                    value: channelName,
+                  },
+                  data: await this.serializeCdpObject(
+                    {
+                      type: 'string',
+                      value: message.result,
+                    },
+                    'root',
```
sadym-chromium, line 481: Please add `TODO` to add `ownership` param to channel.

src/bidiMapper/domains/script/scriptEvaluator.ts
```diff
@@ -251,154 +397,129 @@ export class ScriptEvaluator {
...
+                    realm
+                  ),
```
sadym-chromium, line 483:
```suggestion
                  data: await realm.cdpToBidiValue(message, 'root'),
```

      result: await realm.cdpToBidiValue(
        cdpCallFunctionResult,
        resultOwnership
      ),

src/protocol-parser/protocol-parser.ts
```diff
@@ -302,14 +302,29 @@ export namespace Script {
     return parseObject(params, DisownParametersSchema);
   }
 
+  const ChannelIdSchema = zod.string();
+
+  const ChannelPropertiesSchema = zod.object({
+    channel: ChannelIdSchema,
+    maxDepth: zod.number().int().nonnegative().max(MAX_INT).optional(),
```
sadym-chromium, line 309: FYI: https://github.com/w3c/webdriver-bidi/pull/361/files#r1141961142

src/bidiMapper/domains/script/scriptEvaluator.ts
```diff
@@ -251,154 +397,129 @@ export class ScriptEvaluator {
...
           }
         );
-        // TODO (#375): release `argEvalResult.result.objectId` after using.
+        // TODO(#375): Release `argEvalResult.result.objectId` after using.
         return {objectId: argEvalResult.result.objectId};
       }
 
-      // TODO(#375): dispose of nested objects.
+      case 'channel': {
+        const channelName = argumentValue.value;
```
sadym-chromium, line 405:
```suggestion
        const channelName = argumentValue.value.channel;
```

tests/test_channel.py
```diff
@@ -0,0 +1,44 @@
...
+@pytest.mark.asyncio
+async def test_channel(websocket, context_id):
+    # The channel was successfully created if there's no thrown exception.
+    await execute_command(
+        websocket,
+        {
+            "method": "script.callFunction",
+            "params": {
+                # A small delay is needed to avoid a race condition.
+                "functionDeclaration": "(binding) => setTimeout(() => binding('MY_MESSAGE'), 1)",
```
sadym-chromium, line 30: This tests the serialization, but we have to test message as well.

thiagowfx: TODO: subscribe to event (e.g. similar to network)

TODO: read next message, use read JSON helper

src/bidiMapper/domains/script/scriptEvaluator.ts
```diff
@@ -251,154 +397,122 @@ export class ScriptEvaluator {
...
+        const channelName = argumentValue.value.channel;
 
-      default:
-        throw new Error(
-          `Value ${JSON.stringify(argumentValue)} is not deserializable.`
+        const argEvalResult = await realm.cdpClient.sendCommand(
+          'Runtime.callFunctionOn',
+          {
+            functionDeclaration: String(() => {
+              const queue: string[] = [];
```
sadym-chromium, line 411: As discussed offline, there can be any value
```suggestion
              const queue: any[] = [];
```

src/bidiMapper/domains/script/scriptEvaluator.ts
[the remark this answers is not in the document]
thiagowfx: Done

src/bidiMapper/domains/script/scriptEvaluator.ts
```diff
@@ -251,154 +397,120 @@ export class ScriptEvaluator {
...
+                },
+
+                /**
+                 * Adds a message to the queue.
+                 * If the queue is initially empty, resolve the pending promise.
+                 */
+                sendMessage(message: string) {
+                  queue.push(message);
+                  if (queue.length === 1) {
+                    resolver(message);
```
sadym-chromium, line 435: NIT: Let's set `resolver` to `null` as well.

src/bidiMapper/domains/script/scriptEvaluator.ts
```diff
@@ -251,154 +397,120 @@ export class ScriptEvaluator {
...
+            functionDeclaration: String(() => {
+              const queue: any[] = [];
+              let resolver: (message: string | PromiseLike<string>) => void;
+
+              return {
+                /**
+                 * Gets a promise, which is resolved as soon as a message occurs
+                 * in the queue.
+                 */
+                async getMessage(): Promise<string> {
```
sadym-chromium, line 419:
```suggestion
               getMessage(): Promise<any> {
```

src/bidiMapper/domains/script/scriptEvaluator.ts
```diff
@@ -251,154 +397,123 @@ export class ScriptEvaluator {
...
+
+            this.#eventManager.registerEvent(
+              {
+                method: Script.EventNames.MessageEvent,
+                params: {
+                  channel: channelName,
+                  data: await realm.cdpToBidiValue(
+                    message,
+                    argumentValue.value.ownership ?? 'none'
+                  ),
```
sadym-chromium, line 477: ^ because of this await, the event orders can be messed, if used with `registerEvent`.

src/bidiMapper/domains/script/scriptEvaluator.ts
```diff
@@ -251,154 +397,123 @@ export class ScriptEvaluator {
...
+                    objectId: channelHandle,
+                  },
+                ],
+                awaitPromise: true,
+                executionContextId: realm.executionContextId,
+                generateWebDriverValue: true,
+              }
+            );
+
+            this.#eventManager.registerEvent(
```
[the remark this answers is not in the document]
thiagowfx: Actually, I see what you mean. To use the promise on `data`, correct? PTAL.

src/bidiMapper/domains/script/scriptEvaluator.ts
thiagowfx: Done

src/bidiMapper/domains/script/scriptEvaluator.ts
thiagowfx: Done

src/bidiMapper/domains/script/scriptEvaluator.ts
thiagowfx: Done

thiagowfx:
There's one open conversation but I cannot resolve it as it refers to an old(er) commit. :fail:

thiagowfx:
The only AI missing here is to augment the E2E test. Will do so by EOD.

thiagowfx:
> The only AI missing here is to augment the E2E test. Will do so by EOD.

Done.

thiagowfx:
Modulo pre-refactoring to make the diff easier, this PR is now fully complete.

src/bidiMapper/domains/script/scriptEvaluator.ts
```diff
@@ -17,16 +17,74 @@
 import {Protocol} from 'devtools-protocol';
 
 import {CommonDataTypes, Message, Script} from '../../../protocol/protocol.js';
+import {IEventManager} from '../events/EventManager.js';
 
 import {Realm} from './realm.js';
 
+async function longPoll(
```
thiagowfx, line 24: I welcome suggestions for a better name for this method, instead of `longPoll`.

sadym-chromium: `initChannelListener`?

src/bidiMapper/domains/script/scriptEvaluator.ts
```diff
@@ -17,16 +17,74 @@
...
+  // TODO(#294): Remove this loop after the realm is destroyed.
+  // Rely on the CDP throwing exception in such a case.
+  for (;;) {
+    const message = await realm.cdpClient.sendCommand(
+      'Runtime.callFunctionOn',
+      {
+        functionDeclaration: String(
+          async (channelHandle: {
+            getMessage: () => Promise<unknown>;
+            sendMessage: (message: string) => void;
```
sadym-chromium, line 41: NIT: Listener doesn't need to know about `sendMessage` method.
```suggestion
```

src/bidiMapper/domains/script/scriptEvaluator.ts
```diff
@@ -394,6 +452,74 @@ export class ScriptEvaluator {
         return {objectId: argEvalResult.result.objectId};
       }
 
+      case 'channel': {
+        const argEvalResult = await realm.cdpClient.sendCommand(
```
sadym-chromium, line 456: `argEvalResult` -> `createChannelHandleResult` ?

src/bidiMapper/domains/script/scriptEvaluator.ts
```diff
@@ -394,6 +452,74 @@ export class ScriptEvaluator {
...
+                      : new Promise<void>((resolve) => {
+                          queueNonEmptyResolver = resolve;
+                        });
+                  await onMessage;
+                  return queue.shift();
+                },
+
+                /**
+                 * Adds a message to the queue.
+                 * If the queue was initially empty, resolve the pending promise.
```
sadym-chromium, line 481:
```suggestion
                 * Resolves the pending promise if needed.
```

src/protocol-parser/protocol-parser.ts
```diff
@@ -302,14 +304,31 @@ export namespace Script {
     return parseObject(params, DisownParametersSchema);
   }
 
+  const ChannelIdSchema = zod.string();
+
+  const ChannelPropertiesSchema = zod.object({
+    channel: ChannelIdSchema,
+    // TODO(#294): maxDepth: CommonDataTypes.MaxDepthSchema.optional(),
+    // See: https://github.com/w3c/webdriver-bidi/pull/361/files#r1141961142
+    maxDepth: zod.number().int().min(1).max(1).optional(),
```
sadym-chromium, line 313:
```suggestion
    maxDepth: MaxDepthSchema,
```

tests/test_channel.py
```diff
@@ -0,0 +1,156 @@
...
+@pytest.mark.asyncio
+async def test_channel_twoMessageEvents(websocket, context_id):
+    await subscribe(websocket, "script.message")
+
+    await execute_command(
+        websocket,
+        {
+            "method": "script.callFunction",
+            "params": {
+                # A small delay is needed to avoid a race condition.
```
sadym-chromium, line 30: Can we test this instead of this comment? Like
```
send(MESSAGE_BEFORE_EXEC_FINISHED); 
setTimeout(send(MESSAGE_AFTER_EXEC_FINISHED))
```

tests/test_channel.py
```diff
@@ -0,0 +1,156 @@
...
+             'value': 'x'
+         }], ['b', {
+             'type': 'string',
+             'value': 'y'
+         }]]
+     }), ("() => {}", {
+         'handle': ANY_STR,
+         'type': 'function'
+     })])
+async def test_channel_complexTypes(test_input, expected, websocket,
```
sadym-chromium, line 114: Optional: actually, this tests can be moved to `test_serialization`.

src/bidiMapper/domains/script/scriptEvaluator.ts
thiagowfx: Done

src/bidiMapper/domains/script/scriptEvaluator.ts
thiagowfx: Done

src/bidiMapper/domains/script/scriptEvaluator.ts
```diff
@@ -251,154 +397,130 @@ export class ScriptEvaluator {
...
-        awaitPromise: false,
-        arguments: [arg],
-        generateWebDriverValue: true,
-        executionContextId: realm.executionContextId,
-      });
-    return realm.cdpToBidiValue(cdpWebDriverValue, resultOwnership);
-  }
+        const channelHandle = argEvalResult.result.objectId;
+
+        // Long-poll the message queue asynchronously.
```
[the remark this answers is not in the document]
sadym-chromium: LGTM

sadym-chromium (approved):

src/bidiMapper/domains/script/scriptEvaluator.ts
```diff
@@ -17,16 +17,72 @@
 import {Protocol} from 'devtools-protocol';
 
 import {CommonDataTypes, Message, Script} from '../../../protocol/protocol.js';
+import {IEventManager} from '../events/EventManager.js';
 
 import {Realm} from './realm.js';
 
+async function initChannelListener(
```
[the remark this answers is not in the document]
thiagowfx: Done via https://github.com/GoogleChromeLabs/chromium-bidi/pull/567

PR #321, opened 2022-11-23
Enable strict TypeScript configs

OrKoN:
Proposal to enable strict(er) TypeScript configuration which helps catching unused variables and other things.


src/index.ts
```diff
@@ -32,7 +32,7 @@ function parseArguments() {
   parser.add_argument('-p', '--port', {
     help: 'Port that BiDi server should listen to. Default is 8080.',
     type: 'int',
-    default: process.env.PORT || 8080,
+    default: process.env['PORT'] || 8080,
```
thiagowfx, line 35: Does this make any difference or is it a no-op?

OrKoN: The runtime behaviour is the same but from the types perspective, there is a difference because the env is defined as `{[key: string]: string|undefined}` instead of `{PORT: string|undefined}` because TS does not know which property names can be expected on the env.

PR #322, opened 2022-11-23
Simplify CDP client

OrKoN:
Better type-safety, less memory and network used and more aligned with the CDP client in Puppeteer, no need for custom scripts. WDYT?


sadym-chromium:
Makes sense, even though the current version is more convinient for the developers than the proposed one.

sadym-chromium (approved):

thiagowfx (approved):

PR #326, opened 2022-11-25
Separate mapper tab from the mapper

OrKoN:
This PR separates the mapper tab from the mapper implementation. This allows compiling the mapper implementation without DOM APIs. As the next step we can turn bidiMapper/bidiTab/bidiServer into composite projects to reduce duplication in the build output.

### DIFF
--- a/src/bidiMapper/domains/context/browsingContextImpl.ts
+++ b/src/bidiMapper/domains/context/browsingContextImpl.ts
@@ -491,9 +491,9 @@ export class BrowsingContextImpl {
   ): Promise<BrowsingContext.PROTO.FindElementResult> {
     await this.#targetDefers.targetUnblocked;
 
-    const functionDeclaration = String((resultsSelector: string) =>
-      document.querySelector(resultsSelector)
-    );
+    const functionDeclaration = `
+      (resultsSelector) => document.querySelector(resultsSelector)
+    `;
     const _arguments: Script.ArgumentValue[] = [
       {type: 'string', value: selector},
     ];
--- a/.github/workflows/unit.js.yml
+++ b/.github/workflows/unit.js.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Build
         run: npm run build
       - name: Verify mapper size
-        run: du -sh lib/iife/bidiMapper/mapper.js
+        run: du -sh lib/iife/mapperTab.js
       - name: Run unit tests
         run: npm run unit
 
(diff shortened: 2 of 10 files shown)


src/bidiMapper/domains/context/browsingContextImpl.ts
```diff
@@ -491,9 +491,9 @@ export class BrowsingContextImpl {
   ): Promise<BrowsingContext.PROTO.FindElementResult> {
     await this.#targetDefers.targetUnblocked;
 
-    const functionDeclaration = String((resultsSelector: string) =>
-      document.querySelector(resultsSelector)
-    );
+    const functionDeclaration = `
+      (resultsSelector) => document.querySelector(resultsSelector)
+    `;
```
sadym-chromium, line 496: Did the previous version stop working? Because it allowed to provide TS with all the compilation checks.

sadym-chromium:
I'm not sure how are you going to use Mapper afterwards. Currently, Tab is created by Mapper. So if you plan to use Mapper without tab, this dependency should be inverted.

OrKoN:
>  I'm not sure how are you going to use Mapper afterwards. Currently, Tab is created by Mapper. So if you plan to use Mapper without tab, this dependency should be inverted.

I don't think it's created by the Mapper here. it's now created by the src/bidiTab/bidiTab.ts which is the new entrypoint for the "mapper".

sadym-chromium:
> So the mapper + the tab is in `lib/iife/mapper.js`, in Puppeteer we will consume `lib/cjs/bidiMapper/*` that does not contain the DOM APIs and also does not create the tab. `lib/cjs/bidiTab/*` is the target that contains the BiDi tab code (and bidiMapper as a dependency) and this target is what gets rolled up into `lib/iife/mapper.js`.

Would you mind to rename the `lib/iife/mapper.js` into something like `lib/iife/mapperTab.js` then?

OrKoN:
> > So the mapper + the tab is in `lib/iife/mapper.js`, in Puppeteer we will consume `lib/cjs/bidiMapper/*` that does not contain the DOM APIs and also does not create the tab. `lib/cjs/bidiTab/*` is the target that contains the BiDi tab code (and bidiMapper as a dependency) and this target is what gets rolled up into `lib/iife/mapper.js`.
> 
> Would you mind to rename the `lib/iife/mapper.js` into something like `lib/iife/mapperTab.js` then?

done!

sadym-chromium (approved):

PR #336, opened 2022-11-30
Use Google's TS style guide's format config

thiagowfx:
Mirror puppeteer's approach ref.
https://github.com/puppeteer/puppeteer/pull/8542

### DIFF
--- a/.prettierrc.cjs
+++ b/.prettierrc.cjs
@@ -14,11 +14,9 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+const gtsPrettierrc = require('gts/.prettierrc.json');
 
 module.exports = {
-  bracketSpacing: false, // clang-format does not do this
-  semi: true, // https://google.github.io/styleguide/jsguide.html#disallowed-features-automatic-semicolon-insertion
-  singleQuote: true, // https://google.github.io/styleguide/jsguide.html#features-strings-use-single-quotes
-  trailingComma: 'es5',
-  useTabs: false, // https://google.github.io/styleguide/jsguide.html#whitespace-characters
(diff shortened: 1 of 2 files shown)


.prettierrc.cjs
```diff
@@ -14,11 +14,9 @@
...
+const gtsPrettierrc = require('gts/.prettierrc.json');
 
 module.exports = {
-  bracketSpacing: false, // clang-format does not do this
-  semi: true, // https://google.github.io/styleguide/jsguide.html#disallowed-features-automatic-semicolon-insertion
-  singleQuote: true, // https://google.github.io/styleguide/jsguide.html#features-strings-use-single-quotes
-  trailingComma: 'es5',
-  useTabs: false, // https://google.github.io/styleguide/jsguide.html#whitespace-characters
+  ...gtsPrettierrc,
+  arrowParens: 'always', // https://google.github.io/styleguide/jsguide.html#features-functions-arrow-functions
```
thiagowfx, line 21: This is needed because upstream `gts` uses `arrowParens: 'avoid'` by default which is undesirable. I've flagged this internally via https://docs.google.com/document/d/1XdNFGdGnCXS1MPjQphxI2ztI_G6oHZWswot6NUlvHb0/edit?disco=AAAAjxTJS5Y

sadym-chromium (approved):

package.json
```diff
@@ -32,6 +32,7 @@
     "argparse": "^2.0.1",
     "debug": "^4.3.4",
     "devtools-protocol": "^0.0.1007249",
+    "gts": "^3.1.1",
```
OrKoN, line 35: this should be a dev dependency

thiagowfx: good call

PR #341, opened 2022-12-02
Support python 3.10

sadym-chromium:
Bump dependencies to support Python 3.10


tests/requirements.txt
```diff
@@ -1,3 +1,3 @@
-pytest===6.1.2
-pytest-asyncio==0.14.0
-websockets==8.1
+pytest==7.2.0
```
thiagowfx, line 1: Should we update examples/requirements.txt as well?

sadym-chromium: updated

PR #347, opened 2022-12-05
Backup to `monospace` font in Chromium

sadym-chromium:
As long as Chromium doesn't support some fonts, add rule to fall back to simple `monospace` if `Menlo` is not present.


thiagowfx:
I would suggest to use this instead: https://systemfontstack.com/

```
font-family: Menlo, Consolas, Monaco, Liberation Mono, Lucida Console, monospace;
```

This is friendly to all 3 major OSes (Windows, Linux, macOS) and does not require extra fonts, and still has `monospace` in case it's accessed from yet another OS (e.g. some *BSD).

If you want some more context I write a blog post about this a while ago :D https://thiagowfx.github.io/2022/02/website-fonts-just-use-the-system-font-stack/

sadym-chromium:
> I would suggest to use this instead: https://systemfontstack.com/
> 
> ```
> font-family: Menlo, Consolas, Monaco, Liberation Mono, Lucida Console, monospace;
> ```
> 
> This is friendly to all 3 major OSes (Windows, Linux, macOS) and does not require extra fonts, and still has `monospace` in case it's accessed from yet another OS (e.g. some *BSD).
> 
> If you want some more context I write a blog post about this a while ago :D https://thiagowfx.github.io/2022/02/website-fonts-just-use-the-system-font-stack/

Love it!

thiagowfx (approved):

PR #349, opened 2022-12-07
Introduce BidiTransport based on BiDi messages

OrKoN:
This PR refactors `ITransport` usages in the `BidiServer`class: a new class called `BidiTransport` is introduced that operates on `Message.RawCommandRequest` and `Message.OutgoingMessage` instead of strings. This allows implementing clients that always produce valid BiDi commands and expect valid BiDi events in return. It also allows the clients to avoid the overhead of `JSON.stringify/parse`. Transport level BiDi message validation is now moved to the `bidiTab` to `WindowBidiTransport` that implements `BidiTransport`.

This PR also moves the protocol types to a top-level protocol folder to clearly indicate which projects depend on protocol types.

Next steps:
1) Separate runtime wrapper classes (such as events) in `protocol/types.ts` from the types for that `protocol/types.ts` includes only TS types. This will allow to turn all imports for zod managed types to be `type` imports.
2) Separate the parsing helpers from the type definitions.


sadym-chromium (approved):

thiagowfx (approved):
LGTM for style, prefer Maksim's review for domain review at this time

PR #350, opened 2022-12-12
Separates parser from the types

OrKoN:
I tried to find other alternatives but for now it seems the best way is to hand-write TS types for BiDi. Those types should be used by the mapper. The bidiTab would inject the zod-based parser. The parser implements types using zod leading to some duplication. I think it should be okay because our future goal is to have TS types to be auto-generated from CDDL while the zod-based parser is likely to remain hand-written.


src/bidiMapper/domains/log/logHelper.ts
```diff
@@ -52,13 +52,21 @@ export function logMessageFormatter(
       if (token === '%s') {
         output += stringFromArg(arg);
       } else if (token === '%d' || token === '%i') {
-        if (['bigint', 'number', 'string'].includes(arg.type)) {
+        if (
```
OrKoN, line 55: I am not sure why but hand-written types are slightly different and TS could infer `includes` properly.

src/bidiMapper/domains/log/logManager.ts
```diff
@@ -68,7 +68,7 @@ export class LogManager {
         });
         const argsPromise: Promise<CommonDataTypes.RemoteValue[]> =
           realm === undefined
-            ? Promise.resolve(params.args)
+            ? Promise.resolve(params.args as CommonDataTypes.RemoteValue[])
```
OrKoN, line 71: @sadym-chromium could you advice? it looks like params.args where of the wrong type

sadym-chromium:
on it

src/bidiMapper/BidiServer.ts
```diff
@@ -38,7 +38,8 @@ export class BidiServer extends EventEmitter<BidiServerEvents> {
   private constructor(
     bidiTransport: BidiTransport,
     cdpConnection: CdpConnection,
-    selfTargetId: string
+    selfTargetId: string,
+    parser: BidiParser
```
sadym-chromium, line 42: Did you consider changing the `BidiTransport` interface so that it operates not JSON string commands - responses, but the typed objects? It would allow to move parser into the `BidiTransport` implementation if need, or send unvalidated objects directly in case of Puppeteer. It would allow to remove params parsing from the CommandProcessor.

src/bidiMapper/domains/log/logHelper.ts
sadym-chromium: ouch )

src/bidiMapper/domains/log/logManager.ts
sadym-chromium: Well, without realm we cannot do much, so just put to args whatever we got from CDP. So yes, this line should work as a backup.

sadym-chromium (approved):
LGTM with one suggestion of significant change. Feel free to ignore it.

src/bidiMapper/BidiServer.ts
OrKoN: BidiTransport operates on the typed objects already. It does not allow moving the parser though because of the channels :D

thiagowfx:
No worries, if @sadym-chromium is happy then it's all good, thanks!

PR #355, opened 2022-12-13
Ensure import extensions

OrKoN:
Removes ts-node because it does not work with .js extensions and we already have the test files compiled. Adds eslint to validate that imports have a .js extension. `.js` extensions is needed for compatibility with browsers and bundlers.


.gitignore
```diff
@@ -6,3 +6,4 @@ __pycache__
 logs/
 chromium-bidi.iml
 node_modules/
+chromium-bidi-*
```
sadym-chromium, line 9: NIT: no new line

package.json
```diff
@@ -2,22 +2,25 @@
...
+    "prepare": "npm run clean && npm run build",
+    "prettier": "npm run format:prettier",
+    "format": "npm run format:eslint && npm run format:prettier && npm run yapf",
+    "format:eslint": "eslint --ext js --ext ts --fix .",
+    "format:prettier": "prettier --write .",
     "rollup": "rollup --config configs/rollup.config.js",
     "server-no-build": "node lib/cjs/bidiServer/index.js",
     "server": "npm run build && npm run server-no-build --",
-    "test": "npm run unit",
+    "test": "npm run build && npm run unit",
```
sadym-chromium, line 23: No need in building here then: https://github.com/GoogleChromeLabs/chromium-bidi/blob/f8d953fa7313c562f88471865a4fd18dc81f1715/.github/workflows/unit.js.yml#L23

.eslintignore
```diff
@@ -0,0 +1,9 @@
+.build
+lib
+.cache
+.idea
+__pycache__
+logs/
+chromium-bidi.iml
+node_modules/
+chromium-bidi-*
```
sadym-chromium, line 9: What is `chromium-bidi-*` ?

NIT: no new line

sadym-chromium (approved):

.eslintignore
OrKoN: tar.gz files produced by `npm pack`

package.json
OrKoN: It requires building without ts-node to keep the same behaviour

Issue #353, opened 2022-12-13
Fix WPT tests after bump


sadym-chromium:
After [WPT bump](https://github.com/GoogleChromeLabs/chromium-bidi/pull/351/files), some WPT tests started failing.

There are 2 root causes:
1. Updated [node serialization](https://github.com/w3c/webdriver-bidi/pull/335):
    1. `wpt-metadata/webdriver/tests/bidi/script/call_function/result_node.py.ini`;
    2. `wpt-metadata/webdriver/tests/bidi/script/evaluate/result_node.py.ini`.
2. Updated WPT [`Session.subscribe`](https://github.com/web-platform-tests/wpt/pull/37400) tests:
    1. `wpt-metadata/webdriver/tests/bidi/session/subscribe`;
    2. `wpt-metadata/webdriver/tests/bidi/session/unsubscribe`.

sadym-chromium:
Closing in favour of:
1. #474
2. #482
3. #474

PR #359, opened 2022-12-16
Subscribe to the whole  domain

sadym-chromium:
Addressing https://github.com/GoogleChromeLabs/chromium-bidi/issues/353.

Implement subscription to the whole domain, not only specific event.


thiagowfx:
What is your concern with `EventNamesList`?

thiagowfx (approved):

PR #366, opened 2023-01-02
Add TSLint integration

thiagowfx:


thiagowfx:
Actually, TSLint was deprecated[1]. Closing.

[1]: https://stackoverflow.com/questions/65559369/difference-between-eslint-and-tslint

PR #377, opened 2023-01-11
Add output.name to rollup config to resolve warning

thiagowfx:
> (!) If you do not supply "output.name", you may not be able to access
> the exports of an IIFE bundle.

Tested: `npm run build` does not produce a warning

### DIFF
--- a/configs/rollup.config.js
+++ b/configs/rollup.config.js
@@ -22,6 +22,7 @@ import commonjs from '@rollup/plugin-commonjs';
 export default {
   input: 'lib/cjs/bidiTab/bidiTab.js',
   output: {
+    name: 'mapperTab',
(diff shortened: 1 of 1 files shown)


configs/rollup.config.js
```diff
@@ -22,6 +22,7 @@ import commonjs from '@rollup/plugin-commonjs';
 export default {
   input: 'lib/cjs/bidiTab/bidiTab.js',
   output: {
+    name: 'mapperTab',
```
OrKoN, line 25: do we need to access it? I think we do not?

thiagowfx: We do not need to access it, but the warning is noisy, the only intention of this commit is to silence it. I am also open to other ways of doing so

OrKoN (approved):

PR #387, opened 2023-01-23
pre-commit: add yapf integration

thiagowfx:


.style.yapf
```diff
@@ -1,2 +0,0 @@
-[style]
-based_on_style = pep8
```
thiagowfx, line 2: Deleted because this is the default already.

OrKoN (approved):

Issue #392, opened 2023-01-24
Reduce the number of static methods and variables


OrKoN:
Install https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-extraneous-class.md to detect completely static classes.

How to fix static:

1) if it's a readonly static property used as a const move it to the top module level `const MY_CONST = ..`. Export it if other modules need it.
2) if it's a property holding runtime data, the data should not be stored globally but be scoped to a bidi mapper session.
3) after the data is made local, the static methods should either become a module-level functions or non-static methods operating on the local data.

OrKoN:
The following are static data properties:

ScriptEvaluator.#knownHandlesToRealm
Realm.#realmMap

These should be scoped to src/bidiMapper/BidiServer.

thiagowfx:
![dependencygraph](https://user-images.githubusercontent.com/2840106/215522014-6c854882-165b-430f-b3bc-25e02e5d392d.svg)

Refer to https://github.com/GoogleChromeLabs/chromium-bidi/pull/422#issuecomment-1408841439 to see how this graph was generated.

thiagowfx:
@OrKoN regarding ScriptEvaluator.#knownHandlesToRealm and  Realm.#realmMap: should both of these also become instances of `BidiServer`?

Also: For `realm` you had mentioned earlier to create a `RealmStorage` wrapper. How about `ScriptEvaluator`?

OrKoN:
@thiagowfx yeah, I think the knownHandlesToRealm can be part of the realm storage as it is related to realms too.

PR #393, opened 2023-01-24
eslint: add no-extraneous-class rule and fix BrowsingContextStorage

thiagowfx:

Bug: #392

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -7,7 +7,9 @@ module.exports = {
   parser: '@typescript-eslint/parser',
   plugins: ['@typescript-eslint', 'import', 'mocha'],
   extends: ['plugin:prettier/recommended'],
-  rules: {},
+  rules: {
+    '@typescript-eslint/no-extraneous-class': 'warn',
+  },
   overrides: [
     {
       files: ['*.ts'],
--- a/src/bidiMapper/BidiServer.ts
+++ b/src/bidiMapper/BidiServer.ts
@@ -24,7 +24,7 @@ import {OutgoingBidiMessage} from './OutgoindBidiMessage.js';
 import {EventManager} from './domains/events/EventManager.js';
 import {BidiParser, CommandProcessor} from './CommandProcessor.js';
 import {CdpConnection} from './CdpConnection.js';
-import {BrowsingContextStorage} from './domains/context/browsingContextStorage.js';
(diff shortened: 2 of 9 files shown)


.eslintrc.js
```diff
@@ -7,7 +7,9 @@ module.exports = {
   parser: '@typescript-eslint/parser',
   plugins: ['@typescript-eslint', 'import', 'mocha'],
   extends: ['plugin:prettier/recommended'],
-  rules: {},
+  rules: {
+    '@typescript-eslint/no-extraneous-class': 'warn',
```
thiagowfx, line 11: Warn for now until all occurrences are fixed.

src/bidiMapper/BidiServer.ts
```diff
@@ -85,9 +85,7 @@ export class BidiServer extends EventEmitter<BidiServerEvents> {
       flatten: true,
     });
 
-    await Promise.all(
-      BrowsingContextStorage.getTopLevelContexts().map((c) => c.awaitLoaded())
-    );
+    await Promise.all(getTopLevelContexts().map((c) => c.awaitLoaded()));
```
OrKoN, line 88: this data should be local to the server. Let's not migrate the static until we address the enderlying issue? WDYT? or let's mark it somehow with a TODO so that we can find it (currently, the static keyword is an indicator of the issue.).

src/bidiMapper/domains/context/browsingContextImpl.ts
```diff
@@ -77,7 +81,7 @@ export class BrowsingContextImpl {
 
     this.#initListeners();
 
-    BrowsingContextStorage.addContext(this);
+    addContext(this);
```
OrKoN, line 84: ditto

src/bidiMapper/domains/context/browsingContextImpl.ts
```diff
@@ -175,7 +179,7 @@ export class BrowsingContextImpl {
       },
       this.contextId
     );
-    BrowsingContextStorage.removeContext(this.contextId);
+    removeContext(this.contextId);
```
OrKoN, line 182: ditto

src/bidiMapper/BidiServer.ts
thiagowfx: Ack. Will rework this PR

PR #394, opened 2023-01-24
eslint: add no-extraneous-class rule and fix most of mapperTabPage

thiagowfx:

Bug: #392

### DIFF
--- a/src/bidiTab/mapperTabPage.ts
+++ b/src/bidiTab/mapperTabPage.ts
@@ -22,68 +22,70 @@ declare global {
   }
 }
 
-export class MapperTabPage {
-  // HTML source code for the user-facing Mapper tab.
-  static #mapperPageSource =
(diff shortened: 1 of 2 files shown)


src/bidiTab/mapperTabPage.ts
```diff
@@ -22,68 +22,70 @@ declare global {
...
+}
 
-    return document.getElementById(containerId)!;
+export class MapperTabPage {
+  static generatePage() {
+    // If not in a browser (e.g. unit test), do nothing.
+    if (!globalThis.document?.documentElement) {
+      return;
+    }
+    window.MapperTabPage = MapperTabPage;
```
thiagowfx, line 83: I could get rid of `generatePage` too but I am not sure what to do with this assignment. Thoughts?

thiagowfx:
Superseded by https://github.com/GoogleChromeLabs/chromium-bidi/pull/396

PR #396, opened 2023-01-24
eslint: add no-extraneous-class rule and fix most of mapperTabPage

thiagowfx:

Bug: #392

### DIFF
--- a/src/bidiTab/mapperTabPage.ts
+++ b/src/bidiTab/mapperTabPage.ts
@@ -22,23 +22,53 @@ declare global {
   }
 }
 
-export class MapperTabPage {
-  // HTML source code for the user-facing Mapper tab.
-  static #mapperPageSource =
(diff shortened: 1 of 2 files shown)


src/bidiTab/mapperTabPage.ts
```diff
@@ -22,31 +22,61 @@ declare global {
...
+  const htmlItem = document.createElement('div');
+  htmlItem.className = 'item';
+  htmlItem.innerHTML = `<h3>${logType}</h3><div id="${containerId}" class="log"></div>`;
+  debugElement.appendChild(htmlItem);
+
+  return document.getElementById(containerId)!;
+}
+
+export class MapperTabPage {
   static generatePage() {
```
thiagowfx, line 60: I could get rid of `generatePage` too but I am not sure what to do with this assignment. Thoughts?

src/bidiTab/mapperTabPage.ts
```diff
@@ -22,31 +22,61 @@ declare global {
...
-    this.#findOrCreateTypeLogContainer('BiDi Messages');
-    this.#findOrCreateTypeLogContainer('Browsing Contexts');
-    this.#findOrCreateTypeLogContainer('CDP');
+    findOrCreateTypeLogContainer('System');
+    findOrCreateTypeLogContainer('BiDi Messages');
+    findOrCreateTypeLogContainer('Browsing Contexts');
+    findOrCreateTypeLogContainer('CDP');
   }
 
   static log(logType: LogType, ...messages: unknown[]) {
```
thiagowfx, line 74: If I migrate `log`, the tests fail.

OrKoN (approved):

PR #400, opened 2023-01-25
eslint: add no-extraneous-class rule and fix BrowsingContextStorage

thiagowfx:

- prerefactor BidiServer static factory, create helper function to load all top level contexts
- make EventManager use a non-static version of BrowsingContextStorage

Bug: #392

### DIFF
--- a/src/bidiMapper/BidiServer.ts
+++ b/src/bidiMapper/BidiServer.ts
@@ -14,17 +14,15 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
-import {EventEmitter} from '../utils/EventEmitter.js';
-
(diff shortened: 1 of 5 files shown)


src/bidiMapper/BidiServer.ts
```diff
@@ -85,10 +88,16 @@ export class BidiServer extends EventEmitter<BidiServerEvents> {
       flatten: true,
     });
 
+    await server.awaitLoadTopLevelContexts();
+    return server;
+  }
+
+  async awaitLoadTopLevelContexts() {
```
OrKoN, line 95:
```suggestion
  async topLevelContextsLoaded() {
```

PR #403, opened 2023-01-25
make all EventManager properties and methods non-static

thiagowfx:
Bug: #392

### DIFF
--- a/src/bidiMapper/domains/events/EventManager.ts
+++ b/src/bidiMapper/domains/events/EventManager.ts
@@ -74,13 +74,14 @@ export interface IEventManager {
   ): Promise<void>;
 }
 
+/**
+ * Maps event name to a desired buffer length.
+ */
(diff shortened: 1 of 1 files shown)


src/bidiMapper/domains/events/EventManager.ts
```diff
@@ -78,9 +78,8 @@ export class EventManager implements IEventManager {
   /**
    * Maps event name to a desired buffer length.
    */
-  static readonly #eventBufferLength: Map<string, number> = new Map([
-    ['log.entryAdded', 100],
-  ]);
+  #eventBufferLength = new Map<string, number>([['log.entryAdded', 100]]);
```
OrKoN, line 81: I think it should be a const and `Readonly<Map<string, number>>`

thiagowfx: Done

OrKoN (approved):

PR #405, opened 2023-01-25
make all LogManager properties and methods non-static

thiagowfx:
Bug: #392


src/bidiMapper/domains/log/logManager.ts
```diff
@@ -162,8 +158,8 @@ export class LogManager {
     return 'info';
   }
 
-  // convert CDP StackTrace object to Bidi StackTrace object
-  static #getBidiStackTrace(
+  /** Converts CDP StackTrace object to Bidi StackTrace object. */
+  #getBidiStackTrace(
```
OrKoN, line 162: I think this has to be a module-level function? i.e., if a function does not use any state on the instance it should be a standalone function

src/bidiMapper/domains/log/logManager.ts
```diff
@@ -149,7 +145,7 @@ export class LogManager {
     );
   }
 
-  static #getLogLevel(consoleApiType: string): Log.LogLevel {
+  #getLogLevel(consoleApiType: string): Log.LogLevel {
```
OrKoN, line 148: same here

src/bidiMapper/domains/log/logManager.ts
thiagowfx: Done

src/bidiMapper/domains/log/logManager.ts
thiagowfx: Done

OrKoN (approved):

PR #406, opened 2023-01-25
make ScriptEvaluator non-static

thiagowfx:
Bug: https://github.com/GoogleChromeLabs/chromium-bidi/issues/392

### DIFF
--- a/src/bidiMapper/domains/script/scriptEvaluator.ts
+++ b/src/bidiMapper/domains/script/scriptEvaluator.ts
@@ -20,13 +20,355 @@ import {Protocol} from 'devtools-protocol';
 
 // As `script.evaluate` wraps call into serialization script, `lineNumber`
 // should be adjusted.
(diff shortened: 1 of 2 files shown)


OrKoN (approved):

src/bidiMapper/domains/script/scriptEvaluator.ts
```diff
@@ -57,7 +57,7 @@ export class ScriptEvaluator {
     );
   }
 
-  static #cdpRemoteObjectToCallArgument(
+  #cdpRemoteObjectToCallArgument(
```
OrKoN, line 60: many methods here can probably be top-level functions

thiagowfx: WIP. PTAL. Should we move `readonly #knownHandlesToRealm: Map<string, string> = new Map();` to top-level?

PR #408, opened 2023-01-26
[EXPERIMENT] run wpt tests in parallel to speed up CI

thiagowfx:


thiagowfx:
There are only 2 cores. Improvement is insignificant, still ~5-6min. Closing.

PR #409, opened 2023-01-26
eslint: add eslint:recommended and fix a few more issues

thiagowfx:


src/bidiMapper/domains/script/scriptEvaluator.ts
```diff
@@ -306,11 +306,11 @@ function webDriverValueToBiDi(
   }
 
   if (result.type == 'node') {
-    if (bidiValue.hasOwnProperty('backendNodeId')) {
+    if (Object.prototype.hasOwnProperty.call(bidiValue, 'backendNodeId')) {
```
OrKoN, line 309: why not https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn ?

thiagowfx: Sure! My fix was the recommended one in the rule README page, but that should also work.

thiagowfx: Done.

Except here:

src/bidiMapper/domains/log/logHelper.ts:154:12 - error TS2339: Property 'hasOwn' does not exist on type 'RemoteValue'.
  Property 'hasOwn' does not exist on type 'UndefinedValue'.

src/bidiMapper/domains/log/logHelper.ts
```diff
@@ -151,7 +151,7 @@ function toJson(arg: CommonDataTypes.RemoteValue): string {
 }
 
 function stringFromArg(arg: CommonDataTypes.RemoteValue): string {
-  if (!arg.hasOwnProperty('value')) {
+  if (!Object.prototype.hasOwnProperty.call(arg, 'value')) {
```
OrKoN, line 154: same here

thiagowfx: Done

OrKoN (approved):

PR #413, opened 2023-01-26
github workflows: run npm ci instead of npm install

thiagowfx:
"npm run prepare" is always executed as part of "npm install".
Currently, CI is running "npm run build" redundantly twice, once as part
of "install", and once as part of an explicit build step.

"npm ci" does not execute "npm run prepare".


OrKoN:
The reason is that you don't need to run build explicitly.

OrKoN:
It is also ensure the package is built before publish.

OrKoN (approved):

PR #416, opened 2023-01-26
eslint: no-extraneous-class: fix BidiServerRunner

thiagowfx:

Bug: #392


OrKoN (approved):
SG but files outside of src/bidiMapper are not very relevant for pptr (we only use src/bidiMapper)

thiagowfx:
Ah, OK

PR #418, opened 2023-01-26
apply bunch of misc changes related to style and linting

thiagowfx:


OrKoN:
We usually don't land code without a code review. Let's have a reviewer for every code change.

thiagowfx:
Sure! My rationale was to land things directly when they are stylistic in nature / no-ops, but I could send them for review in the future as well.

PR #419, opened 2023-01-26
WIP realm

thiagowfx:


thiagowfx:
Here the goal is to move the class variable `Realm.#realmMap` to `bidiServer` (and outside of `Realm`), correct?

Or do you think we should keep the class variable within `Realm`, and instead just add an instance of `realm` to `bidiServer`?

src/bidiMapper/domains/script/realm.ts
```diff
@@ -27,9 +27,45 @@ export enum RealmType {
 
 const scriptEvaluator = new ScriptEvaluator();
 
-export class Realm {
-  static readonly #realmMap: Map<string, Realm> = new Map();
+const realmMap: Map<string, Realm> = new Map();
```
OrKoN, line 30: I thought I wrote a comment by I didn't: let's create a RealmStorage class to hold this map. The instance of that class should be instantiated on the BidiServer and passed around to other classes.

thiagowfx:
This is ready and final, however the tests are failing. Let me know if you have any idea what could be wrong here. Will try to debug it.

How to trigger it:

```
Terminal 1: npm run server
Terminal 2: python3 -m pytest -k tests/test_log_events.py -v -k test_consoleLog_textAndArgs
```

thiagowfx:
I figured it out. Superseded by https://github.com/GoogleChromeLabs/chromium-bidi/pull/429

PR #422, opened 2023-01-30
Make BrowsingContextStorage non-static, and an instance of BidiServer

thiagowfx:
As a side effect it is now possible to have multiple
`#BrowsingContextStorage.contexts`.

Furthermore, enforce the `no-extraneous-class` ESLint rule.

The following classes were refactored in order to accommodate that:

- BidiServer
- BrowsingContextImpl
- BrowsingContextProcessor
- EventManager
- Realm
- SubscriptionManager

Bug: https://github.com/GoogleChromeLabs/chromium-bidi/issues/392

### DIFF
--- a/src/bidiMapper/domains/context/browsingContextStorage.ts
+++ b/src/bidiMapper/domains/context/browsingContextStorage.ts
@@ -19,37 +19,35 @@ import {BrowsingContextImpl} from './browsingContextImpl.js';
 import {Message} from '../../../protocol/protocol.js';
 
 export class BrowsingContextStorage {
-  static #contexts: Map<string, BrowsingContextImpl> = new Map();
+  readonly #contexts = new Map<string, BrowsingContextImpl>();
 
-  static getTopLevelContexts(): BrowsingContextImpl[] {
-    return Array.from(BrowsingContextStorage.#contexts.values()).filter(
+  getTopLevelContexts(): BrowsingContextImpl[] {
+    return Array.from(this.#contexts.values()).filter(
       (c) => c.parentId === null
     );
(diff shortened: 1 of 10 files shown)


OrKoN (approved):

thiagowfx:
realm is added

thiagowfx:
dependencygraph

src/bidiMapper/domains/context/browsingContextStorage.ts
```diff
@@ -19,37 +19,35 @@ import {BrowsingContextImpl} from './browsingContextImpl.js';
...
-    return BrowsingContextStorage.#contexts.has(contextId);
+  hasKnownContext(contextId: string): boolean {
+    return this.#contexts.has(contextId);
   }
 
-  static findContext(contextId: string): BrowsingContextImpl | undefined {
-    return BrowsingContextStorage.#contexts.get(contextId)!;
+  findContext(contextId: string): BrowsingContextImpl {
+    return this.#contexts.get(contextId)!;
   }
```
sadym-chromium, line 47: `find` differs from `get` by the fact it can return undefined if item not found.

thiagowfx: The point here is that it doesn't make sense for the signature to be `| undefined` if we're doing `!`.

We could make this change the other way around: re-add `| undefined` and remove `!`.

thiagowfx: https://github.com/GoogleChromeLabs/chromium-bidi/pull/476

PR #425, opened 2023-01-31
make a few TypeScript improvements not covered by ESLint

thiagowfx:
- prefer ReadonlyMap over Readonly<Map>
- prefer unknown or specific types over any: https://google.github.io/styleguide/tsguide.html#any


OrKoN (approved):
Nice!

Issue #361, opened 2022-12-20
Add e2e / wpt test coverage for Windows and Mac


OrKoN:
Currently, the implementation is only tested against Linux binaries. We should expand the testing to Mac and Windows, and possibly headful, headless, new headless to catch flakiness and platform differences early.

sadym-chromium:
I'm not sure if we should run both new headless and headful. And not sure if we should run all 3 options in all 3 OSes. We can start with running only "new headless" in Mac and Win.

thiagowfx:
@OrKoN: @sadym-chromium suggests we do not need macOS coverage at this time (e.g. as delivered by https://github.com/GoogleChromeLabs/chromium-bidi/pull/427), as we do not iterate _that_ quickly.

Should we close this issue?

mathiasbynens:
What’s the motivation behind not wanting to keep the macOS coverage?

macOS support is crucial for BiDi-in-Puppeteer. If we don’t test it at this repository’s level, how can we be confident that the integration into Puppeteer works?

sadym-chromium:
My suggestion is not to abandon but to deprioritise it.

thiagowfx:
https://github.com/GoogleChromeLabs/chromium-bidi/pull/427 is already written so you can consider it a low-hanging fruit to enable E2E on macOS.

That said, yes I agree this is a good time to shift gears and switch into some real feature work / fixing tests, as our CI got in really good shape recently.

thiagowfx:
Added a `Status` section to the first comment. I don't think we want that explosion of CI combinations though.

Could you please add notes to the first comment to clarify which combinations we're NOT interested in?

thiagowfx:
I'll close this as a duplicate of https://github.com/GoogleChromeLabs/chromium-bidi/issues/876. It's trivial to make combinations of {linux, mac} x {headful, headless} x {wpt, e2e}.

The difficult part is to add Windows support, which is more appropriately tracked in the other bug.

PR #427, opened 2023-01-31
use puppeteer/browsers to install google chrome in CI and add macOS E2E tests

thiagowfx:
Make the installation method platform-agnostic, in order to eventually support running CI on Windows and macOS.

While we're here, add macOS to E2E tests.

Bug: #361
Docs: https://github.com/puppeteer/puppeteer/tree/main/packages/browsers

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,4 +1,4 @@
-# This workflow will run headful and headless e2e tests.
+# This workflow will run headful and headless E2E tests.
 
 name: E2E tests
 
@@ -10,19 +10,15 @@ on:
 jobs:
   e2e:
     timeout-minutes: 15
-    runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
+        os: [ubuntu-latest, macos-latest]
         head: [headful, headless]
+    runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout
         uses: actions/checkout@v3
-      - name: Install chrome-dev
-        run: |
-          wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | sudo apt-key add -
-          sudo sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list'
-          sudo apt-get update
-          sudo apt-get install -y google-chrome-unstable
       - name: Set up node.js
         uses: actions/setup-node@v3
         with:
@@ -30,6 +26,10 @@ jobs:
           cache: npm
       - name: Install and build npm dependencies
         run: npm ci
+      - name: Install Chromium
+        run: |
+          # browsers from @puppeteer/browsers
+          echo "chromium_binary=$(npx browsers install chromium@latest | cut -f 2 -d' ')" >> $GITHUB_ENV
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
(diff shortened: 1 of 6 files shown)


thiagowfx:
For whatever reason, [browser-actions/setup-chrome](https://github.com/browser-actions/setup-chrome) does not work well with macOS. Its outputted version is v93 which is super old.

.github/workflows/e2e.yml
```diff
@@ -34,12 +35,16 @@ jobs:
...
           cache: 'pip'
       - name: Install Python dependencies
         run: pip install -r tests/requirements.txt
       - name: Run e2e tests
-        run: CHANNEL=chrome-dev xvfb-run --auto-servernum npm run e2e-${{ matrix.headful-argument }}
+        if: runner.os == 'Linux'
+        run: PUPPETEER_EXECUTABLE_PATH="$(which google-chrome-unstable)" xvfb-run --auto-servernum npm run e2e-${{ matrix.headful-argument }}
+      - name: Run e2e tests
+        if: runner.os == 'macOS'
+        run: PUPPETEER_EXECUTABLE_PATH="${{ steps.chrome.outputs.path }}/${{ steps.chrome.outputs.binary }}" npm run e2e-${{ matrix.headful-argument }}
```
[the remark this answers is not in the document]
thiagowfx: It is correct. In order to unblock this, I reverted the WPT-specific part.

.github/workflows/e2e.yml
```diff
@@ -10,35 +10,46 @@ on:
...
         uses: actions/setup-node@v3
         with:
           node-version: 16
           cache: npm
       - name: Install and build npm dependencies
         run: npm ci
+      - name: Install Chromium
+        id: chromium
+        run: |
+          version=$(npx @puppeteer/browsers install chrome@latest | cut -f 1 -d' ' | sed -e 's/chrome@//')
```
mathiasbynens, line 32: Worth adding `@puppeteer/browsers` to devDependencies instead? Then we can just use `browsers` here, and it might be helpful when we testing/debugging locally anyhow.

mathiasbynens (approved):

sadym-chromium (approved):
Do you want to change it in wpt as well?

.github/workflows/e2e.yml
thiagowfx: Done

.github/workflows/e2e.yml
```diff
@@ -10,35 +10,47 @@ on:
...
         with:
           node-version: 16
           cache: npm
       - name: Install and build npm dependencies
         run: npm ci
+      - name: Install Chromium
+        id: chromium
+        run: |
+          # browsers from @puppeteer/browsers
+          version=$(npx browsers install chrome@latest | cut -f 1 -d' ' | sed -e 's/chrome@//')
```
OrKoN, line 33: with https://github.com/puppeteer/puppeteer/pull/9797 you should be able to do `cut -f 2 -d' '` which would give you the executable path directly.

thiagowfx: Integrated

OrKoN (approved):

.github/workflows/e2e.yml
mathiasbynens: We don’t need `npx` anymore now that it’s a `devDependency`

thiagowfx: I forgot to update this comment, but I had done it and then it errored out, so in the end I undid the change.

You're saying that `browsers install chrome@latest` alone should work here, correct?
I can try again in a follow-up

mathiasbynens: Hmm, I’m pretty sure this is what I’m doing in a bunch of GitHub Actions workflows I set up. But now that you mention it, it might be that I explicitly did `echo "$(npm bin)" >> $GITHUB_PATH` as part of the setup step (that also runs `npm ci`).

thiagowfx:
This change may need to be reverted. Although it has fully passed in the PR, 2 of the workflows are failing on `main`: https://github.com/GoogleChromeLabs/chromium-bidi/actions/runs/4368453683/jobs/7641057899

It could be unrelated though, let's see.

thiagowfx:
Partial revert is here: https://github.com/GoogleChromeLabs/chromium-bidi/pull/521

OrKoN:
Perhaps we could add logging to the install action to make sure we accidentally don't proceed on errors. The test errors look unrelated but I could not see what is the root cause, smth failing the in python code.

PR #429, opened 2023-01-31
make Realm.#realmMap non-static by moving it to RealmStorage

### LINKED ISSUE #392: Reduce the number of static methods and variables

thiagowfx:
With the intent to support the ability to run multiple bidi sessions simultaneously.

Closes #392

### DIFF
--- a/src/bidiMapper/domains/context/browsingContextImpl.ts
+++ b/src/bidiMapper/domains/context/browsingContextImpl.ts
@@ -16,13 +16,14 @@
  */
 
 import {BrowsingContext, Message} from '../../../protocol/protocol.js';
-import {Realm, RealmType} from '../script/realm.js';
 import {BrowsingContextStorage} from './browsingContextStorage.js';
 import {CdpClient} from '../../CdpConnection.js';
 import {Deferred} from '../../../utils/deferred.js';
 import {IEventManager} from '../events/EventManager.js';
 import {LogManager} from '../log/logManager.js';
 import {Protocol} from 'devtools-protocol';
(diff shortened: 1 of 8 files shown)


src/bidiMapper/domains/context/browsingContextImpl.ts
```diff
@@ -377,14 +389,15 @@ export class BrowsingContextImpl {
...
+        const realm = new Realm(
+          this.#realmStorage,
           params.context.uniqueId,
           this.contextId,
           this.navigableId ?? 'UNKNOWN',
           params.context.id,
           this.#getOrigin(params),
           // TODO: differentiate types.
-          RealmType.window,
+          'window',
```
OrKoN, line 400: should this remain a named constant?

OrKoN (approved):

src/bidiMapper/domains/context/browsingContextImpl.ts
thiagowfx: RealmType.window causes a compilation error

PR #432, opened 2023-02-07
Update `yapf` style

sadym-chromium:
Update `yapf` style to increase readability of the 
```
"prop_1":
"value",
"prop_2":
"value",
```

to 

```
"prop_1": "value",
"prop_2": "value",
```

Related issue: https://github.com/google/yapf/issues/392


.pre-commit-config.yaml
```diff
@@ -43,7 +43,7 @@ repos:
     rev: v0.32.0
     hooks:
       - id: yapf
-        args: ['--in-place', '--parallel']
+        args: ['--in-place', '--parallel', '--recursive']
```
thiagowfx, line 46: If I remember correctly, `--recursive` isn't needed here.

`pre-commit` applies changes to all files included in the commit (`git diff` / `git show --stat`).

thiagowfx: Just do `pre-commit run --all-files` once locally to apply it to all files in the repository, that should be enough.

.style.yapf
```diff
@@ -0,0 +1,4 @@
+[style]
+based_on_style = pep8
+INDENT_DICTIONARY_VALUE=True
+ALLOW_SPLIT_BEFORE_DICT_VALUE=False
```
thiagowfx, line 4:
```suggestion
based_on_style = pep8
indent_dictionary_value = True
allow_split_before_dict_value = False
```

(consistent case with `based_on_style`)

.pre-commit-config.yaml
sadym-chromium: done

thiagowfx (approved):

Issue #421, opened 2023-01-30
BiDiMapper should only show logs when needed


Lightning00Blade:
Logs coming from https://github.com/GoogleChromeLabs/chromium-bidi/blob/main/src/utils/log.ts#L26 are always show.  It should only show them under some condition - Example would be debugging or the https://github.com/GoogleChromeLabs/chromium-bidi/blob/main/src/bidiTab/bidiTab.ts that uses it to display logs. 

Suggestions:
Something similar to the Puppeteer [Debug](https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer-core/src/common/Debug.ts) can be implemented or the ability to provide a custom logger to `chromium-bidi` can be added.

thiagowfx:
- [ ] First comment
- [ ] Thanks. I also need to check if the logic on https://github.com/GoogleChromeLabs/chromium-bidi/blob/main/src/utils/log.ts#L26 still works, especially given https://github.com/GoogleChromeLabs/chromium-bidi/pull/420/files.

OrKoN:
@thiagowfx let's make sure that the debug library is not statically imported in the bidiMapper subfolder

PR #434, opened 2023-02-09
Regression fix: output BiDi Mapper Tab logs

### LINKED ISSUE #433: Regression: BiDi Mapper Tab Logs are not displayed
https://github.com/GoogleChromeLabs/chromium-bidi/issues/392 introduced a regression: The chrome mapper tab does not output bidi server logs anymore.

It has been broken by https://github.com/GoogleChromeLabs/chromium-bidi/pull/420/files.

https://github.com/GoogleChromeLabs/chromium-bidi/blob/8a14af627dbafef643c528ec86c52e3638d2eca2/src/utils/log.ts#L26 needs to be reverted.

thiagowfx:
Changes the loggers to be injected instead of being global objects. The logger is removed from the command processor as it is used in a single place and is duplicating a thrown error.

Closes #433 #421


thiagowfx:
Sorry, I changed the wrong class. It's not `log.ts`, it's `mapperTabPage.js`. Please hold off review until I mark it as ready.

README.md
```diff
@@ -306,10 +306,18 @@ https://www.google.com/chrome/dev/
...
+Oneshot:
+
 ```sh
 npm run build
 ```
 
+Continuously with live-reload:
+
+```sh
+npm run watch
```
sadym-chromium, line 318: Does it re-run the Mapper?

src/utils/log.ts
```diff
@@ -23,17 +29,20 @@ export enum LogType {
...
-    if ('MapperTabPage' in globalThis) {
-      (
-        globalThis as unknown as {
-          MapperTabPage: {log: (...message: unknown[]) => void};
-        }
-      )['MapperTabPage'].log(logType, ...messages);
-    }
-  };
+/* eslint "@typescript-eslint/no-extraneous-class": "off" */
+export class MapperTabPage {
```
sadym-chromium, line 33: `MapperPageTab` is not a thing in case of Puppeteer, so I'd prefer to keep it more abstract

sadym-chromium:
> Sorry, I changed the wrong class. It's not `log.ts`, it's `mapperTabPage.js`. Please hold off review until I mark it as ready.

oops, read it after the review comments were published

OrKoN:
WDYT if we make the log messages to emitted on the bidiserver (in mapper) instance? we have our own logging in Puppeteer and there is a separate logging in the BidiTab so it makes sense. If you need logs, you just add an event listener to the bidi server instance.

README.md
thiagowfx: No. But it's easy to accomplish so with `entr`. I'll remove "live-reload" to make it clearer.

OrKoN (approved):

thiagowfx:
@sadym-chromium the current form reflects Alex's idea. Let me know what you think.

It's not fully complete yet (some more logs need to be migrated) but the gist is there.

sadym-chromium:
The approach LGTM, as it's scalable for Puppeteer use case. As discussed, I'd add e2e tests for the tab. It can be done via `cdp.sendCommand`.

thiagowfx:
Sounds good, I'll look into expanding it to more files then.

thiagowfx:
I split this CL into smaller ones to make it easier to submit.

thiagowfx:
@OrKoN I applied the same pattern to `browsingContextProcessor.ts`, which is instantiated from `CommandProcessor.ts`, which in turn is instantiated from `BidiServer.ts`, however the logs do not show up on the "Browsing Contexts" section.

Could you help me debug it? The relevant commit is the second one within this PR.

src/bidiTab/bidiTab.ts
```diff
@@ -68,14 +65,21 @@ const _waitSelfTargetIdPromise = _waitSelfTargetId();
 
   const bidiServer = await _createBidiServer(selfTargetId);
 
-  logSystem('launched');
+  bidiServer.getCommandProcessor().on('log', (...messages: unknown[]) => {
```
OrKoN, line 68: let's not do this: all logs should be handled by the bidiServer.on('log', () => {}). If you need to differentiate between various types inside the bidiServer, let's include it as the first argument (this does not need to be type-safe I think so no cross-package imports are needed).

thiagowfx: even if I remove this, the browsing context log doesn't show up anywhere

(prior to this, it would show up in the console, as it had `console.log`)

OrKoN: I have added a commit to this branch. PTAL.

OrKoN: I think an alternative would be to always provide a logger function instead of relying on events. I think it makes things easier.

thiagowfx: Thanks. Browsing contexts seem to be still empty though:
image

thiagowfx: I wonder if it could be a race condition, or just that we're setting up the listener too late (after the event is logged).

OrKoN: You need to open a new context to see them I believe

thiagowfx: Ah, yes! 
image

`browsingContext.create`.

However, we still do not see the _very first_ context. Is this OK?

thiagowfx:
Progress update: The only missing file now is "command parser".

src/bidiTab/bidiTab.ts
OrKoN: I am sure but I don't see a race condition: perhaps @sadym-chromium knows

OrKoN:
@thiagowfx I think using the event emitter for logs turns out to be not so great. Let's inject the log function in those cases too.

src/bidiTab/bidiTab.ts
sadym-chromium: The context with the Mapper itself should not be shown there. We hide it from BiDi users (but not from CDP).

thiagowfx:
> @thiagowfx I think using the event emitter for logs turns out to be not so great. Let's inject the log function in those cases too.

@OrKoN We cannot inject the function though, because it includes modules from another tree/project (remember, an error with `rootDir`)? If we follow this route then we will need to either:

- move some files around (so that they are in the same project / tsconfig scope)
- expand tsconfig to include more directories (we currently have lots of small tsconfig per directory)
- refactor some methods to eliminate inter-directory dependencies

Any preference? There's also this file: https://github.com/GoogleChromeLabs/chromium-bidi/blob/51cee1ce9dd557532f12e4990a6d358fa3761a80/src/cdp/index.ts that exports several classes from other modules. I wonder if we could maybe leverage it to reduce coupling.

src/bidiTab/bidiTab.ts
thiagowfx: Then it's WAI. Although, previously, the context with the mapper itself was being unconditionally logged. If hiding the very first context is not a regression then it's an improvement of the status quo.

OrKoN:
@thiagowfx we can inject the function without static imports. Do you want me to take over this task?

thiagowfx:
@OrKoN sure, go ahead, you can continue from this PR.

OrKoN:
@thiagowfx thanks, I will try to finalise it today.

src/bidiTab/bidiTab.ts
```diff
@@ -141,8 +143,9 @@ async function _createBidiServer(selfTargetId: string) {
 
     async sendMessage(message: Message.OutgoingMessage): Promise<void> {
       const messageStr = JSON.stringify(message);
-      logBidi('sent > ', messageStr);
+      const messagePretty = JSON.stringify(message, null, 2);
```
sadym-chromium, line 146: NIT: it's nice, but the tab style doesn't respect the line breaks in the log

src/bidiTab/mapperTabPage.ts
```diff
@@ -18,7 +18,7 @@ import {LogType} from '../utils/log.js';
...
+  '<!DOCTYPE html><title>BiDi-CDP Mapper</title><style>body{font-family: Roboto, serif; font-size: 13px; color: #202124;}.log{padding: 12px; font-family: Menlo, Consolas, Monaco, Liberation Mono, Lucida Console, monospace; font-size: 11px; line-height: 180%; background: #f1f3f4; border-radius: 4px;}.pre{overflow-wrap: break-word; padding: 10px;}.card{margin: 60px auto; padding: 2px 0; max-width: 900px; box-shadow: 0 1px 4px rgba(0, 0, 0, 0.15), 0 1px 6px rgba(0, 0, 0, 0.2); border-radius: 8px;}.divider{height: 1px; background: #f0f0f0;}.item{padding: 16px 20px;}</style><div class="card"><div class="item"><h1>BiDi-CDP Mapper is controlling this tab</h1><p>Closing or reloading it will stop the BiDi process. <a target="_blank" title="BiDi-CDP Mapper GitHub Repository" href="https://github.com/GoogleChromeLabs/chromium-bidi">Details</a>.</p></div><div id="details"></div></div>';
```
sadym-chromium, line 21: Any particular reason to expand the debug logs by default?

PR #445, opened 2023-02-14
npm: always recompile sources upon running unit tests

thiagowfx:
Remove "unit" target, default to standard "test" target. Previously the "unit" target would not recompile sources.

Side effect: It's possible to do "npm test" in lieu of "npm run test".


thiagowfx:
We may or may not keep the "unit" target around. Let me know your thoughts. I removed it.

sadym-chromium (approved):

PR #446, opened 2023-02-14
use debug package for logging in lieu of console.log

thiagowfx:
Bug: #421


thiagowfx:
`log.ts` should be handled differently so it shall be done in another PR.

OrKoN (approved):

PR #449, opened 2023-02-15
Refactor logging in the mapper and the tab

OrKoN:
Changes the loggers to be injected instead of being global objects. The logger is removed from the command processor as it is used in a single place and is duplicating a thrown error.

Closes https://github.com/GoogleChromeLabs/chromium-bidi/issues/433 https://github.com/GoogleChromeLabs/chromium-bidi/issues/421

### DIFF
--- a/src/bidiMapper/CommandProcessor.ts
+++ b/src/bidiMapper/CommandProcessor.ts
@@ -32,7 +32,6 @@ import {OutgoingBidiMessage} from './OutgoingBidiMessage.js';
 import {RealmStorage} from './domains/script/realmStorage.js';
 
 type CommandProcessorEvents = {
-  log: unknown[];
   response: Promise<OutgoingBidiMessage>;
 };
 


sadym-chromium (approved):

src/bidiMapper/CommandProcessor.ts
```diff
@@ -27,10 +27,12 @@ import {BrowsingContextStorage} from './domains/context/browsingContextStorage.j
 import {CdpConnection} from './CdpConnection.js';
 import {EventEmitter} from '../utils/EventEmitter.js';
 import {IEventManager} from './domains/events/EventManager.js';
+import {LoggerFn} from '../utils/log.js';
 import {OutgoingBidiMessage} from './OutgoingBidiMessage.js';
 import {RealmStorage} from './domains/script/realmStorage.js';
 
 type CommandProcessorEvents = {
+  log: unknown[];
```
thiagowfx, line 35: You probably forgot to remove `log` from `CommandProcessorEvents`.

thiagowfx, line 35:
```suggestion
```

thiagowfx (approved):

Issue #317, opened 2022-11-16
Deflake WPT and e2e


sadym-chromium:
Sometimes the WPT and e2e tests are flaky. They should be fixed.

* https://github.com/GoogleChromeLabs/chromium-bidi/actions/runs/3481576076

thiagowfx:
The following test targets are increasingly flake in the past few days:

```
tests/test_browsing_context.py 
  test_browsingContext_navigateSameDocumentNavigation_waitComplete_navigated
  test_browsingContext_navigateSameDocumentNavigation_waitInteractive_navigated 
  test_browsingContext_navigateSameDocumentNavigation_navigated

tests/test_nested_browsing_context.py 
  test_nestedBrowsingContext_navigateSameDocumentNavigation_navigated
```

thiagowfx:
Example: https://github.com/GoogleChromeLabs/chromium-bidi/actions/runs/4203980102/jobs/7294046160

sadym-chromium:
Consider fixed by #462

PR #450, opened 2023-02-15
split a few E2E tests into smaller ones to attempt to reduce flakiness

thiagowfx:
Bug: #317


thiagowfx:
Split off https://github.com/GoogleChromeLabs/chromium-bidi/pull/434

sadym-chromium:
I'd prefer to fix the flakiness first

thiagowfx:
Even if the flakiness is fixed, don't you agree these two unit tests are quite large? Splitting them should be beneficial either way.

sadym-chromium (approved):

PR #462, opened 2023-02-20
Fix same document navigation race condition

sadym-chromium:
In case of 2 consequent "same document navigation", the `navigatedWithinDocument` deferred was kept finished instead of being re-created.

This PR should fix the issue and flakiness of e2e tests

### DIFF
--- a/src/bidiMapper/domains/context/browsingContextImpl.ts
+++ b/src/bidiMapper/domains/context/browsingContextImpl.ts
@@ -330,7 +330,7 @@ export class BrowsingContextImpl {
         }
 
         if (params.name === 'init') {
-          this.#documentMaybeChanged(params.loaderId);
+          this.#documentChanged(params.loaderId);
           this.#targetDefers.documentInitialized.resolve();
         }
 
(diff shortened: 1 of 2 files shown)


OrKoN:
Let's add a test case when you navigate to a page and the page redirects to a same-document URL (not sure if it will work or not, but I think it's a good test to have if there is no such test yet).

sadym-chromium:
> Let's add a test case when you navigate to a page and the page redirects to a same-document URL (not sure if it will work or not, but I think it's a good test to have if there is no such test yet).

It's a nice use-case, but we don't have infrastructure to test `301` codes. If we want to test it, I'd test it in WPT.

OrKoN:
@sadym-chromium you don't need 301 codes for a client-side redirect. `window.location.href = '...'`

sadym-chromium:
> @sadym-chromium you don't need 301 codes for a client-side redirect. `window.location.href = '...'`

This use-case is orthogonal to the suggested changes, as with the external navigation we just deal with the fact URL and probably document changed. I consider it out of scope.

thiagowfx (approved):

OrKoN:
@OrKoN not blocking from my side

OrKoN:
@sadym-chromium note the a proposal of the test is related to this change: a new-document navigation that also triggers a same-document navigation (potentially multiple times).

PR #464, opened 2023-02-20
Don't rerun failed e2e tests

sadym-chromium:
Don't rerun e2e tests to be aware of flakiness.


pytest.ini
```diff
@@ -7,4 +7,4 @@ testpaths = tests
 md_report = True
 
 # https://pypi.org/project/pytest-rerunfailures/
-reruns = 3
+reruns = 1
```
thiagowfx, line 10: If you want to remove it, then set it to `reruns = 0`.

`1` means it will rerun once.

sadym-chromium: So confusing! I understand "to re-run all test failures, use the --reruns command line option with the maximum number of times you’d like the tests to run" as "0" means don't even try it.

sadym-chromium: We also can remove the tool for now.

sadym-chromium: Actually, current revision removes `pytest-rerunfailures` completely. We can always roll it again if it is required.

pytest.ini
```diff
@@ -3,8 +3,5 @@
 minversion = 7.0
 testpaths = tests
 
-# https://pypi.org/project/pytest-md-report/
+# pytest-md-report plug-in
```
thiagowfx, line 6:
```suggestion
# https://pypi.org/project/pytest-md-report/
```

thiagowfx (approved):

pytest.ini
thiagowfx: OK, fine as well

PR #474, opened 2023-02-22
Handle subscriptions on the top level contexts

sadym-chromium:
Align subscription mechanism with specification. Partly addresses #353.


tests/test_subscription.py
```diff
@@ -109,25 +127,136 @@ async def test_subscribeWithContext_subscribesToEventsInNestedContext(
...
+            "params": {
+                "expression": "console.log('SOME_MESSAGE')",
+                "target": {
+                    "context": context_id
+                },
+                "awaitPromise": True
+            }
+        })
+
+    # Assert event received in `CHANNEL_2`.
```
thiagowfx, line 147: typo? There's no CHANNEL_2 here

src/bidiMapper/domains/events/SubscriptionManager.ts
```diff
@@ -173,15 +186,25 @@ export class SubscriptionManager {
     }
 
     if (!this.#channelToContextToEventMap.has(channel)) {
-      return;
+      throw new InvalidArgumentException(
+        `Cannot unsubscribe from ${event}, ${contextId} as not subscribed.`
```
thiagowfx, line 190: did you mean: `is not subscribed`? (same below)

thiagowfx (approved):

tests/test_subscription.py
sadym-chromium: done

PR #479, opened 2023-02-23
improve pytest ergonomics with native assertions

thiagowfx:
- introduce and use `anys` pytest matchers
- use native assert comparisons instead of custom `recursive_compare`
- use type comparisons instead of `any_value` in some circumstances

Reference: https://github.com/jwodder/anys

As of now:

image

With Anys, but without icdiff:

image

With this PR (=Anys + icdiff):

image

### DIFF
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -1,3 +1,4 @@
+anys==0.2.1
 pytest==7.2.0
 pytest-asyncio==0.20.2
 pytest-icdiff==0.6
--- a/tests/test_browsing_context.py
+++ b/tests/test_browsing_context.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 import pytest
+from anys import ANY_STR
 from test_helpers import *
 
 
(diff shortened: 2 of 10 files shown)


sadym-chromium:
My concerns: 
1. `any_value` seems to be more readable when the specific value is not important for the test.
2. Is it feasible to add the same `anys` to WPT? I find it useful having not that diverse infra for e2e and WPT.

I assume `dict1 <= dict2` would work as well?

thiagowfx:
> 1. `any_value` seems to be more readable when the specific value is not important for the test.

We could still use an `any_value` equivalent, not a problem.

> 2/ Is it feasible to add the same anys to WPT? I find it useful having not that diverse infra for e2e and WPT.

I see. Possibly, but it would require extra investigation. Will defer it to another time then.

> I assume assert `dict1 <= dict2` would work as well?

What do you expect that comparison to do? (And how is it done today?)

thiagowfx:
I'll archive/simmer this, and only extract the `icdiff` part to another PR.

thiagowfx:
> 1. any_value seems to be more readable when the specific value is not important for the test.

Done. Replace with `from unittest.mock import ANY`.

> 2. Is it feasible to add the same anys to WPT? I find it useful having not that diverse infra for e2e and WPT.

Yes. I looked into wpt tests today and it should be feasible.

sadym-chromium (approved):

PR #490, opened 2023-03-01
zod: drop "Schema" suffix

thiagowfx:
`zod` documentation happily uses schemas and types with the same name. Adopt the same in protocol parser for simplicity.

### DIFF
--- a/src/protocol-parser/protocol-parser.ts
+++ b/src/protocol-parser/protocol-parser.ts
@@ -45,52 +45,52 @@ export function parseObject<T extends ZodType>(
 }
 
 export namespace CommonDataTypes {
-  export const SharedReferenceSchema = zod.object({
+  export const SharedReference = zod.object({
     sharedId: zod.string().min(1),
   });
-  export type SharedReference = zod.infer<typeof SharedReferenceSchema>;
+  export type SharedReference = zod.infer<typeof SharedReference>;
 
-  export const RemoteReferenceSchema = zod.object({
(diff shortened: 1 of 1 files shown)


src/protocol-parser/protocol-parser.ts
```diff
@@ -20,8 +20,6 @@
  * Parser types should match the `../protocol` types.
  */
 
-const MAX_INT = 9007199254740991 as const;
```
sadym-chromium, line 23: NIT: Strictly speaking we cannot rely on `Number.MAX_SAFE_INTEGER` to always be equal to the const [defined in BiDi spec](https://w3c.github.io/webdriver-bidi/#protocol-definition). But I cannot see any reasonable-enough reason why we want to change it in the protocol.

sadym-chromium:
I personally find this reusing of the names confusing.

sadym-chromium (approved):
LGTM, even though IMO the readability decreased with this PR.

src/protocol-parser/protocol-parser.ts
thiagowfx: Oh, that's surprising, but makes sense.

> But I cannot see any reasonable-enough reason why we want to change it in the protocol.

Why?

thiagowfx:
> I personally find this reusing of the names confusing.
> LGTM, even though IMO the readability decreased with this PR.

If you don't think it's readable, then let's not merge it. That's what code reviews are for :-)

Issue #482, opened 2023-02-23
`session.unsubscribe` should be atomic


sadym-chromium:
The command `session.unsubscribe` should be atomic. If operation failed to some reason (not subscribed, browsing context not found, etc):

Expected: nothing should be unsubscribed.
Actual: some of the events processed before the error happened are unsubscribed.

[Related WPT test](https://github.com/web-platform-tests/wpt/blob/ef22a833e19a288616e478ebaffe6f346bc53826/webdriver/tests/bidi/session/unsubscribe/invalid.py#L75).

thiagowfx:
Code pointers:

https://github.com/GoogleChromeLabs/chromium-bidi/blob/471636bd9214be316087b918a98cff63baa6555d/src/bidiMapper/domains/events/EventManager.ts#L203

https://github.com/GoogleChromeLabs/chromium-bidi/blob/471636bd9214be316087b918a98cff63baa6555d/src/bidiMapper/domains/events/SubscriptionManager.ts#L165

PR #493, opened 2023-03-01
write a test to check for atomicity of session.unsubscribe

thiagowfx:
If an operation failed for some reason (not subscribed, browsing context not found, etc), do not unsubscribe partially.

Bug: #482

### DIFF
--- a/tests/test_subscription.py
+++ b/tests/test_subscription.py
@@ -16,7 +16,7 @@
 from unittest.mock import ANY
 
 import pytest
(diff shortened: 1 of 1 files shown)


thiagowfx:
Verified the unit test fails (as expected!) correctly. Need to change the mapper implementation next.

@sadym-chromium here you can appreciate the recursion of the `AnyWithEntries` matcher :D

sadym-chromium (approved):
Feel free to disable the test and merge this PR

PR #505, opened 2023-03-03
typescript: remove unnecessary async in some methods

thiagowfx:
Those were found with the aid of 'require-await': 'error' from ESLint.

The ESLint rule will not be enabled at this time though because it is too strict.

Docs: https://eslint.org/docs/latest/rules/require-await


src/bidiMapper/CommandProcessor.ts
```diff
@@ -113,7 +113,7 @@ export class CommandProcessor extends EventEmitter<CommandProcessorEvents> {
   }
 
   // noinspection JSMethodCanBeStatic,JSUnusedLocalSymbols
```
sadym-chromium, line 115: This inspection is not needed anymore as well.

src/bidiMapper/CommandProcessor.ts
```diff
@@ -113,7 +113,7 @@ export class CommandProcessor extends EventEmitter<CommandProcessorEvents> {
...
-  async #process_session_status(): Promise<Session.StatusResult> {
+  #process_session_status(): Session.StatusResult {
```
sadym-chromium, line 116: can be static

sadym-chromium (approved):

PR #512, opened 2023-03-03
miscellaneous infra/TS improvements

thiagowfx:
See commit message for the list of improvements.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
     "eslint": "eslint --ext js --ext ts --fix .",
     "flake8": "flake8 examples/ tests/",
     "format": "npm run eslint && npm run prettier & npm run yapf",
+    "pre-commit": "pre-commit run --all-files",
     "prepare": "npm run clean && npm run build",
     "prettier": "prettier --write .",
     "rollup": "rollup -c",
--- a/src/bidiMapper/BidiServer.ts
+++ b/src/bidiMapper/BidiServer.ts
@@ -40,6 +40,20 @@ export class BidiServer extends EventEmitter<BidiServerEvents> {
   #realmStorage: RealmStorage;
   #logger?: LoggerFn;
 
+  #handleIncomingMessage = async (message: Message.RawCommandRequest) => {
+    this.#commandProcessor.processCommand(message);
+  };
+
+  #processOutgoingMessage = async (messageEntry: OutgoingBidiMessage) => {
+    const message = messageEntry.message as any;
+
+    if (messageEntry.channel !== null) {
+      message['channel'] = messageEntry.channel;
+    }
+
+    await this.#transport.sendMessage(message);
+  };
+
   private constructor(
     bidiTransport: BidiTransport,
     cdpConnection: CdpConnection,
@@ -75,7 +89,7 @@ export class BidiServer extends EventEmitter<BidiServerEvents> {
     );
   }
 
-  public static async createAndStart(
+  static async createAndStart(
     bidiTransport: BidiTransport,
     cdpConnection: CdpConnection,
     selfTargetId: string,
@@ -113,16 +127,6 @@ export class BidiServer extends EventEmitter<BidiServerEvents> {
     );
   }
 
-  #processOutgoingMessage = async (messageEntry: OutgoingBidiMessage) => {
-    const message = messageEntry.message as any;
-
-    if (messageEntry.channel !== null) {
-      message['channel'] = messageEntry.channel;
-    }
-
-    await this.#transport.sendMessage(message);
-  };
-
   /**
    * Sends BiDi message.
    */
@@ -130,13 +134,9 @@ export class BidiServer extends EventEmitter<BidiServerEvents> {
     this.#messageQueue.add(messageEntry);
   }
 
-  close(): void {
+  close() {
     this.#transport.close();
   }
-
-  #handleIncomingMessage = async (message: Message.RawCommandRequest) => {
-    this.#commandProcessor.processCommand(message);
-  };
 
   getBrowsingContextStorage(): BrowsingContextStorage {
     return this.#browsingContextStorage;
--- a/src/bidiMapper/CdpConnection.ts
+++ b/src/bidiMapper/CdpConnection.ts
@@ -15,6 +15,10 @@
  * limitations under the License.
  */
 
+/**
+ * @fileoverview CDP interfaces and types that BiDi Mapper expects.
+ */
+
 import type {ProtocolMapping} from 'devtools-protocol/types/protocol-mapping.js';
 
 import {EventEmitter} from '../utils/EventEmitter.js';
@@ -23,18 +27,14 @@ type CdpEvents = {
   [Property in keyof ProtocolMapping.Events]: ProtocolMapping.Events[Property][0];
 };
 
-/**
- * CDP interfaces that BiDi Mapper expects.
- */
-
 export interface CdpConnection {
   browserClient(): CdpClient;
   getCdpClient(sessionId: string): CdpClient;
 }
 
 export interface CdpClient extends EventEmitter<CdpEvents> {
-  sendCommand<T extends keyof ProtocolMapping.Commands>(
-    method: T,
-    ...params: ProtocolMapping.Commands[T]['paramsType']
-  ): Promise<ProtocolMapping.Commands[T]['returnType']>;
+  sendCommand<CdpMethod extends keyof ProtocolMapping.Commands>(
+    method: CdpMethod,
+    ...params: ProtocolMapping.Commands[CdpMethod]['paramsType']
+  ): Promise<ProtocolMapping.Commands[CdpMethod]['returnType']>;
 }
--- a/src/bidiTab/bidiTab.ts
+++ b/src/bidiTab/bidiTab.ts
@@ -74,16 +74,16 @@ const waitSelfTargetIdPromise = waitSelfTargetId();
 })();
 
 function createCdpConnection() {
-  // A CdpTransport implementation that uses the window.cdp bindings
-  // injected by Target.exposeDevToolsProtocol.
+  /**
+   * A CdpTransport implementation that uses the window.cdp bindings
+   * injected by Target.exposeDevToolsProtocol.
+   */
   class WindowCdpTransport implements ITransport {
     #onMessage: ((message: string) => void) | null = null;
 
     constructor() {
       window.cdp.onmessage = (message: string) => {
-        if (this.#onMessage) {
-          this.#onMessage.call(null, message);
-        }
+        this.#onMessage?.call(null, message);
       };
     }
 
@@ -116,9 +116,9 @@ async function createBidiServer(selfTargetId: string) {
     constructor() {
       window.onBidiMessage = (messageStr: string) => {
         log(LogType.bidi, 'received ◂', messageStr);
-        let messageObj;
+        let messageObject: Message.RawCommandRequest;
         try {
-          messageObj = WindowBidiTransport.#parseBidiMessage(messageStr);
+          messageObject = WindowBidiTransport.#parseBidiMessage(messageStr);
         } catch (e: any) {
           // Transport-level error does not provide channel.
           this.#respondWithError(
@@ -129,9 +129,7 @@ async function createBidiServer(selfTargetId: string) {
           );
           return;
         }
-        if (this.#onMessage) {
-          this.#onMessage.call(null, messageObj);
-        }
+        this.#onMessage?.call(null, messageObject);
       };
     }
 
@@ -212,20 +210,20 @@ async function createBidiServer(selfTargetId: string) {
     }
 
     static #parseBidiMessage(messageStr: string): Message.RawCommandRequest {
-      let messageObj: any;
+      let messageObject: Message.RawCommandRequest;
       try {
-        messageObj = JSON.parse(messageStr);
+        messageObject = JSON.parse(messageStr);
       } catch {
         throw new Error('Cannot parse data as JSON');
       }
 
-      const parsedType = WindowBidiTransport.#getJsonType(messageObj);
+      const parsedType = WindowBidiTransport.#getJsonType(messageObject);
       if (parsedType !== 'object') {
         throw new Error(`Expected JSON object but got ${parsedType}`);
       }
 
-      // Extract amd validate id, method and params.
-      const {id, method, params} = messageObj;
+      // Extract and validate id, method and params.
+      const {id, method, params} = messageObject;
 
       const idType = WindowBidiTransport.#getJsonType(id);
       if (idType !== 'number' || !Number.isInteger(id) || id < 0) {
@@ -244,7 +242,7 @@ async function createBidiServer(selfTargetId: string) {
         throw new Error(`Expected object params but got ${paramsType}`);
       }
 
-      let channel = messageObj.channel;
+      let channel = messageObject.channel;
       if (channel !== undefined) {
         const channelType = WindowBidiTransport.#getJsonType(channel);
         if (channelType !== 'string') {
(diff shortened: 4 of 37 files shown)


thiagowfx:
Side note: `tested_nested_browsing_context.py` had to be reverted because of an E2E timeout that was only happening in CI (=not reproducible locally).

src/utils/deferred.ts
```diff
@@ -50,7 +50,7 @@ export class Deferred<T> implements Promise<T> {
     this.#resolve(value);
   }
 
-  public reject(reason: any) {
+  public reject(reason: T) {
```
sadym-chromium, line 53: reject is supposed to have a error, while resolve should have event message. So this line is not valid.

tests/conftest.py
```diff
@@ -74,6 +78,7 @@ async def context_id(websocket):
 
 @pytest_asyncio.fixture
 async def another_context_id(websocket):
+    """Return the context id from a newly created browsing context."""
```
sadym-chromium, line 81:
```suggestion
    """Create a browsing context and return its id."""
```

tests/conftest.py
```diff
@@ -85,12 +90,14 @@ async def another_context_id(websocket):
 
 @pytest_asyncio.fixture
 async def page_with_nested_iframe_url():
+    """Return a page with a nested iframe of about:blank."""
```
sadym-chromium, line 93:
```suggestion
    """Return a page URL with a nested iframe of about:blank."""
```

tests/conftest.py
```diff
@@ -85,12 +90,14 @@ async def another_context_id(websocket):
...
     return 'data:text/html,<h1>MAIN_PAGE</h1>' \
            '<iframe src="about:blank" />'
 
 
 @pytest_asyncio.fixture
 async def iframe_id(context_id, websocket, page_with_nested_iframe_url):
+    """Return the iframe id from a page with a nested iframe of about:blank."""
```
sadym-chromium, line 100:
```suggestion
    """Navigate to a page with nested iframe of `about:blank`, and return iframe's BrowserContextId."""
```

tests/test_browsing_context.py
```diff
@@ -254,12 +253,11 @@ async def test_browsingContext_getTreeWithNestedCrossOriginContexts_contextsRetu
...
 
 
-# TODO(sadym): make offline.
 @pytest.mark.asyncio
 async def test_browsingContext_afterNavigation_getTreeWithNestedCrossOriginContexts_contextsReturned(
         websocket, context_id):
-    nested_iframe = 'https://example.com/'
-    another_nested_iframe = 'https://example.org/'
+    nested_iframe = 'about:blank'
+    another_nested_iframe = 'about:about'
```
sadym-chromium, line 260: Are they considered to be cross-origin?

src/utils/deferred.ts
thiagowfx: Are you sure?

reject() calls this.#reason with the same `reason` argument, and the signature of reason is `#reject: (value: T) => void = () => {};`. So in practice, this is `T` as well.

tests/test_browsing_context.py
thiagowfx: It it's a requirement, I'll revert it just to be safe.

src/utils/deferred.ts
sadym-chromium: I assume you meant "`reject()` calls **`this.#reject`**". You are right, there is an issue, but it is in the `#reject` method. It should be `#reject: (value: unknown)`. I committed fix for it in this PR.

thiagowfx: Thanks, rebased

.eslintrc.js
```diff
@@ -10,7 +10,7 @@ module.exports = {
     globalThis: false,
   },
   parser: '@typescript-eslint/parser',
-  plugins: ['@typescript-eslint', 'import', 'mocha'],
+  plugins: ['@typescript-eslint', 'import', 'mocha', 'simple-import-sort'],
```
Lightning00Blade, line 13: I have not look at `simple-import-sort`, `import` already has similar rule that has a fixer.

src/cdp/cdpConnection.ts
```diff
@@ -78,15 +79,15 @@ export class CdpConnection {
...
+  sendCommand<CdpMethod extends keyof ProtocolMapping.Commands>(
+    method: CdpMethod,
+    params: ProtocolMapping.Commands[CdpMethod]['paramsType'][0] | undefined,
     sessionId: string | null
   ): Promise<object> {
     return new Promise((resolve, reject) => {
       const id = this.#nextId++;
       this.#commandCallbacks.set(id, {resolve, reject});
-      const messageObj: CdpMessage = {id, method, params};
+      const messageObj: CdpMessage<typeof method> = {id, method, params};
```
Lightning00Blade, line 90: Can we use just `CdpMethod` method here?

thiagowfx: Good catch. Done!

package.json
```diff
@@ -59,6 +59,7 @@
     "eslint-plugin-import": "2.26.0",
     "eslint-plugin-mocha": "10.1.0",
     "eslint-plugin-prettier": "4.2.1",
+    "eslint-plugin-simple-import-sort": "^10.0.0",
```
Lightning00Blade, line 62: I think you changed for the other plugin, so let's remove this from dependencies

src/protocol/protocol.ts
```diff
@@ -797,12 +799,12 @@ export namespace CDP {
   };
 
   export type SendCommandParams = {
-    cdpMethod: string;
+    cdpMethod: keyof ProtocolMapping.Commands;
```
Lightning00Blade, line 802: If you look at the top there is a comment that this file should not have dependencies. I have no context about why exactly.

Lightning00Blade (commented):
Some comments, but looks good, there are some python files changes + .vscode ones that I don't have context/knowledge how to handle.

.eslintrc.js
```diff
@@ -44,6 +44,11 @@ module.exports = {
     '@typescript-eslint/no-non-null-assertion': 'off',
     '@typescript-eslint/no-this-alias': 'off',
     'func-names': 'error',
+    'import/first': 'error',
+    'import/newline-after-import': 'error',
+    'import/no-duplicates': 'error',
+    'import/order': ['error', {'newlines-between': 'always'}],
+    'no-console': 'error',
```
sadym-chromium, line 51: Actually, I use `console.log` in [`src/bidiTab/mapperTabPage.ts`](https://github.com/GoogleChromeLabs/chromium-bidi/blob/1a62b7ea8822291bd88aabe42a3fbf608f232756/src/bidiTab/mapperTabPage.ts#L67). Current `logger` does not log anything on the server side, which makes debugging really hard. Consider adding `debug` flag on the server side, which would send all the logs to the server. Until that `console.log` is the only way to get the logs.

.eslintrc.js
```diff
@@ -44,6 +44,11 @@ module.exports = {
     '@typescript-eslint/no-non-null-assertion': 'off',
     '@typescript-eslint/no-this-alias': 'off',
     'func-names': 'error',
+    'import/first': 'error',
+    'import/newline-after-import': 'error',
+    'import/no-duplicates': 'error',
+    'import/order': ['error', {'newlines-between': 'always'}],
```
sadym-chromium, line 50: Please make sure this is done by some automatic tool (and I would strictly prefer not to add one more tool).

src/bidiMapper/BidiServer.ts
```diff
@@ -112,31 +127,17 @@ export class BidiServer extends EventEmitter<BidiServerEvents> {
     );
   }
 
-  #processOutgoingMessage = async (messageEntry: OutgoingBidiMessage) => {
```
sadym-chromium, line 115: what is the ratio here?

src/bidiMapper/CdpConnection.ts
```diff
@@ -15,25 +15,26 @@
  * limitations under the License.
  */
 
-import {EventEmitter} from '../utils/EventEmitter.js';
+/**
+ * @file CDP interfaces and types that BiDi Mapper expects.
+ */
```
sadym-chromium, line 20: This comment belongs to the `CdpConnection`, why moving away?

sadym-chromium, line 20: Why is file doc preferable over the interface's?

src/bidiMapper/bidiMapper.ts
```diff
@@ -21,7 +21,7 @@
  * TODO: eslint rule for this.
  */
 
-export {CdpConnection, CdpClient} from './CdpConnection.js';
 export {BidiServer} from './BidiServer.js';
 export {BidiTransport} from './BidiTransport.js';
+export {CdpClient, CdpConnection} from './CdpConnection.js';
```
sadym-chromium, line 26: Unless it's done by automatic tool (and I believe it is not), I'd avoid this import re-arrangements, as it will end up with reformat mess.

src/bidiTab/bidiTab.ts
```diff
@@ -282,7 +285,11 @@ class BidiParserImpl implements BidiParser {
     return Parser.Script.parseDisownParams(params);
   }
   parseSendCommandParams(params: object): CDP.SendCommandParams {
-    return Parser.CDP.parseSendCommandParams(params);
+    return Parser.CDP.parseSendCommandParams(params) as {
+      cdpMethod: keyof ProtocolMapping.Commands;
+      cdpParams: object;
+      cdpSession?: any;
+    };
```
sadym-chromium, line 292: Why not to change `SendCommandParams` accordingly?

src/cdp/cdpClient.spec.ts
```diff
@@ -69,7 +71,7 @@ describe('CdpClient tests.', () => {
     await mockCdpServer.emulateIncomingMessage({id: 0, result: {}});
 
     // Assert 'cdpClient' resolved message promise.
-    chai.assert.eventually.deepEqual(commandPromise, {});
+    expect(commandPromise).to.eventually.equal({});
```
sadym-chromium, line 74: Doesn't `equal` compare references?

.eslintrc.js
Lightning00Blade: FYI I need to this VsCode setting for it to work on save consistently in Puppeteer.
```
"editor.codeActionsOnSave": {
  "source.fixAll.eslint": true
},
 ```

File: .eslintrc.js (hunk 10 above)
Comment (thiagowfx) [reply .eslintrc.js:51]: This just means we cannot _check in_ code with `console.*`. You can still use it locally.

I will downgrade it to `warn` instead, which is a sweet spot.

File: .eslintrc.js (hunk 11 above)
Comment (thiagowfx) [reply .eslintrc.js:50]: > Please make sure this is done by some automatic tool (and I would strictly prefer not to add one more tool).

This is not a new tool: It was already preexisting, AND

Yes, it's done automatically!

Unlike the other one, this one doesn't care about sorting alphabetically, so it doesn't get in the way.

File: src/bidiMapper/CdpConnection.ts (hunk 13 above)
Comment (thiagowfx) [reply src/bidiMapper/CdpConnection.ts:20]: /**
 * CDP interfaces that BiDi Mapper expects.
 */

export interface CdpConnection {
  browserClient(): CdpClient;
  getCdpClient(sessionId: string): CdpClient;
}

export interface CdpClient extends EventEmitter<CdpEvents> {
  sendCommand<T extends keyof ProtocolMapping.Commands>(
    method: T,
    ...params: ProtocolMapping.Commands[T]['paramsType']
  ): Promise<ProtocolMapping.Commands[T]['returnType']>;
}


...means that the JSDOC comment is only set to the first interface. But it really applies to both. So, a @fileoverview comment is more appropriate.

In fact, it's @fileoverview, not @file. Will go ahead and fix that too (ref: https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler)

File: src/bidiMapper/BidiServer.ts (hunk 12 above)
Comment (thiagowfx) [reply src/bidiMapper/BidiServer.ts:115]: Did you mean rationale?

Those are member variables, not methods. Just grouped the member variables closer together - at the top of the class.

Before I did that I tried to convert them to functions (`async #processOutgoingMessages(messageEntry: OutgoingBidiMessage) {...` but that resulted in an error.

File: .eslintrc.js [hunk 17]
```diff
@@ -44,6 +44,11 @@ module.exports = {
     '@typescript-eslint/no-non-null-assertion': 'off',
     '@typescript-eslint/no-this-alias': 'off',
     'func-names': 'error',
+    'import/first': 'error',
```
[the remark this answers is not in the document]
thiagowfx: Sure! The idea of this change is to alleviate the pains you were having with the other formatter, that didn't have an autofixer. This one does, and those changes make the imports become more deterministic - which will result in less bikeshedding in future reviews for us.

src/bidiTab/bidiTab.ts
thiagowfx: Done!

src/cdp/cdpClient.spec.ts
thiagowfx: The docs use `equal` all over the place: https://www.chaijs.com/plugins/chai-as-promised/

It seems to be: `be` for `chai`, and `equal` for `chai-as-promised`.

thiagowfx:
This is now ready to go and the tests are passing :D

thiagowfx:
Discussed offline: The formatting changes will be splitted to its own PR. All other changes will remain here.

thiagowfx:
Done. This PR contains no intentional formatting changes.

src/bidiTab/mapperTabPage.ts
```diff
@@ -17,8 +17,75 @@
 import {LogType} from '../utils/log.js';
 
 /** HTML source code for the user-facing Mapper tab. */
-const mapperPageSource =
```
sadym-chromium, line 20: It was minimised to reduce the Mapper size, as this string will be delivered without any compression. This increases the Mapper size for `518` bytes (`1484` - `966`) .

src/protocol/protocol.ts
sadym-chromium: I believe it is because of the way Puppeteer uses Mapper. @OrKoN knows for sure.

OrKoN: Type dependency should be alright because it is compiled away. The reason for no deps is that the file will be run by the browser so cannot import node_modules. E.g., put a note there so no one adds zod or smth like that in there. Thanks @Lightning00Blade for spotting.

src/utils/EventEmitter.spec.ts
```diff
@@ -19,7 +19,7 @@ import {EventEmitter} from './EventEmitter.js';
 import {expect} from 'chai';
 import sinon from 'sinon';
 
-describe('EventEmitter', () => {
+describe('test EventEmitter', () => {
```
OrKoN, line 22: is the `test` needed? Looks redundant

src/utils/processingQueue.ts
```diff
@@ -55,8 +55,7 @@ export class ProcessingQueue<T> {
           .catch((e) => {
             this.#logger?.(LogType.system, 'Event was not processed:', e);
             this.#catch(e);
-          })
-          .finally();
```
sadym-chromium, line 59: If provided `_catch` throws exception, the message processing will stop, but it should not.

src/utils/EventEmitter.spec.ts
OrKoN: it looks like it is written everywhere, wdyt about removing it everywhere?

src/utils/processingQueue.ts
thiagowfx: I asked in an internal TypeScript chat (before removing it) and people said this is a no-op. Let me double check.

src/bidiTab/mapperTabPage.ts
thiagowfx: Oh, I see. Reverted.

src/utils/EventEmitter.spec.ts
thiagowfx: It is redundant, I just chose the style that was already predominant ('test' everywhere) and made everything consistent.

I'll remove all "test" then.

src/utils/processingQueue.ts
thiagowfx: Since this is the only blocker, for now I re-added the `.finally()` just so that we can submit this PR before the weekend.

sadym-chromium (approved):

src/utils/processingQueue.ts
sadym-chromium: Oh, you are right! `finally` does not change the state of promise, so another catch is needed there. Out of scope of this PR of cause.

Issue #514, opened 2023-03-06
Implement: `browsingContext.captureScreenshot`


thiagowfx:
- Docs: https://w3c.github.io/webdriver-bidi/#command-browsingContext-captureScreenshot
- CDP: https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-captureScreenshot

## Tracking

- [x] Spec subsubsubsection 7.2.3.1
- [x] WPT tests: `/webdriver/tests/bidi/browsing_context/capture_screenshot/*`
- [x] Add implementation to Command Processor
- [x] Associated CDP command: [`Page.captureScreenshot`](https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-captureScreenshot) -> need viewport
- [x] Underneath mechanism: `/session/{id}/screenshot` (via WebDriver) --> this is merely an implementation detail
- [x] Add sample gradient golden page for testing --> use it to compare (base64-wise) with the implementation -> canvas.base64
- [x] Implement `scrollIntoView`
- [x] Closed frames
- [x] https://github.com/GoogleChromeLabs/chromium-bidi/pull/580
- [x] Height is slightly different in WPT tests. Maybe it's related to the Chromium bar (automated software) or display differences: https://github.com/GoogleChromeLabs/chromium-bidi/issues/547
- [x] Finish milestone: https://github.com/GoogleChromeLabs/chromium-bidi/milestone/4
  - [x] ~~iframes: set [`captureBeyondViewport`](https://github.com/GoogleChromeLabs/chromium-bidi/issues/514#issuecomment-1470303113) to true~~
  - [x] ~~Tab focus: https://github.com/GoogleChromeLabs/chromium-bidi/pull/796~~
  - [x] Tab focus: Focus originally focused tab after taking the screenshot, or change the spec: https://github.com/w3c/webdriver-bidi/issues/171 / https://github.com/w3c/webdriver-bidi/issues/440
  - [x] Throw error on unsupported edge cases, as per the spec: ~~https://github.com/GoogleChromeLabs/chromium-bidi/pull/797~~ / https://github.com/w3c/webdriver-bidi/issues/441
  - [x] Re-enable headful screenshot tests, if at all possible: https://github.com/GoogleChromeLabs/chromium-bidi/pull/813, https://github.com/GoogleChromeLabs/chromium-bidi/pull/823 and https://github.com/web-platform-tests/wpt/pull/40382 
[...]

PR #515, opened 2023-03-06
Implement `browsingContext.captureScreenshot` command

thiagowfx:
Bug: #514
Spec: https://w3c.github.io/webdriver-bidi/#command-browsingContext-captureScreenshot

### DIFF
--- a/src/bidiMapper/domains/context/browsingContextProcessor.ts
+++ b/src/bidiMapper/domains/context/browsingContextProcessor.ts
@@ -237,6 +237,15 @@ export class BrowsingContextProcessor {
     );
   }
 
+  async process_browsingContext_captureScreenshot(
+    params: BrowsingContext.CaptureScreenshotParameters
+  ): Promise<BrowsingContext.CaptureScreenshotResult> {
+    const context = this.#browsingContextStorage.getKnownContext(
+      params.context
+    );
+    return context.captureScreenshot();
+  }
+
   async #getRealm(target: Script.Target): Promise<Realm> {
     if ('realm' in target) {
       return this.#realmStorage.getRealm({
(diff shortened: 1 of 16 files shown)


src/bidiMapper/domains/context/browsingContextProcessor.ts
```diff
@@ -237,6 +237,21 @@ export class BrowsingContextProcessor {
     );
   }
 
+  async process_browsingContext_captureScreenshot(
+    params: BrowsingContext.CaptureScreenshotParameters
+  ): Promise<BrowsingContext.CaptureScreenshotResult> {
+    const context = this.#browsingContextStorage.getKnownContext(
+      params.context
+    );
+    // TODO: return context.captureScreenshot(params);
```
sadym-chromium, line 246: Correct. And `context.captureScreenshot` in turn should call `await this.#cdpClient.sendCommand('Page.captureScreenshot', {})`

thiagowfx:
c.f. https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-captureScreenshot

thiagowfx:
Capture screenshot base64:
image

thiagowfx:
Other than improving the E2E test, this is complete.

thiagowfx:
@sadym-chromium Test is now fully complete as we discussed it, comparing the output with a hard-coded golden string.

thiagowfx:
The issue is that the screenshot seems to be different depending on the environment where the test is run, likely because of the viewport size. For example, on my laptop it's currently passing, but on GitHub actions it is failing.

There are no viewport params at the moment (in the spec, that is).

Should we revert to comparing it with `ANY_STR`?

sadym-chromium:
> The issue is that the screenshot seems to be different depending on the environment where the test is run, likely because of the viewport size. For example, on my laptop it's currently passing, but on GitHub actions it is failing.
> 
> There are no viewport params at the moment (in the spec, that is).
> 
> Should we revert to comparing it with `ANY_STR`?

I'd rather get a piece of result, let's say 100x100 pixels, and compare it vs golden standard.

sadym-chromium:
another option is to get the window size, generate expected image and compare the result to it.

sadym-chromium (approved):

thiagowfx:
We need to focus the active tab with CDP Command `Page.bringToFront`, or else the screenshot action blocks in headful mode. This was discussed off-line with @sadym-chromium.

I confirm this works as we predicted.

thiagowfx:
Added a TODO to reflect the custom tab focus quirk.

PR #517, opened 2023-03-08
chore: remove console.error statement

Lightning00Blade:
Remove a forgotten `console.error`. If you think we should use the logger here, happy to make the change.

### DIFF
--- a/src/bidiMapper/CommandProcessor.ts
+++ b/src/bidiMapper/CommandProcessor.ts
@@ -234,7 +234,6 @@ export class CommandProcessor extends EventEmitter<CommandProcessorEvents> {
         );
       } else {
         const error = e as Error;
-        console.error(error);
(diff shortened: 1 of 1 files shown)


src/bidiMapper/CommandProcessor.ts
```diff
@@ -234,7 +234,6 @@ export class CommandProcessor extends EventEmitter<CommandProcessorEvents> {
         );
       } else {
         const error = e as Error;
-        console.error(error);
```
thiagowfx, line 237: > Remove a forgotten console.error. If you think we should use the logger here, happy to make the change.

Considering we already have the `logger` in the constructor, it should be easy to call it here. You can store it as a private member variable in the constructor. Let's do that.

thiagowfx (approved):

PR #519, opened 2023-03-08
miscellaneous protocol improvements

thiagowfx:
- ~remove redundant comments, as they are hard to keep up-to-date as the spec evolves: http://go/java-practices/comments#what_for~
- `params: unknown` -> `params: object`
- add type to attributes: from "unknown" to "Record<string, string>" (`?attributes: {*text => text}`)
- add documentation links to all top-level modules
- group methods by module
- introduce a type for all BiDi commands

### DIFF
--- a/src/bidiMapper/CommandProcessor.ts
+++ b/src/bidiMapper/CommandProcessor.ts
@@ -42,9 +42,9 @@ export interface BidiParser {
   parseDisownParams(params: object): Script.DisownParameters;
   parseSendCommandParams(params: object): CDP.SendCommandParams;
   parseGetSessionParams(params: object): CDP.GetSessionParams;
+  parseSubscribeParams(params: object): Session.SubscribeParameters;
   parseNavigateParams(params: object): BrowsingContext.NavigateParameters;
   parseGetTreeParams(params: object): BrowsingContext.GetTreeParameters;
-  parseSubscribeParams(params: object): Session.SubscribeParameters;
   parseCreateParams(params: object): BrowsingContext.CreateParameters;
   parseCloseParams(params: object): BrowsingContext.CloseParameters;
 }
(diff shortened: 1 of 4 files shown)


src/protocol/protocol.ts
```diff
@@ -136,54 +152,31 @@ export namespace CommonDataTypes {
     sharedId: string;
   };
 
-  // UndefinedValue = {
```
OrKoN, line 139: I actually like those comments because they make it clear what spec text was used to author types (not blocking opinion).

OrKoN (approved):

src/bidiMapper/CommandProcessor.ts
```diff
@@ -42,9 +42,9 @@ export interface BidiParser {
   parseDisownParams(params: object): Script.DisownParameters;
   parseSendCommandParams(params: object): CDP.SendCommandParams;
   parseGetSessionParams(params: object): CDP.GetSessionParams;
+  parseSubscribeParams(params: object): Session.SubscribeParameters;
```
sadym-chromium, line 45: I wonder what this re-arrange improves

src/protocol/protocol.ts
sadym-chromium: +1 for this comment.

src/bidiMapper/CommandProcessor.ts
thiagowfx: It helps when adding new methods. All `BrowsingContext` method signatures are now grouped together.

Previously, `Session` was in the middle of `BrowsingContext`.

src/protocol/protocol.ts
thiagowfx: OK, I'll revert the comment removals for now for the sake of being less contentious.

That said, I think it's not worth the effort to add these comments, plus they can get outdated easily as the spec evolves (e.g. whenever new subfields/keys are added to compound objects).

thiagowfx: All comments were re-added.

PR #521, opened 2023-03-09
revert CI mapper change back to not using puppeteer/browsers

thiagowfx:
For some reason, using chrome latest is making the CI fail. Revert back to chrome dev.

Bug: #361

### DIFF
--- a/.github/workflows/wpt-mapper.yml
+++ b/.github/workflows/wpt-mapper.yml
@@ -48,10 +48,15 @@ jobs:
       - name: Set up hosts
         run: ./wpt make-hosts-file | sudo tee -a /etc/hosts
         working-directory: wpt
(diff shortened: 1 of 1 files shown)


thiagowfx:
Note: I went ahead with this in order to unblock CI and make it green again.

OrKoN:
Do you have links to failures? the bug seems to be unrelated

thiagowfx:
Here's one: https://github.com/GoogleChromeLabs/chromium-bidi/actions/runs/4368453683/jobs/7641057899

OrKoN:
I see. Note that if you installing the latest without pinning a version there is always a risk that you get a broken latest build. The latest is like the true latest version available, fresh from main.

PR #523, opened 2023-03-09
eslint: standardize import order with the import/ autofixer

thiagowfx:
Previously we were using `simple-import-sort` which did not support an autofixer and thus was causing development pain. The new setup includes an autofixer and does not fiddle with sorting imports, but merely with grouping similar groups together.


thiagowfx:
Note: This was agreed off-line with @sadym-chromium after the retrospective meeting.

sadym-chromium (approved):

PR #524, opened 2023-03-10
run scripts: augment with env variables, and have runWPT.sh update ex…

thiagowfx:
…pectations

To make it easier to switch between headful and headless. For example:

HEADLESS=false ./runWPT.sh


runWPT.sh
```diff
@@ -1,27 +1,47 @@
...
-  --timeout-multiplier 8 \
-  chromium \
+  --webdriver-arg="--headless=$HEADLESS" \
+  --log-wptreport "$WPT_REPORT" \
+  --manifest "$MANIFEST" \
+  --metadata "$WPT_METADATA" \
+  "$PRODUCT" \
   "$@")
 
+echo "--> Updating expectations..."
```
thiagowfx, line 37: If you want, we could also add an `UPDATE_EXPECTATIONS` variable to control whether we run this part.

thiagowfx: Done

runWPT.sh
```diff
@@ -1,31 +1,60 @@
...
 
 # Go to the project root folder.
 (cd "$(dirname "$0")/" && \
-# Create a folder to store logs.
-mkdir -p "$LOG_DIR" && \
-npm run build && \
 ./wpt/wpt run \
   --binary "$WPT_BROWSER_PATH" \
   --webdriver-binary runBiDiServer.sh \
-  --webdriver-arg="--verbose" \
```
thiagowfx, line 15: These were invalid! They are not correct options for Bidi Server (only for classic chromedriver).

sadym-chromium (approved):

sadym-chromium:
It would be nice to have `chromium_path` param as well for local testing.

thiagowfx:
`executable_path` would be `WPT_BROWSER_PATH`, correct?

sadym-chromium:
Yes, but Mapper Server does not support that argument yet. Can be out of scope.

thiagowfx:
I see. Let's do that in a follow up then.

PR #525, opened 2023-03-10
wpt: browsing_context/capture_screenshot/frame.py: mark as failing

thiagowfx:
This test is currently failing for the time being, likely due to the automation info bar in chrome.

Bug: #514


thiagowfx:
It fails on `main` but passes here...so weird.

https://github.com/GoogleChromeLabs/chromium-bidi/actions/runs/4379798624/jobs/7666342651

Closing then.

Issue #518, opened 2023-03-08
Implement `browsingContext.print` (print to PDF)


sadym-chromium:
Implement [Add browsingContext.print command by jgraham · Pull Request #363 · w3c/webdriver-bidi](https://github.com/w3c/webdriver-bidi/pull/363/files).

CDP method: [`Page.printToPDF`](https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-printToPDF).
Spec: https://w3c.github.io/webdriver-bidi/#command-browsingContext-print

## Tracking

- [x] Basic implementation
- [x] ~CDP does not have a `shrinkToFit` equivalent (c.f. [comment](https://github.com/GoogleChromeLabs/chromium-bidi/issues/518#issuecomment-1465967305))~ it is called `preferCSSPageSize`
- [x] Implement `shrinkToFit`
- [x] #577 
- [x] [Issue 1430696: CDP `Page.printToPDF` crashes on tiny pages](https://bugs.chromium.org/p/chromium/issues/detail?id=1430696)
- [x] [CDP `Page.printToPDF` throws invalid parameters exception on pages with tiny dimensions (e.g. 1x1 pixel) (cont.)](https://bugs.chromium.org/p/chromium/issues/detail?id=1454035)
- [x] (optional) Add E2E test: print to pdf -> open it in the browser (either `application/pdf` or `file:///`) -> take screenshot -> compare golden: https://chat.google.com/room/AAAAw7biRSY/UdLyoDHCNgM. Or similar: https://github.com/GoogleChromeLabs/chromium-bidi/pull/912 and #914
- [ ] Fix WPT tests [`/webdriver/tests/bidi/browsing_context/print/`](https://github.com/GoogleChromeLabs/chromium-bidi/tree/main/wpt-metadata/mapper/headful/webdriver/tests/bidi/browsing_context/print)

thiagowfx:
@sadym-chromium `shrinkToFit` does not have a CDP equivalent. What do we do in these situations? File a FR against CDP?

sadym-chromium:
> What do we do in these situations?

In general there are 3 options:
1. Implement in CDP.
1. Implement on top of existing CDP functionality + some extra logic.
1. Roll back the spec part.

thiagowfx:
> 1. Implement in CDP.

How can we do so? This was also suggested by @whimboo on https://github.com/web-platform-tests/wpt/pull/38931#issuecomment-1468350079

[...]

PR #526, opened 2023-03-10
Implement `browsingContext.print` command

thiagowfx:
Bug: #518
Spec: https://w3c.github.io/webdriver-bidi/#command-browsingContext-print
CDP method: https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-printToPDF

## Status

- [x] [UP: #518] CDP does not have a `shrinkToFit` equivalent.
- [x] Use `.refine()` / `.transform()` to check the first number is lower than the second as per the spec (https://w3c.github.io/webdriver/#dfn-parse-a-page-range)
- [x] Do we need E2E tests?

### DIFF
--- a/src/protocol-parser/protocol-parser.ts
+++ b/src/protocol-parser/protocol-parser.ts
@@ -601,13 +601,15 @@ export namespace BrowsingContext {
     | NavigateCommand
     | CreateCommand
     | CloseCommand
-    | CaptureScreenshotCommand;
+    | CaptureScreenshotCommand
+    | PrintCommand;
   export type CommandResult =
     | GetTreeResult
     | NavigateResult
     | CreateResult
     | CloseResult
-    | CaptureScreenshotResult;
+    | CaptureScreenshotResult
+    | PrintResult;
   export type Event =
     | LoadEvent
     | DomContentLoadedEvent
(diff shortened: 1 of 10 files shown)


thiagowfx:
See `Status` subsection in the first PR comment to see what's still missing.

src/protocol-parser/protocol-parser.ts
```diff
@@ -744,6 +746,68 @@ export namespace BrowsingContext {
...
+   */
+  const PrintPageRangesSchema = zod
+    .array(
+      zod.union([
+        zod
+          .string()
+          .min(1)
+          .regex(/^((?:\d+)|(?:\d+[-])|(?:[-]\d+)|(?:\d+[-]\d+))$/),
+        // TODO: Use `.refine()` to check the first number is lower than the second as per the spec.
+        zod.preprocess((val) => String(val), zod.number().int().min(1)),
```
sadym-chromium, line 780: > If range is not an integer or is less than 0, return [error](https://w3c.github.io/webdriver/#dfn-error) with [error code](https://w3c.github.io/webdriver/#dfn-error-code) [invalid argument](https://w3c.github.io/webdriver/#dfn-invalid-argument).

1. It should be non-negative.
2. Do we need a preprocessor for the number?

```suggestion
        zod.number().int().min(0),
```

src/protocol-parser/protocol-parser.ts
```diff
@@ -744,6 +746,68 @@ export namespace BrowsingContext {
...
+   * 1-based index, e.g. 1, '5', '2-4', '6-', '-3'
+   * @see https://w3c.github.io/webdriver/#dfn-parse-a-page-range
+   */
+  const PrintPageRangesSchema = zod
+    .array(
+      zod.union([
+        zod
+          .string()
+          .min(1)
+          .regex(/^((?:\d+)|(?:\d+[-])|(?:[-]\d+)|(?:\d+[-]\d+))$/),
```
sadym-chromium, line 778: Can we use a preprocess here instead of regex?

sadym-chromium (approved):

src/protocol-parser/protocol-parser.ts
thiagowfx: It says:

> If page ranges is not [empty](https://infra.spec.whatwg.org/#list-empty), let pages be the result of [trying](https://w3c.github.io/webdriver/#dfn-try) to [parse a page range](https://w3c.github.io/webdriver/#dfn-parse-a-page-range) with page ranges and the number of pages contained in pdf data, then remove any pages from pdf data whose **one-based index** is not contained in pages.

...on the BiDi side. But you're right, the WebDriver side says 0. I am curious what "0" would mean in this context, but regardless, that's transparent from an API point of view, we're just forwarding it to CDP. Done.

thiagowfx: 2) The preprocessor is meant to convert everything to String, so that we can have a single logic to handle both `number` and `string` for validation.

Otherwise, in our validation code, we will need to either (i) check the type (e.g. with `typeof`) and have two branches, or (ii) cast number to string anyway.

thiagowfx:
Side note: The `print` method does not seem to be deterministic, so I will refrain from adding an E2E test with a golden:
image

Instead, I'll just write a basic one that compares it with `ANY_STR`.

src/protocol-parser/protocol-parser.ts
thiagowfx: Array of integers:

- (1) is not expressive enough: for example, we cannot express both '1' and '3-5' with one
- (2) is not efficient: for example, if we printed '1-1000000' we would have to have an `[1,2,...,1000000]` array

That's how I read your suggestion; please correct me if you thought of another representation. String seems more flexible for what we're looking for, especially considering that the CDP param is also a string.

thiagowfx:
For the validation, I'll try to use `.refine()` from `zod`.

thiagowfx:
```
      zod.union([
        zod.string().min(1),
        zod.coerce.string(zod.number().int().min(0)),
      ])
```

...works great! This required an update to `zod` in `package.json` to a newer version that supports `zod.coerce` ([docs](https://zod.dev/?id=coercion-for-primitives).)

thiagowfx:
Actually, that didn't work. I had to use this:

```
    .array(
      zod.union([
        zod.string().min(1),
        zod
          .number()
          .int()
          .min(0)
          .transform((n) => String(n)),
      ])
    )
```

thiagowfx:
Also, there's no need to validate cases like `4-3` at the mapper level, because CDP already does so:

```
>       assert resp == {'id': 1, 'result': {'data': ANY_STR}}
E       AssertionError: assert equals failed
E         {                                                    {'id': 1, 'result': {'data': ANY_STR}}              
E           'error': 'unknown error',                                                                              
E           'id': 1,                                                                                               
E           'message': 'Page range is invalid (start > end)',                                                      
E         }
```

thiagowfx:
Will track the `shrinkToFit` CDP change on https://github.com/GoogleChromeLabs/chromium-bidi/issues/518

thiagowfx:
I can't decide whether it makes more sense to have `.default([])` be part of `PrintPageRangesSchema` or part of `PrintParametersSchema`.  I went with `PrintParametersSchema` because it's the most conservative/flexible approach.

src/protocol-parser/protocol-parser.ts
thiagowfx: Resolving. Accomplished with `.refine()` and `.transform()`.

PR #527, opened 2023-03-10
bump wpt

thiagowfx:


thiagowfx:
Since this is failing anyway, wait until Monday so that dependabot does the job.

PR #532, opened 2023-03-13
use source maps when running mocha unit tests

thiagowfx:
This improves debugging natively with our original TS files.

Previously, we would need to debug using the generated JS files.

References:

- https://jonnyreeves.co.uk/2015/getting-proper-stack-traces-with-mocha-and-typescript/
- https://futurestud.io/tutorials/mocha-migrating-from-mocha-opts-to-mocharc-js-configuration-file

### DIFF
--- a/.mocharc.json
+++ b/.mocharc.json
@@ -1,3 +1,5 @@
 {
+  "require": ["source-map-support/register"],
+  "source-maps": true,
   "spec": "lib/**/*.spec.js"
 }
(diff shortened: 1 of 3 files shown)


thiagowfx:
Status quo, JS:
image

Source maps, TS:
image

thiagowfx:
It's even better with VSCode debugging (instead of CLI):
image

sadym-chromium (approved):

PR #534, opened 2023-03-13
browsingContext.captureScreenshot: use fixed viewport in test

thiagowfx:
Bug: #514
Reference: https://chromedevtools.github.io/devtools-protocol/tot/Emulation/#method-setDeviceMetricsOverride


thiagowfx:
This seems to work for our local E2E test. We can now hard-code the base64 representation of the screenshot instead of matching against `ANY_STR`.

However, the upstream WPT test still fails.

OrKoN:
Screenshots are not cross-platform unfortunately. You need to have a screenshot golden per platform or exclude the screenshot content from the tests.

tests/test_screenshot.py
```diff
@@ -24,6 +24,30 @@
...
+
+    # Set a fixed viewport to make the test deterministic.
+    await execute_command(
+        websocket, {
+            "method": "cdp.sendCommand",
+            "params": {
+                "cdpMethod": "Emulation.setDeviceMetricsOverride",
+                "cdpParams": {
+                    "width": 1050,
+                    "height": 931,
```
sadym-chromium, line 43: It should work with smaller sizes, like 200x200

sadym-chromium: #537

sadym-chromium (approved):

PR #539, opened 2023-03-14
pre-commit: add keep-sorted

thiagowfx:


sadym-chromium:
Please provide motivation for the change

thiagowfx:
Context: This is globally and out-of-the-box available in google3 (c.f. go/keep-sorted). I am quite used to it, and it was recently (<1 month) open-sourced (you can see it's in the `google/` org. It's helpful in situations where you would say "please sort this list" in code reviews. You don't need to care about sorting, you just mark a designated block and then the presubmit takes care of automatically sorting it. This is intended to reduce future code review burden.

I had it locally for a while but didn't have a strong reason to submit it because there were only ~2 occurrences, but after Alex's recent browsers change I figured there was enough usage to justify adding it.

sadym-chromium (approved):

PR #540, opened 2023-03-14
`browsingContext.print`: fix a couple of unit tests

thiagowfx:
Bug: #518


thiagowfx:
I had a nice implementation with `superRefine`, but because it does not return the proper error type, it fails the test. I'll fall back to a simpler implementation instead.

sadym-chromium (approved):

PR #542, opened 2023-03-14
refactor script evaluator

thiagowfx:
In preparation for the introduction of the binding mechanism.

- 1) move public methods to the top, and private methods to the bottom of the class, for readability
- 2) move a few methods to within ScriptEvaluator, because we will need to use `this` to access the to be introduced event manager member variable

This is intended to reduce diff noise in the main PR (#319).

Furthermore, remove unnecessary switch case braces.

This is a no-op.

Bug: #294

### DIFF
--- a/src/bidiMapper/domains/script/scriptEvaluator.ts
+++ b/src/bidiMapper/domains/script/scriptEvaluator.ts
@@ -38,260 +38,6 @@ function cdpRemoteObjectToCallArgument(
   return {value: cdpRemoteObject.value};
 }
 
-async function deserializeToCdpArg(
-  argumentValue: Script.ArgumentValue,
-  realm: Realm
-): Promise<Protocol.Runtime.CallArgument> {
-  if ('sharedId' in argumentValue) {
-    const [navigableId, rawBackendNodeId] =
-      argumentValue.sharedId.split(SHARED_ID_DIVIDER);
-
-    const backendNodeId = parseInt(rawBackendNodeId ?? '');
-    if (
-      isNaN(backendNodeId) ||
-      backendNodeId === undefined ||
-      navigableId === undefined
-    ) {
-      throw new Message.InvalidArgumentException(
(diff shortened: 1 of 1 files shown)


src/bidiMapper/domains/script/scriptEvaluator.ts
```diff
@@ -319,6 +65,205 @@ export async function stringifyObject(
 }
 
 export class ScriptEvaluator {
+  async deserializeToCdpArg(
```
sadym-chromium, line 68: 1. `#deserializeToCdpArg`, as it is used only in the class.
2. Why not to move the other block as well?
```
const CALL_FUNCTION_STACKTRACE_LINE_OFFSET = 1;
const EVALUATE_STACKTRACE_LINE_OFFSET = 0;
export const SHARED_ID_DIVIDER = '_element_';
function cdpRemoteObjectToCallArgument(
  cdpRemoteObject: Protocol.Runtime.RemoteObject
): Protocol.Runtime.CallArgument {
  if (cdpRemoteObject.objectId !== undefined) {
    return {objectId: cdpRemoteObject.objectId};
  }
  if (cdpRemoteObject.unserializableValue !== undefined) {
    return {unserializableValue: cdpRemoteObject.unserializableValue};
  }
  return {value: cdpRemoteObject.value};
}
```

sadym-chromium:
Did you merge it on purpose or by mistake?

src/bidiMapper/domains/script/scriptEvaluator.ts
thiagowfx: Staged in https://github.com/GoogleChromeLabs/chromium-bidi/pull/319

thiagowfx: 2) Likewise, staged in https://github.com/GoogleChromeLabs/chromium-bidi/pull/319

sadym-chromium: Consider making it in a dedicated PR

PR #553, opened 2023-03-17
Enable `Network` domain globally

sadym-chromium:
When user subscribes to `network` events globally, all new browsing contexts should enable it.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -74,15 +74,24 @@ async def context_id(websocket):
 
 
 @pytest_asyncio.fixture
-async def another_context_id(websocket):
-    """Create a browsing context and return its id."""
-    result = await execute_command(websocket, {
-        "method": "browsingContext.create",
-        "params": {
-            "type": "tab"
-        }
-    })
-    return result['context']
+async def another_context_id(create_context):
+    return await create_context()
+
+
+@pytest_asyncio.fixture
+def create_context(websocket):
+
+    async def _(createType="tab"):
+        """Create a browsing context and return its id."""
+        result = await execute_command(websocket, {
+            "method": "browsingContext.create",
+            "params": {
+                "type": createType
+            }
+        })
+        return result['context']
+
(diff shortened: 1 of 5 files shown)


thiagowfx (approved):

tests/conftest.py
```diff
@@ -74,15 +74,24 @@ async def context_id(websocket):
...
-    return result['context']
+async def another_context_id(create_context):
+    return await create_context()
+
+
+@pytest_asyncio.fixture
+def create_context(websocket):
+
+    async def _(createType="tab"):
+        """Create a browsing context and return its id."""
```
thiagowfx, line 85: Ideally, this should be documented at the fixture level (i.e. outermost `def`), so that it shows up in `pytest --fixtures`

tests/conftest.py
```diff
@@ -74,15 +74,24 @@ async def context_id(websocket):
...
-    })
-    return result['context']
+async def another_context_id(create_context):
+    return await create_context()
+
+
+@pytest_asyncio.fixture
+def create_context(websocket):
+
+    async def _(createType="tab"):
```
thiagowfx, line 84: Is it possible to pass another `createType` from the caller? I haven't seen an example of that yet.

Otherwise we may as well hard-code the type.

tests/test_helpers.py
```diff
@@ -135,3 +135,19 @@ async def wait_for_event(websocket, event_method):
         if "method" in event_response and event_response[
                 "method"] == event_method:
             return event_response
+
+
+def AnyExtending(expected):
```
thiagowfx, line 140: Would be nice to document this.

Also, this seems to re-create `recursive_compare`. Did you check if the diffs lock nicer when the test fails? Otherwise we may just have walked in a circle

sadym-chromium: https://github.com/GoogleChromeLabs/chromium-bidi/pull/560

PR #554, opened 2023-03-17
Add CDP events debug output

sadym-chromium:


OrKoN (approved):

src/bidiServer/mapperServer.ts
```diff
@@ -64,9 +68,11 @@ export class MapperServer {
   setOnMessage(handler: (message: string) => void): void {
     this.#handlers.push(handler);
   }
+
   sendMessage(messageJson: string): Promise<void> {
-    return this.#sendBidiMessage(messageJson);
+    return MapperServer.#sendBidiMessage(messageJson, this.#mapperCdpClient);
```
thiagowfx, line 73: Good change.
We should have a linter check to catch `this.#` with static methods. For another PR.

PR #555, opened 2023-03-17
Make `verbose` option to log all the CDP traffic in Mapper Tab

sadym-chromium:


src/bidiTab/bidiTab.ts
```diff
@@ -52,21 +53,28 @@ declare global {
 
     // `window.setSelfTargetId` is called via `Runtime.evaluate` from the server side.
     setSelfTargetId: (targetId: string) => void;
+
+    // `window.verboseDebug` is set via `Runtime.evaluate` in case of verbose debugging mode.
+    verboseDebugChannel: string | undefined;
```
thiagowfx, line 58: https://google.github.io/styleguide/tsguide.html#optionals-vs-undefined-type:

> Use optional fields (on interfaces or classes) and parameters rather than a |undefined type.

src/bidiTab/logger.ts
```diff
@@ -0,0 +1,63 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+import {LogType} from '../utils/log';
+import {BidiServer} from '../bidiMapper/BidiServer';
+import {OutgoingBidiMessage} from '../bidiMapper/OutgoingBidiMessage';
+
+import {log as logToPage} from './mapperTabPage.js';
```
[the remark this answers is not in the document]
OrKoN: it should always have .js or it will not run in the browser.

sadym-chromium:
I re-worked the solution and switched to a dedicated binding for debugging, which is more logical.

src/bidiServer/mapperServer.ts
```diff
@@ -19,12 +19,14 @@ import Protocol from 'devtools-protocol';
 import WebSocket from 'ws';
 import debug from 'debug';
 
-import {CdpClient, CdpConnection, WebSocketTransport} from '../cdp/index.js';
+import {CdpClient, CdpConnection, WebSocketTransport} from '../cdp';
```
OrKoN, line 22: let's use '.js' consistently and let's not import using node's resolution to be able to potentially re-use the code on the browser side.

src/bidiServer/mapperServer.ts
```diff
@@ -95,31 +102,52 @@ export class MapperServer {
...
+    if (params.name === 'sendDebugMessage') {
+      this.#onDebugMessage(params.payload);
+    }
+  };
+
+  #onDebugMessage = (debugMessageStr: string) => {
+    try {
+      const debugMessage = JSON.parse(debugMessageStr) as {
+        logType: string;
+        messages: any[];
```
thiagowfx, line 128: unless the compilation fails, prefer `unknown[]` consistently for log messages

https://google.github.io/styleguide/tsguide.html#any

src/bidiTab/mapperTabPage.ts
```diff
@@ -70,6 +70,11 @@ export function log(logType: LogType, ...messages: unknown[]) {
     return;
   }
 
+  // If `sendDebugMessage` is defined, send the log message there.
+  (globalThis as unknown as Window).sendDebugMessage?.(
```
thiagowfx, line 74: does `as global.Window` work by any chance? To leverage the definitions from `bidiTab.ts`

src/bidiServer/mapperServer.ts
sadym-chromium: done

src/bidiServer/mapperServer.ts
sadym-chromium: done

src/bidiTab/mapperTabPage.ts
sadym-chromium: done

thiagowfx (approved):

src/bidiServer/mapperServer.ts
thiagowfx: Should this be applied to the entire codebase? I.e. always import with `.js` suffix?

PR #565, opened 2023-03-22
pre-refactor script evaluator for #319

thiagowfx:
Bug: #294

### DIFF
--- a/tests/test_print.py
+++ b/tests/test_print.py
@@ -19,7 +19,7 @@ from test_helpers import goto_url, read_JSON_message, send_JSON_command
 
 
 @pytest.mark.asyncio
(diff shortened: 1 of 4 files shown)


thiagowfx:
@sadym-chromium as we agreed yesterday, I am going ahead with this without review, to expedite the review of #319 which is the important/non-trivial change.

tests/test_print.py
```diff
@@ -19,7 +19,7 @@
 
 
 @pytest.mark.asyncio
-async def test_print_happy_path(websocket, context_id):
+async def test_print(websocket, context_id):
```
thiagowfx, line 22: This is just a small drive-by

PR #566, opened 2023-03-22
make type of eventName stronger, from string to SubscribeParametersEvent

thiagowfx:


sadym-chromium:
This is not 100% correct, as allows sending domain-level event like `browsingContext`, which can be used only for subscription / un-subscription for all the domain events. If it is really wanted, please export `EventNames = BrowsingContext.EventNames | Log.EventNames | ...` and reuse it in subscription: `SubscribeParametersEvent = EventNames | typeof BrowsingContext.AllEvents | ...`.

src/bidiMapper/domains/events/EventManager.ts
```diff
@@ -265,7 +268,7 @@ export class EventManager implements IEventManager {
   #markEventSent(
     eventWrapper: EventWrapper,
     channel: string | null,
-    eventName: string
+    eventName: Session.SubscribeParametersEvent
```
sadym-chromium, line 271: And I believe here it should be the specific event as well
```suggestion
    eventName: Message.EventNames
```

thiagowfx:
The buffer only works if I add casts, e.g.

```
this.#bufferEvent(eventWrapper, eventName as Message.EventNames);
```

Is this OK?

sadym-chromium:
`registerPromiseEvent`'s argument  should be `Message.EventMessage`, not `Session.SubscribeParametersEvent`

PR #568, opened 2023-03-24
eslint: add parser services, for advanced introspection

thiagowfx:
Caveat: Typed rules come with a catch. By including parserOptions.project in our config, we incur the performance penalty of asking TypeScript to do a build of your project before ESLint can do its linting. For small projects this takes a negligible amount of time (a few seconds or less); for large projects, it can take longer.

Docs:
- https://typescript-eslint.io/linting/typed-linting/
- https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/recommended-requiring-type-checking.ts


thiagowfx:
This will be a draft for a while, do not review yet. I am iterating on it in-between feature PRs.

sadym-chromium (approved):
Is unnecessary `await` causes warning or an error? If the latest, I'd avoid such a change. Otherwise LGTM

OrKoN:
Leaving review to Maksim

thiagowfx:
> Is unnecessary await causes warning or an error? If the latest, I'd avoid such a change. Otherwise LGTM

I'll downgrade it to a warning instead, then.

thiagowfx:
@sadym-chromium I downgraded a bunch of rules to `'warn'` so that they are less intrusive during local iteration, but still caught by pre-commit.

thiagowfx:
> Is unnecessary await causes warning or an error? If the latest, I'd avoid such a change. Otherwise LGTM

That would be this rule: https://eslint.org/docs/latest/rules/require-await

It was out of the config (implicitly `'error'` due to `'plugin:@typescript-eslint/recommended-requiring-type-checking',`), I added it explicitly and set it to `'warn'` as well.

This is now done.

PR #572, opened 2023-03-24
Augment error codes to fully match the spec

thiagowfx:
Spec: https://w3c.github.io/webdriver-bidi/#errors


src/protocol/protocol.ts
```diff
@@ -90,11 +90,18 @@ export namespace Message {
   // keep-sorted end;
 
   export type ErrorCode =
-    | 'unknown error'
-    | 'unknown command'
+    // keep-sorted start
     | 'invalid argument'
```
sadym-chromium, line 94: If move them to enum like event names, we wouldn't need unit tests for them.

sadym-chromium (approved):

src/protocol/protocol.ts
sadym-chromium: > Should we still keep one custom class per error code?

I guess it still makes sense to have a class per exception. Otherwise you have to throw `Response`, which is odd, and it becase less readable.

thiagowfx: Done

PR #581, opened 2023-03-28
examples: use counters for IDs

thiagowfx:


examples/_helpers.py
```diff
@@ -13,12 +13,15 @@
...
 # limitations under the License.
 
+import itertools
 import json
 import os
 
 import requests
 import websockets
 
+ID = itertools.count(1000)
```
sadym-chromium, line 23: I want to keep the examples as simple as possible, and I'm, not sure if this simplifies the reading, especially having magical `1000`

thiagowfx: `1000` was already there. `itertools` is a popular module, it doesn't have much magic. Messing up the counters is error-prone, you can even think of this PR as a bug fix, as it is making all IDs be sequential. Previously there were some examples with 0 and then skipping right to 1000.

sadym-chromium: ok

PR #591, opened 2023-04-02
Remove "open-pull-requests-limit" from dependabot

thiagowfx:
With a cadence of monthly updates, restricting the PR limits to 1 proves to be too restrictive. The default is 5.

Docs: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#open-pull-requests-limit

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,19 +9,15 @@ updates:
     directory: /
     schedule:
       interval: monthly
-    open-pull-requests-limit: 1
   - package-ecosystem: pip
     directory: examples/
     schedule:
       interval: monthly
-    open-pull-requests-limit: 1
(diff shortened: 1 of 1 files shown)


thiagowfx:
Although, one thing that would be nicer is if dependabot could consolidate all of these PRs into a single one:

- https://www.hrvey.com/blog/combine-dependabot-prs
- https://marcreichel.dev/blog/combine-dependabot-pull-requests

Merging a couple of PRs is tedious. Then, it's a fine balance between not worrying about updates (as dependabot handles everything for us) and rather dealing with all of them at once every month, versus manually updating everything every now and then.

We could look into this some other time. For now, I am happy to do the tedious merges to bump versions.

PR #603, opened 2023-04-03
Add wireit and integrate it with build/test/lint lifecycle

thiagowfx:
I am still evaluating whether this PR brings any value (given the cost of the extra dev dependency), mostly for the `watch` workflow. For now, consider it a draft, do not merge it.

This is mostly inspired by Puppeteer (https://github.com/puppeteer/puppeteer/commit/78a302174e040311f7b3a4c047c5a9bee45929f5)


package.json
```diff
@@ -3,7 +3,8 @@
   "version": "0.4.6",
   "description": "An implementation of the WebDriver BiDi protocol for Chromium implemented as a JavaScript layer translating between BiDi and CDP, running inside a Chrome tab.",
   "scripts": {
-    "build": "tsc -b src/tsconfig.json && npm run rollup",
+    "build": "wireit",
+    "build:watch": "npm run tsc:watch && npm run rollup:watch",
```
Lightning00Blade, line 7: Not sure that this will work as `&&` waits for the first command to finish and exit with code `0` to execute the second one.

package.json
```diff
@@ -74,6 +114,7 @@
     "tslib": "2.5.0",
     "typescript": "4.7.4",
     "websocket": "1.0.34",
+    "wireit": "^0.9.5",
```
Lightning00Blade, line 117: Usually dependencies need to be pinned.
```suggestion
    "wireit": "0.9.5",
```

package.json
```diff
@@ -16,13 +17,52 @@
...
     "prettier": "prettier --write .",
-    "rollup": "rollup -c",
+    "rollup": "wireit",
+    "rollup:watch": "rollup -c --watch",
     "server": "npm run build && npm run server-no-build --",
     "server-no-build": "node lib/cjs/bidiServer/index.js",
-    "test": "npm run build && mocha",
-    "watch": "tsc -b src/tsconfig.json --watch && rollup -c --watch",
+    "test": "wireit",
+    "test:watch": "npm run build && mocha --watch",
```
Lightning00Blade, line 25: This can be a wireit command as well dependency of build and command `mocha --watch`

package.json
```diff
@@ -16,13 +17,52 @@
     "pre-commit": "pre-commit run --all-files",
     "prepare": "npm run clean && npm run build",
     "prettier": "prettier --write .",
-    "rollup": "rollup -c",
+    "rollup": "wireit",
```
Lightning00Blade, line 20: You don't need to define every `wireit` command here (examples may be outdated on the `wireit` part as that is newish [CHANGELOG](https://github.com/google/wireit/blob/main/CHANGELOG.md#changed-1)), I think it would be better to only have a single `build` command.

package.json
thiagowfx: When we use `&`, the output is garbled. Removed in favor of `wireit`'s `--watch`.

package.json
thiagowfx: Ack

package.json
thiagowfx: Done

thiagowfx:
This is now complete.

Lightning00Blade (approved):

package.json
```diff
@@ -3,26 +3,73 @@
   "version": "0.4.7",
   "description": "An implementation of the WebDriver BiDi protocol for Chromium implemented as a JavaScript layer translating between BiDi and CDP, running inside a Chrome tab.",
   "scripts": {
-    "build": "tsc -b src/tsconfig.json && npm run rollup",
+    "build": "wireit",
     "clean": "rimraf lib",
     "coverage": "nyc mocha",
-    "e2e-headful": "npm run server-no-build -- --headless=false & npm run e2e-only",
-    "e2e-headless": "npm run server-no-build & npm run e2e-only",
+    "e2e-headful": "npm start -- --headless=false & npm run e2e-only",
```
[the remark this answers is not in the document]
thiagowfx: I tried to convert the `start` block into a "service", but it is making the CI fail. I will investigate deeper in a follow-up.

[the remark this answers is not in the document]
thiagowfx: Fixed.

sadym-chromium:
This PR broke verbose logging

Issue #605, opened 2023-04-06
Race condition when processing messages in ProcessingQueue


Lightning00Blade:
In some instances it's possible create a deadlock the processing queue. 
This happens in the following puppeteer test [`should fail when navigating to bad SSL`](https://github.com/puppeteer/puppeteer/blob/main/test/src/navigation.spec.ts#L161)
The navigation command is send with `id` the error is correctly handle and pushed to ProcessingQueue
but it never create an outbound response to the Transport.

Adding a hack to ProcessingQueue:46 to have
```
if (this.#isProcessing && this.#queue.length !== 1) {
``` 
unblocks the process but better solution should be created.

sadym-chromium:
It's not a race condition. The issue is in the fact we don't process `Network.loadingFailed` event.

PR #607, opened 2023-04-11
Process network errors

sadym-chromium:
Initial implementation of the `Network.fetchError`. Not fully compliant with the spec, but pushing anyways to avoid hanging event queue eg. like in #605.


tests/test_network.py
```diff
@@ -241,3 +241,42 @@ async def test_network_network_response_completed_event_emitted(
             }
         }
     }
+
+
+@pytest.mark.asyncio
+async def test_network_bad_ssl(websocket, context_id):
+    BAD_SSL_URL = "https://expired.badssl.com/"
```
thiagowfx, line 248: We have a bunch of TODOs to make the tests work offline. Does it make sense to add one here as well?

tests/test_helpers.py
```diff
@@ -12,6 +12,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from __future__ import annotations
```
thiagowfx, line 15: Why is this needed?

thiagowfx (approved):

tests/test_helpers.py
sadym-chromium: In my setup it is needed to use types:
```
test_helpers.py:138: in <module>
    def AnyExtending(expected: list | dict):
E   TypeError: unsupported operand type(s) for |: 'type' and 'type'
```

tests/test_network.py
sadym-chromium: added

thiagowfx (approved):

Lightning00Blade:
FYI this produces an unhandled error atm:
```
Uncaught Error: the object {
  "canceled": false
  "errorText": "net::ERR_CERT_AUTHORITY_INVALID"
  "requestId": "6734277685B2C23C5D68E738D948F08A"
  "timestamp": 473342.498087
  "type": "Document"
} was thrown, throw an Error :)
```
Please revisit when you do further implementations here.

Issue #293, opened 2022-10-13
Implement `script.addPreloadScript` and `script.removePreloadScript`


sadym-chromium:
Implement suggestion in https://github.com/w3c/webdriver-bidi/pull/292, which can be directly mapped to [`Page.addScriptToEvaluateOnNewDocument`](https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-addScriptToEvaluateOnNewDocument).

e2e tests are included in the task.

**Out of scope**: channels

## Tracking

- [x] `crypto` document: [go/webdriver:bidi-import-crypto-module](http://go/webdriver:bidi-import-crypto-module)
- [x] `crypto` implementation not depending on external module
- [x] mock/stub unit tests for preload script storage
- [x] add unit tests
- [x] fix last round of WPT tests
- [x] preload scripts should load in new contexts/tabs (`test_addPreloadScriptGlobally_loadedInNewContexts` E2E test should pass): https://github.com/GoogleChromeLabs/chromium-bidi/pull/673
- [x] make changes to CDP target w.r.t. unblocking order
- [x] OOPIF edge cases
- [x] ensure to account for parent context when `context` is provided in `add preload script`
- [x] ~~deduplicate CDPsession IDs when adding new preload scripts with a given `context` (they should be unique)~~
- [x] make WPT tests pass
- [x] make all E2E tests pass (`test_preload_script.py`)
- [x] ~~Add E2E test to expose race condition between [`CdpTarget.create` -> `void cdpTarget.#unblock()`](https://github.com/GoogleChromeLabs/chromium-bidi/blob/43882f46dca6e203b0bdbb12c4aa9d79a6a3895b/src/bidiMapper/domains/context/cdpTarget.ts#L55) and adding preload script: follow-up of https://github.com/GoogleChromeLabs/chromium-bidi/pull/673#discussion_r1184848960~~: superseded by item below 
  - [x] Upstream: [Issue 1449849: `Page.addScriptToEvaluateOnNewDocument` added before `Runtime.runIfWaitingForDebugger` should be run](https://bugs.chromium.org/p/chromium/issues/detail?id=1449849)
  - [x] Apply it in the codebase (`runImmediately`): https://github.com/GoogleChromeLabs/chromium-bidi/pull/919
- [x] support arguments / channels
- [x] support sandbox: https://github.com/GoogleChromeLabs/chromium-bidi/pull/978
[...]

PR #608, opened 2023-04-11
Add E2E test for add preload script for new contexts/frames

thiagowfx:
Bug: #293

Script:

- [x] Add Map ( from browsing context id to preload script) to browsing context processor.
- [x] Global: map null.
- [x] Verify browsing context id exists.
- [x] Use Map

Tests to add:

- [x] add preload script to a context that does not exist
- [x] add more than one script (different scripts) to the same context
- [x] add exactly the same script to the same context
- [x] @sadym-chromium: OOPIF (`test_addPreloadScript_loadedInNewIframes_withCrossOrigin`)

 >I would add additional test when same-origin frame became OOPiF, to cover [this logic](https://github.com/GoogleChromeLabs/chromium-bidi/blob/618b4d0c09b2ecc47bf1278a91bc931781a8a44b/src/bidiMapper/domains/context/browsingContextProcessor.ts#L178).


tests/test_preload_script.py
```diff
@@ -95,3 +95,146 @@ async def test_addPreloadScript_logging(websocket, context_id):
     # Assert navigation is finished.
     result = await read_JSON_message(websocket)
     assert result == {"id": command_id, "result": ANY_DICT}
+
+
+@pytest.mark.asyncio
+async def test_addPreloadScript_loadedInMultipleContexts(
```
thiagowfx, line 101: I am not sure if there's much value in this test. Let me know if I should remove it.

tests/test_preload_script.py
```diff
@@ -95,3 +95,143 @@ async def test_addPreloadScript_logging(websocket, context_id):
...
```
sadym-chromium, line 101:
```suggestion
async def test_addPreloadScriptGlobally_loadedInMultipleContexts(
```

tests/test_preload_script.py
```diff
@@ -95,3 +95,143 @@ async def test_addPreloadScript_logging(websocket, context_id):
...
+    result = await execute_command(
+        websocket, {
+            "method": "script.addPreloadScript",
+            "params": {
+                "expression": "() => { window.count=(window.count ?? 0) + 1 }",
+                "context": another_context_id,
+            }
+        })
+    assert result == {'script': ANY_STR}
+
```
sadym-chromium, line 145: It doesn't make sense to add another script, as the original one is global now.
```suggestion

```

tests/test_preload_script.py
```diff
@@ -95,3 +95,143 @@ async def test_addPreloadScript_logging(websocket, context_id):
...
+    result = await execute_command(
+        websocket, {
+            "method": "script.addPreloadScript",
+            "params": {
+                "expression": "() => console.log('my preload script')",
+                "context": context_id,
+            }
+        })
+    assert result == {'script': ANY_STR}
```
sadym-chromium, line 183:
```suggestion
    await execute_command(
        websocket, {
            "method": "script.addPreloadScript",
            "params": {
                "expression": "() => console.log('my preload script')",
                "context": context_id,
            }
        })
```

tests/test_preload_script.py
```diff
@@ -95,3 +95,143 @@ async def test_addPreloadScript_logging(websocket, context_id):
...
+    result = await read_JSON_message(websocket)
+    assert result == {"id": command_id, "result": ANY_DICT}
+
+    # Create a new iframe within the same context.
+    result = await execute_command(
+        websocket, {
+            "method": "script.evaluate",
+            "params": {
+                "expression": """const iframe = document.createElement('iframe');
+                    iframe.src = 'https://example.com';
```
sadym-chromium, line 217: `data:html..... console.log('I'm the page')`.

thiagowfx:
- [ ] Add Map of preload scripts to browsing context processor.
- [ ] Map: from browsing context id to preload script.
- [ ] Verify browsing context id exists.
- [ ] When context is destroyed, remove preload script.
- [ ] Global: map `null`.

thiagowfx:
My understanding:

- `about:blank` => same-origin
- https://example.com, `data:text/html,<h2>foo</h2>` => cross-origin (OOPIF)

thiagowfx:
I believe all intersections are being tested now. Please let me know if you agree.

thiagowfx:
Sent https://github.com/GoogleChromeLabs/chromium-bidi/pull/612 to help with syntactic sugar for fixtures.

thiagowfx:
Q: Do I need to retrieve the context id of the created iframes?

tests/test_preload_script.py
```diff
@@ -95,3 +95,290 @@ async def test_addPreloadScript_logging(websocket, context_id):
...
+            }]
+        }
+    })
+
+    # Assert navigation is finished.
+    result = await read_JSON_message(websocket)
+    assert result == {"id": command_id, "result": ANY_DICT}
+
+    # Create a new iframe within the same context.
+    result = await execute_command(
```
sadym-chromium, line 146: Consider using `await send_JSON_command` / `await read_JSON_message` here to avoid timeout in case of preload script message was not sent. Same in other places. It's way better when test fails fast.

tests/test_preload_script.py
```diff
@@ -95,3 +95,290 @@ async def test_addPreloadScript_logging(websocket, context_id):
...
+            }
+        })
+
+    result = await execute_command(
+        websocket, {
+            "method": "script.evaluate",
+            "params": {
+                "expression": "window.foo",
+                "target": {
+                    "context": context_id
```
sadym-chromium, line 378: You check the original page has `window.foo`, not the newly created iframe.

tests/test_preload_script.py
```diff
@@ -95,3 +95,290 @@ async def test_addPreloadScript_logging(websocket, context_id):
...
+async def test_addPreloadScript_loadedInNewIframes(websocket, context_id,
+                                                   nested_iframe):
+    await subscribe(websocket, "log.entryAdded")
+
+    await execute_command(
+        websocket, {
+            "method": "script.addPreloadScript",
+            "params": {
+                "expression": "() => console.log('my preload script')",
+                "context": context_id,
```
sadym-chromium, line 115: Consider parametrize `globally` / `for a given context` as well.
```suggestion
@pytest.mark.parametrize("nested_iframe", [
    'about:blank',
    'https://example.com/',
])
@pytest.mark.parametrize("specific_context", [True, False])
@pytest.mark.asyncio
async def test_addPreloadScript_loadedInNewIframes(websocket, context_id,
                                                   nested_iframe):
    await subscribe(websocket, "log.entryAdded")

    await execute_command(
        websocket, {
            "method": "script.addPreloadScript",
            "params": (specific_context ? {"context": context_id} : {}) | {
                "expression": "() => console.log('my preload script')",
```

tests/test_preload_script.py
```diff
@@ -95,3 +95,290 @@ async def test_addPreloadScript_logging(websocket, context_id):
...
+                    "context": context_id
+                },
+                "awaitPromise": True,
+                "resultOwnership": "root"
+            }
+        })
+
+    # Asset that the preload script is executed in the new iframe.
+    result = await read_JSON_message(websocket)
+    assert result == AnyExtending({
```
sadym-chromium, line 163: I'd check the event's `contextId` is not equal to the original `context_id` to make sure the event came not from the original context. Checking very specific context_id as mentioned in [the comment](https://github.com/GoogleChromeLabs/chromium-bidi/pull/608/files#r1165178275) would be event better.

tests/test_preload_script.py
thiagowfx: Done.

thiagowfx:
All suggestions resolved. All tests pass except the last one, `test_addPreloadScriptGlobally_loadedInMultipleContexts_withIframes`, due to:

```
Failed: [undefined]Exception: {'error': 'unknown error', 'message': 'No default realm for browsing context 
```

thiagowfx:
We need a "defaultdict" (similar to python) data structure in TS to simplify the map to set logic.

thiagowfx:
> When context is created, add global scripts if any

Where should this live? Within ` this.#cdpTarget.cdpClient.on('Runtime.executionContextCreated',` ?

tests/test_preload_script.py
```diff
@@ -95,3 +95,332 @@ async def test_addPreloadScript_logging(websocket, context_id, html):
...
+    # Assert navigation is finished.
+    result = await read_JSON_message(websocket)
+    assert result == {"id": command_id, "result": ANY_DICT}
+
+    # Create a new iframe within the same context.
+    result = await execute_command(
+        websocket, {
+            "method": "script.evaluate",
+            "params": {
+                "expression": f"""const iframe = document.createElement('iframe');
```
sadym-chromium, line 145: I would add additional test when same-origin frame became OOPiF, to cover [this logic](https://github.com/GoogleChromeLabs/chromium-bidi/blob/618b4d0c09b2ecc47bf1278a91bc931781a8a44b/src/bidiMapper/domains/context/browsingContextProcessor.ts#L178).

thiagowfx: Resolving this comment and keeping track of it within the first comment of this PR as a task

thiagowfx:
This PR has been reverted to add tests only. I'll send a separate PR for the implementation.

tests/test_preload_script.py
```diff
@@ -74,7 +95,280 @@ async def test_addPreloadScript_logging(websocket, context_id, html):
...
+            "args": [{
+                "type": "string",
+                "value": "my preload script"
+            }]
+        }
+    })
+
+    # Assert navigation is finished.
+    result = await read_JSON_message(websocket)
+    assert result == {"id": command_id, "result": ANY_DICT}
```
sadym-chromium, line 118: NIT
```suggestion
    assert result == AnyExtending({"id": command_id})
```

tests/test_preload_script.py
```diff
@@ -74,7 +95,280 @@ async def test_addPreloadScript_logging(websocket, context_id, html):
...
+            "args": [{
+                "type": "string",
+                "value": "my preload script"
+            }]
+        }
+    })
+
+
+@pytest.mark.asyncio
+async def test_addPreloadScript_loadedInNewIframes_withSameOrigin(
```
sadym-chromium, line 283: The only difference with the previous test is the url, so consider parametrising the test.

tests/test_preload_script.py
sadym-chromium: ?

tests/test_preload_script.py
```diff
@@ -74,7 +95,280 @@ async def test_addPreloadScript_logging(websocket, context_id, html):
...
+                "value": "my preload script"
+            }]
+        }
+    })
+
+    result = await read_JSON_message(websocket)
+
+
+@pytest.mark.asyncio
+async def test_addPreloadScript_loadedInNewIframes_withChildScript(
```
sadym-chromium, line 354: Is it any different from the previous 2 tests?

tests/test_preload_script.py
```diff
@@ -95,3 +389,152 @@ async def test_addPreloadScript_logging(websocket, context_id, html):
...
+                "type": "string",
+                "value": "I am the child"
+            }]
+        }
+    })
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("subscribe_to_context", [True, False])
+async def test_addPreloadScriptGlobally_loadedInMultipleContexts(
```
sadym-chromium, line 436: test name and parameter are controversial

tests/test_preload_script.py
thiagowfx: The `iframe` with `console.log` is the differential. Do you want me to delete it? Deleting is easier than writing ;)

tests/test_preload_script.py
thiagowfx: Do you have a specific suggestion?

tests/test_preload_script.py
thiagowfx: You're absolutely right, the main intent of the split was to make it explicit what is same origin and what is cross origin, for readability. That said, merging them is trivial. Done.

tests/test_preload_script.py
sadym-chromium: Oh, I see. Then it should be called smth like `test__addPreloadScript_runBeforePageScript`.

tests/test_preload_script.py
sadym-chromium: e.g. you can merge it with the previous test and rely on the events or both on event and global property

sadym-chromium (approved):

tests/test_preload_script.py
thiagowfx: Removed `Globally` from method name

PR #611, opened 2023-04-11
Make deferred don't  raise `Uncaught (in promise)`

### LINKED ISSUE #528: Fix no-op `.finally()` call in ProcessingQueue
https://github.com/GoogleChromeLabs/chromium-bidi/blob/adc59a3030152b48599ee669533b137508a93d4f/src/utils/processingQueue.ts#L59

        await entryPromise
          .then((entry) => this.#processor(entry))
          .catch((e) => {
            this.#logger?.(LogType.system, 'Event was not processed:', e);
            this.#catch(e);
          })
          .finally();
      }

There should be another `.catch()` there, in lieu of the `.finally()`, which is a no-op that does not deal with the underlying promise. This has been confirmed with other folks in our internal TS Readability chat.

sadym-chromium:
Closes #528

I didn't find a way to add an auto test for it, but I was able to reproduce the issue manually, and it's gone after the fix.

The following exception is gone:
```
  bidiMapper:log exceptionThrown {
  timestamp: 1681290303299.3938,
  exceptionDetails: {
    exceptionId: 1,
    text: 'Uncaught (in promise)',
    lineNumber: 0,
    columnNumber: 73033,
    scriptId: '6',
    stackTrace: { callFrames: [Array] },
    exception: {
      type: 'object',
      className: 'Object',
      description: 'Object',
      objectId: '-111297914861493100.1.1',
      preview: [Object]
    },
    executionContextId: 1
  }
} +0ms
```

### DIFF
--- a/src/utils/deferred.ts
+++ b/src/utils/deferred.ts
@@ -31,6 +31,9 @@ export class Deferred<T> implements Promise<T> {
       this.#resolve = resolve;
       this.#reject = reject;
     });
+    // Needed to avoid `Uncaught (in promise)`. The promises returned by `then`
+    // and `catch` will be rejected anyway.
+    this.#promise.catch(() => {});
   }
 
   then<TResult1 = T, TResult2 = never>(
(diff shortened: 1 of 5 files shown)


thiagowfx (approved):

thiagowfx:
Closes #528

src/utils/deferred.ts
```diff
@@ -31,6 +31,9 @@ export class Deferred<T> implements Promise<T> {
       this.#resolve = resolve;
       this.#reject = reject;
     });
+    // Needed to avoid `Uncaught (in promise)`. The promises returned by `then`
```
thiagowfx, line 34: Does it make a sense to add an unit test for this?

sadym-chromium: good point. Done.

PR #612, opened 2023-04-11
pytest: misc fixture additions: html, iframe, parameterized fixtures

thiagowfx:
Docs: https://docs.pytest.org/en/latest/how-to/fixtures.html#fixture-parametrize

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -15,6 +15,7 @@
 
 import os
 
+import pytest
 import pytest_asyncio
 import websockets
 from test_helpers import execute_command, get_tree, goto_url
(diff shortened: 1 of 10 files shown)


tests/conftest.py
```diff
@@ -30,7 +31,36 @@ async def websocket():
 
 
 @pytest_asyncio.fixture
```
thiagowfx, line 33: Just moving context a bit earlier in the file to make the order a bit more logical.

tests/conftest.py
```diff
@@ -66,45 +96,64 @@ async def sandbox_realm(context_id, websocket):
...
-        })
-        return result['context']
 
-    return _
+@pytest.fixture
+def html():
+    """Return a factory for HTML data URL with the given content."""
 
+    def html(content=""):
+        return f'data:text/html,{content}'
```
sadym-chromium, line 131: url encode?

thiagowfx: This will break client code. There's some downstream code that is quoting manually. Can we do that in a follow-up? For now, I'd like this to be a no-op transformation where possible.

PR #617, opened 2023-04-14
add browsing context storage docs and unit tests

thiagowfx:
- Add jsdoc
- Add basic unit tests
- Rename methods consistently
- Rename remove -> delete to standardize terminology


thiagowfx:
It's quite hard to add more unit tests at the moment because the classes are quite coupled with each other, i.e. it's hard to define a browser context storage instance because it requires us to instantiate pretty much every other class in the codebase. In the future we should try to decouple the classes from each other with builder pattern and dependency injection.

PR #618, opened 2023-04-16
eslint: add consistent-generic-constructors rule

thiagowfx:
Currently our codebase is adopting a mixed style. Make it uniform to increase readability.


thiagowfx:
Going ahead with this as it's non-controversial and so that I can send another ESLint PR that deserves review.

Context: This PR stems from https://github.com/GoogleChromeLabs/chromium-bidi/pull/608, because I added a map there and thought we could benefit from some consistency.

PR #619, opened 2023-04-16
eslint: add a couple of rules from strict config

thiagowfx:
- array-type
- no-unsafe-enum-comparison
- prefer-return-this-type

Reference: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/strict.ts


.eslintrc.js
```diff
@@ -56,8 +56,10 @@ module.exports = {
     // https://denar90.github.io/eslint.github.io/docs/rules/
     // Some rules use 'warn' in order to ease local development iteration.
     // keep-sorted start
-    "@typescript-eslint/explicit-member-accessibility": ["warn", {accessibility: 'no-public'}],
+    "@typescript-eslint/explicit-member-accessibility": ['warn', {accessibility: 'no-public'}],
+    '@typescript-eslint/array-type': 'warn',
     '@typescript-eslint/consistent-generic-constructors': 'warn',
+    '@typescript-eslint/consistent-type-definitions': ['warn', 'type'],
```
thiagowfx, line 62: Both "type" and "interface" work. I picked "type" because we use it way more than "interface", especially in the protocol file.

sadym-chromium: I guess it makes sense to keep distinguishing between `type` and `interface`.

thiagowfx: OK, reverted this part.

Issue #620, opened 2023-04-17
No `log.entryAdded` after navigation


Lightning00Blade:
If we navigate to a page that log things on load (`ReadinessState == 'complete'`). 
An repo Puppeteer Script be seen at [#10032](https://github.com/puppeteer/puppeteer/issues/10032). 

After an investigation I was able to deduce that the `RealmStorage` is being cleared by the CDP `Page.frameNavigated` event.
And the `Runtime.executionContextCreated` event creates a new Realm.
The first event handler is Async while the second is Sync. I think we are creating a `new Realm` then the first is awaited and we delete all the old Realms for the BrowsingContext included the newly created on.

The line that is doing [this](https://github.com/GoogleChromeLabs/chromium-bidi/blob/main/src/bidiMapper/domains/context/browsingContextImpl.ts#L282).

If I reorder the ` this.#realmStorage.deleteRealms` before `await this.#deleteChildren();` things work as expected.
Please check if there is a specific reason for the current order of operation.

thiagowfx:
I wonder if we can replace

```
await this.#deleteChildren();

// Remove all the already created realms.
this.#realmStorage.deleteRealms({browsingContextId: this.contextId});
```

with

```
this.delete();
```

?

thiagowfx:
I cannot reproduce it in BiDi. The following test passes:

```
@pytest.mark.asyncio
async def test_browsingContext_logAfterNavigation(websocket, context_id, html):
    await subscribe(websocket, "log.entryAdded")

    await send_JSON_command(
        websocket, {
            "method": "browsingContext.navigate",
            "params": {
                "url": html("<script>console.log('mylog');</script>"),
                "context": context_id,
                "wait": "complete"
            }
        })

    result = await wait_for_event(websocket, "log.entryAdded")
    assert result == {
        "method": "log.entryAdded",
        "params": AnyExtending({
            "text": "mylog",
        })
    }
```

thiagowfx:
@sadym-chromium let me know if you manage to create a repro.

Lightning00Blade:
I can confirm that the suggestion in [comment](https://github.com/GoogleChromeLabs/chromium-bidi/issues/620#issuecomment-1511610695) works, as long as it's not awaited.

PR #621, opened 2023-04-17
DRAFT: try to reproduce #620

thiagowfx:


thiagowfx:
The test is passing, but I was purposely trying to make it fail in order to reproduce the bug. I could not.

tests/test_browsing_context.py
```diff
@@ -732,6 +732,29 @@ async def test_browsingContext_navigateSameDocumentNavigation_waitComplete_navig
...
 
+@pytest.mark.asyncio
+async def test_browsingContext_logAfterNavigation(websocket, context_id, html):
+    await subscribe(websocket, "log.entryAdded")
+
+    await send_JSON_command(
+        websocket, {
+            "method": "browsingContext.navigate",
+            "params": {
+                "url": html("<script>console.log('mylog');</script>"),
```
sadym-chromium, line 743: try OOPiF sending log. I believe the root cause is there

sadym-chromium: Eg you can use Mathias' demo (as long as our e2e is not local anyways)

thiagowfx:
During the stand-up yesterday, we discussed using `another_context_id` to trigger the failure. The test still passes though.

PR #623, opened 2023-04-17
Introduce Default Map, inspired by defaultdict in python

thiagowfx:
It's like a plain ES6 Map, with the difference that it returns a default value whenever the key is absent from the Map.

This simplifies constructs like:

    if (!this.has(key)) {
        this.set(key, new Set());
    }
    this.get(key)!.add(1);

To simply:

    this.get(key).add(1);

### DIFF
--- a/src/bidiMapper/domains/events/EventManager.ts
+++ b/src/bidiMapper/domains/events/EventManager.ts
@@ -25,6 +25,7 @@ import type {BidiServer} from '../../BidiServer.js';
 import {Buffer} from '../../../utils/buffer.js';
 import {IdWrapper} from '../../../utils/idWrapper.js';
 import {OutgoingBidiMessage} from '../../OutgoingBidiMessage.js';
+import {DefaultMap} from '../../../utils/DefaultMap.js';
 
 import {SubscriptionManager} from './SubscriptionManager.js';
 
@@ -96,10 +97,10 @@ export class EventManager implements IEventManager {
    * Needed for getting buffered events from all the contexts in case of
    * subscripting to all contexts.
    */
-  #eventToContextsMap = new Map<
+  #eventToContextsMap = new DefaultMap<
     string,
     Set<CommonDataTypes.BrowsingContext | null>
-  >();
+  >(() => new Set());
   /**
    * Maps `eventName` + `browsingContext` to buffer. Used to get buffered events
    * during subscription. Channel-agnostic.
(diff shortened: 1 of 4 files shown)


src/bidiMapper/domains/events/EventManager.ts
```diff
@@ -312,7 +310,7 @@ export class EventManager implements IEventManager {
 
     if (contextId === null) {
       // For global subscriptions, events buffered in each context should be sent back.
-      Array.from(this.#eventToContextsMap.get(eventName)?.keys() ?? [])
+      Array.from(this.#eventToContextsMap.get(eventName).keys() ?? [])
```
sadym-chromium, line 313: This will create a set for each event user subscribed to, not necessary existing. But I guess it's fine.

src/utils/DefaultMap.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+import {DefaultMap} from './DefaultMap.js';
+
+const expect = chai.expect;
+
+describe.only('DefaultMap', () => {
+  it('returns the default value when key does not exist', () => {
+    const defaultValue = 0;
+
+    const cutenessMap = new DefaultMap<string, number>(
+      () => defaultValue,
```
sadym-chromium, line 29: Can you please use a bit less trivial default value, like `42`?

src/utils/DefaultMap.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+      [['dog', 100]]
+    );
+
+    expect(cutenessMap.get('dog')).to.deep.equal(100);
+    expect(cutenessMap.get('cat')).to.deep.equal(defaultValue);
```
sadym-chromium, line 34: Also we should test `.values()` contains `cat` after accessing it. We check that we don't return a new default each time.

src/bidiMapper/domains/events/EventManager.ts
thiagowfx: Yes, in the end the loop would just be skipped as the set is empty. Unless we rely on the empty set semantics elsewhere, it shouldn't be an issue.

src/utils/DefaultMap.spec.ts
thiagowfx: Very reasonable. Done.

src/utils/DefaultMap.spec.ts
thiagowfx: Done...I think(?) You probably meant `keys()`, not `values()`.

PR #629, opened 2023-04-20
Fix BrowsingContext re-initialisation race condition

sadym-chromium:
The root cause of the #620 was exactly as described @Lightning00Blade. While `Page.frameNavigated` event was fired before `Runtime.executionContextCreated`, as it's async, it's last part with cleaning realms was run after `Runtime.executionContextCreated` handled.

This fix includes 2 parts:
1. Remove all the async event handlers.
2. Rely on the `Runtime.executionContextsCleared` event rather then `Page.frameNavigated` to clean Realms.

Tested manually with Puppeteer and repro script from the https://github.com/puppeteer/puppeteer/issues/10032. Couldn't reproduce with Mapper Tab and therefor couldn't add e2e test.

### DIFF
--- a/src/bidiMapper/domains/context/browsingContextImpl.ts
+++ b/src/bidiMapper/domains/context/browsingContextImpl.ts
@@ -282,9 +282,6 @@ export class BrowsingContextImpl {
         // previous page are detached and realms are destroyed.
         // Remove context's children.
         await this.#deleteChildren();
-
-        // Remove all the already created realms.
-        this.#realmStorage.deleteRealms({browsingContextId: this.contextId});
       }
     );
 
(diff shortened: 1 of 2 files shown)


src/bidiMapper/domains/context/browsingContextImpl.ts
```diff
@@ -180,8 +180,8 @@ export class BrowsingContextImpl {
     this.#children.set(child.contextId, child);
   }
 
-  async #deleteChildren() {
-    await Promise.all(this.children.map((child) => child.delete()));
+  #deleteChildren() {
```
thiagowfx, line 183: delete() is not `async`, I wonder why there was an `await` here in the first place?

I was going to suggest to add a comment here to state that this is purposely without `await`, but there's no need because the underlying function isn't async anyway.

thiagowfx (approved):

src/bidiMapper/domains/context/browsingContextImpl.ts
thiagowfx: Actually, disregard this, I reviewed an older commit. Let me take another look.

Lightning00Blade:
Closed #620, Closed #621

PR #630, opened 2023-04-21
Add PreloadScriptIdStorage and implement add/remove preload script

thiagowfx:
Bug: #293, #584

PR #633, opened 2023-04-21
github actions: add timeout for wpt tests

thiagowfx:


thiagowfx:
https://github.com/GoogleChromeLabs/chromium-bidi/pull/630 is running for >1h

.github/workflows/wpt-chromedriver.yml
```diff
@@ -10,6 +10,7 @@ on:
 
 jobs:
   wpt:
+    timeout-minutes: 20
```
Lightning00Blade, line 13: I would put it on the step rather then the job, we have observe some slow set-ups from time to time.

thiagowfx: Didn't know it was possible. Done!

Lightning00Blade (approved):

PR #636, opened 2023-04-24
rollup: add output.globals for crypto module

thiagowfx:
Otherwise the mapper times out. This fixes the following error message in the logs:

	lib/cjs/bidiTab/bidiTab.js → lib/iife/mapperTab.js...
	(!) Missing global variable name
	Use output.globals to specify browser global variable names corresponding to external modules
	crypto (guessing 'require$$0')
	created lib/iife/mapperTab.js in 1s
	✅ [rollup] Executed successfully

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,8 @@
         "tsc"
       ],
       "files": [
-        "lib/cjs/bidiMapper/index.js"
+        "lib/cjs/bidiMapper/index.js",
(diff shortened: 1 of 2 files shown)


Lightning00Blade (approved):

thiagowfx:
https://github.com/ionic-team/rollup-plugin-node-polyfills would have been another way to fix this issue.

PR #638, opened 2023-04-24
CI: make wpt-mapper use @puppeteer/browsers package

thiagowfx:
All other CI actions already use it, this is the last one left to migrate.


thiagowfx:
Dunno why the CI fails when using the puppeteer package.

sadym-chromium (approved):

PR #640, opened 2023-04-24
wireit and npm run server improvements

thiagowfx:
Follow up of https://github.com/GoogleChromeLabs/chromium-bidi/commit/5678af1c65dcb062ace20897d31dc599c82573d7, with a few improvements.

This also fixes `--verbose`, broken by #603.


package.json
```diff
@@ -85,10 +84,7 @@
       },
       "dependencies": [
         "build"
-      ],
-      "env": {
-        "DEBUG": "bidiServer:log"
-      }
```
thiagowfx, line 91: Interesting dilemma here. `wireit` makes the server start-up more reliable and less flaky as we can rely on `lineMatches` above to ensure the server has fully started.

OTOH, this requires `DEBUG=bidiServer:log` or `DEBUG=bidiServer:*` or `DEBUG=*` to be set. Hard-coding it here is a bad idea because then the `DEBUG` environment variable is ignored (e.g. breaks `--verbose`).

Let me try to find a compromise.

thiagowfx: `runBiDiServer.sh` sets `DEBUG=*` by default. Everything in this repo (including `runWPT.sh`) uses this script indirectly. This should suffice, except a few github actions.

There are two options:

1) Make the github actions use `runWPT.sh`, OR
2) Set `DEBUG=` in the github actions

thiagowfx: I'll go with the second approach because it's quicker. Later on we can refactor 1).

runWPT.sh
```diff
@@ -34,15 +34,22 @@ readonly WPT_BROWSER_PATH="${WPT_BROWSER_PATH:-/Applications/Google Chrome Dev.a
 # The path to the WPT report file.
 readonly WPT_REPORT="${WPT_REPORT:-wptreport.json}"
 
-# The path to the WPT metadata directory.
+# Only set WPT_METADATA if it's not already set.
+if [ -z ${WPT_METADATA+x} ]; then
```
Lightning00Blade, line 38: Can you give a little bit of inside on why this change is made? I have no context.

sadym-chromium: This line allows providing a custom metadata folder, which is very useful, if you want to re-baseline the tests.

sadym-chromium (approved):

Lightning00Blade (approved):

PR #645, opened 2023-04-25
fix: network request respects hasExtraInfo field

Lightning00Blade:
If the Mapper encountered a Request that had no Extra info it would wait for the request response indefinitely.
That would in term block the `ProcessingQueue` from doing anything else.
Puppeteer spawns only 1 mapper per spec so if one test backs in this way all the following test will timeout.
We need to think of a more robust solution. 

Misc fix:
`wireit` broke `npm link` as some files were restored by it during `npm prepare`.


package.json
```diff
@@ -114,7 +113,7 @@
         "src/**/*.ts"
       ],
       "output": [
-        "lib/**"
+        "lib/cjs/*"
```
Lightning00Blade, line 116: Explicitly point to dir so `wireit` creates the proper cache else every rerun would replace the file leading to a possible bricked state.

package.json
```diff
@@ -55,7 +55,6 @@
     },
     "prepare": {
       "dependencies": [
-        "clean",
```
Lightning00Blade, line 58: This should be fine as we usually use it only on Publish from what I understand where no cache exist

sadym-chromium: NIT: next time please split the commit into 2, as in case of issues with `package` we will need to revert the network as well

package.json
thiagowfx: Don't you want `lib/cjs/**` here, to include subdirectories as well?

Lightning00Blade: Ops, that is correct, thank you for noticing! 
```suggestion
        "lib/cjs/**"
```

thiagowfx (approved):

sadym-chromium (approved):

PR #647, opened 2023-04-25
Use `runBiDiServer.sh` as a single entry point

sadym-chromium:
The goal is to provide verbose logs of e2e run to help investigate failures if any.

### DIFF
--- a/src/bidiServer/index.ts
+++ b/src/bidiServer/index.ts
@@ -64,6 +64,7 @@ function parseArguments() {
   parser.add_argument('-v', '--verbose', {
     help: 'If present, the Mapper debug log, including CDP commands and events will be logged into the server output.',
     action: 'store_true',
+    default: process.env['VERBOSE'] === 'true' || false,
   });
 
   const args = parser.parse_known_args();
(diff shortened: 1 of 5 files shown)


src/bidiServer/index.ts
```diff
@@ -130,12 +130,12 @@ async function onNewBidiConnectionOpen(
...
-    ? computeSystemExecutablePath({
-        browser: Browser.CHROME,
-        channel: chromeChannel,
-      })
-    : process.env['CHROME_BIN'];
+  const executablePath =
+    process.env['CHROME_BIN'] ??
+    computeSystemExecutablePath({
+      browser: Browser.CHROME,
+      channel: chromeChannel || ChromeReleaseChannel.DEV,
```
thiagowfx, line 137:
```suggestion
      channel: chromeChannel ?? ChromeReleaseChannel.DEV,
```

thiagowfx: Or, even better: Add a default argument to `chromeChannel` in the method signature.

thiagowfx (approved):

sadym-chromium:
> The only thing that is missing -- but can be a separate PR, maybe I'll tackle it -- is to use runWPT.sh as a single entry point (as well), for the CI actions

Out of scope of this PR, as the goal was to provide e2e logging.

thiagowfx:
I have another theory about the E2E error.

Let's try to downgrade `"@puppeteer/browsers": "0.1.1"` in package.json. It was changed ~4 days ago by myself, and it seems to coincide with the recent flakiness.

Issue #650, opened 2023-04-26
Implement `browsingContext.reload`


OrKoN:
Spec: https://w3c.github.io/webdriver-bidi/#command-browsingContext-reload

## Tracking

- [x] Implementation
  - [x] Handle `wait`: `interactive` and `complete` (implementation)
- [x] Ping(2023-05-31): E2E tests
  - [x] basic
  - [x] wait
  - [x] ~~`ignoreCache`: https://github.com/GoogleChromeLabs/chromium-bidi/pull/742~~
  - [x] Ping(2023-05-31): `ignoreCache`: https://github.com/GoogleChromeLabs/chromium-bidi/pull/794
- [x] WPT tests (upstream)
  - [x] validation: https://github.com/web-platform-tests/wpt/pull/39739
  - [x] functionality: https://github.com/web-platform-tests/wpt/pull/40134
- [x] ~~(?) support for local server~~

OrKoN:
cc @sadym-chromium @thiagowfx (let me know if I overlooked it, perhaps types are not right)

thiagowfx:
You didn't, it's indeed missing. I am sending an initial PR shortly.

sadym-chromium:
It's odd that there are no WPT tests for the command

whimboo:
Well, this command hasn't been implemented yet in Firefox. So you are very welcome to create the related wdspec tests. Thanks!

thiagowfx:
Taking a stab at it here: https://github.com/web-platform-tests/wpt/pull/39739

thiagowfx:
Upstream WPT CL is complete and approved. Just waiting for merge. Closing.

PR #654, opened 2023-04-27
Implement browsing context reload

thiagowfx:
Bug: #650
CDP: https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-reload
Spec: https://w3c.github.io/webdriver-bidi/#command-browsingContext-reload


sadym-chromium:
I don't think it makes sense to merge changes without any test coverage unless we have very strong arguments for that

sadym-chromium (commented):
In general approach seems correct, but I don't think it makes sense to merge changes without any test coverage unless we have very strong reasons for that.

sadym-chromium:
sorry for spam, I was fighting with the UI )

thiagowfx:
@sadym-chromium What kind of E2E do you think it's appropriate to add? Which events are we interested to subscribe to?

src/bidiMapper/domains/context/browsingContextImpl.ts
```diff
@@ -519,6 +518,28 @@ export class BrowsingContextImpl {
...
+
+    switch (wait) {
+      case 'none':
+        break;
+      // TODO: handle interactive.
+      case 'interactive':
+        break;
+      // TODO: handle complete.
+      case 'complete':
+        break;
```
sadym-chromium, line 537: Can we return "not implemeted" error instead?

tests/test_browsing_context.py
```diff
@@ -786,3 +786,19 @@ async def test_browsingContext_waitForSelectorWithMissingElement_failedWithTimeo
...
     pass
+
+
+@pytest.mark.asyncio
+async def test_browsingContext_reload(websocket, context_id, html):
+    url = html()
+
+    await goto_url(websocket, context_id, url)
+
+    result = await execute_command(websocket, {
```
sadym-chromium, line 797: We should verify that the actual reload eventually happened. You can use `test_browsingContext_navigateWait...`  as blueprints.

sadym-chromium:
> @sadym-chromium What kind of E2E do you think it's appropriate to add? Which events are we interested to subscribe to? I added a basic one.

Same as for navigation: `test_browsingContext_navigateWait...`.

src/bidiMapper/domains/context/browsingContextImpl.ts
```diff
@@ -519,6 +518,28 @@ export class BrowsingContextImpl {
     };
   }
 
+  async reload(
+    ignoreCache: boolean,
+    wait: BrowsingContext.ReadinessState
+  ): Promise<Message.EmptyResult> {
+    await this.#cdpTarget.cdpClient.sendCommand('Page.reload', {
+      ignoreCache,
```
sadym-chromium, line 526: To test it properly we need a local server. Than we could make a page returning current timestamp each time it is loaded, and after reload verify the content has changed or not.

The local server is not setup for e2e, but it is for WPT, so I guess we can test it from WPT, as from our perspective it's only 1 flag.

src/bidiMapper/domains/context/browsingContextImpl.ts
thiagowfx: Done

whimboo:
Hi @thiagowfx, could you please give an update what specifically is left to do for this PR? Thanks.

thiagowfx:
@whimboo Just E2E tests, which I am planning to write this week.

I plan to address the `wait` conditions in a follow-up PR (no ETA on this yet, hopefully this week but no promises).

src/bidiMapper/domains/context/browsingContextImpl.ts
thiagowfx: Ack

tests/test_browsing_context.py
thiagowfx: Done

thiagowfx:
@sadym-chromium E2E tests are added, this is ready to go.

sadym-chromium (approved):
`ignoreCache` is not tested. Please add comment regarding that, create an issue, or note it in some other way, so that we don't forget it. Or ideally just addd an e2e checking network events.

thiagowfx:
Sure, I'll add a comment in the child PR (`wait`).

thiagowfx:
@whimboo This is done now, in case you aren't subscribed.

whimboo:
@thiagowfx thanks. Will those tests be upstreamed to the wpt repository?

thiagowfx:
Yes, I will do so.

PR #668, opened 2023-05-02
chore: revert Protocol to Puppeteer's version

Lightning00Blade:
Building mapper locally (`npm link`) will use it's `node_modules` so we will have two different version of `devtools-protocol` of the Puppeteer side. That makes the Types incompatible with each other. Let's keep them in sync unless that is something that mapper need.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -33,7 +33,7 @@
         "chai-as-promised": "7.1.1",
         "chai-exclude": "2.1.0",
         "debug": "4.3.4",
-        "devtools-protocol": "0.0.1134390",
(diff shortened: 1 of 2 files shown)


thiagowfx:
I'd ask you to add a comment but `package.json` doesn't support comments AFAIK (https://stackoverflow.com/questions/14221579/how-do-i-add-comments-to-package-json-for-npm-install)

Could you file a bug in either bugtracker to ensure we add a mechanism to ensure they stay in sync?

thiagowfx:
I think a `pre-commit` hook could work in this case, I can try to add it later.

thiagowfx (approved):

PR #670, opened 2023-05-02
pre-commit: ensure devtools-protocol version is synced with puppeteer

thiagowfx:
Enforces e.g. https://github.com/GoogleChromeLabs/chromium-bidi/pull/668


thiagowfx:
I wonder if we really need a pre-commit hook though. Would it also work to set `peerDependencies.devtools-protocol` in `package.json` to a fixed value, instead of `'*'`?

OrKoN:
I don't think we should ensure the same version long-term. Can we change the interfaces in such way so that devtools-protocol is not part of the public chromium-bidi interface?

thiagowfx:
We use `devtools-protocol` for type annotations, mainly. @Lightning00Blade, what exactly was breaking on puppeteer's side with the version mismatch?

thiagowfx:
It's already declared as a peer dependency, see `package.json` in our repo.

Lightning00Blade:
Let's say Puppeteer keep version X, and BiDi Mapper version Y. Due to the first being a Dependency and the latter being a DevDependency, only the Puppeteer one will be installed when running `npm install` (Mapper has a PeerDependency of '*'). So normal user will never face this issue. 
The problem is when trying thing with local version of Puppeteer + local version of Mapper.
Using `npm link`. Puppeteer's Protocol resolves to X, then you go to Mapper Protocol which tries to resolve node_modules but due to being symlink it does not look at Puppeteer's `node_modules` folder but it's own.
Then during building phase TypeScripts tries to match the `CDP.send` commands which are strongly typed, but there are some extra types on Protocol Y that don't exist on Protocol X, thus type become incomparable -> build fails.

OrKoN:
@Lightning00Blade how about using the `pack -> install` workflow for local testing? it is a bit longer but more similar to what happens for the users.

thiagowfx:
@Lightning00Blade did you have a chance to see if the native `npm` approach makes sense?

Lightning00Blade:
It mostly works. Puppeteer builds even the version are mismatched and changes exits, (used an old and a new version in Mapper to test). The only problem is the Puppeteer's `package.json` is changed but as long as we don't commit it we are good.

thiagowfx:
Deleting branch. For reference:

```
pre-commit hook:

  - repo: local
    hooks:
      - id: puppeteer-devtools-protocol
        name: check devtools-protocol in package.json matches the one used by puppeteer
        files: ^package.json$
        language: system
        entry: .hooks/puppeteer-devtools-protocol.sh
```

```
puppeteer-devtools-protocol.sh:

#!/bin/bash
set -euo pipefail

LOCAL="$(npm list devtools-protocol | grep devtools-protocol | cut -d "@" -f2)"
REMOTE="$(npx npm-remote-ls puppeteer@latest | grep devtools-protocol | cut -d "@" -f2)"

if [[ "$LOCAL" != "$REMOTE" ]]; then
    echo "devtools-protocol version ($LOCAL) does not match the one in puppeteer ($REMOTE)"
    exit 1
fi
```

Lightning00Blade:
A small update to this issue - 
Just tried it after #691 was merged and it yield TypeScript error with both `npm link` and `npm pack | npm install (...).tgz`  
```
../../node_modules/chromium-bidi/lib/cjs/bidiMapper/domains/script/realm.d.ts(32,59): error TS2694: Namespace '"/usr/local/google/home/nvitkov/puppeteer/node_modules/devtools-protocol/types/protocol".Protocol.Runtime' has no exported member 'DeepSerializedValue'.
```
I will try to create a workaround from Puppeteer side.

PR #673, opened 2023-05-02
preload script

thiagowfx:


thiagowfx:
@sadym-chromium I am blocked here.

```
> chromium-bidi@0.4.7 build
> wireit

✅ [tsc] Already fresh
✅ [rollup] Already fresh
✅ [build] No command to execute
(./runWPT.sh) Running WPT in headless mode...
(./runWPT.sh) Using pure mapper...
Running 1 tests in web-platform-tests

  ▶ Unexpected subtest result in /webdriver/tests/bidi/script/remove_preload_script/remove_preload_script.py:
  │ FAIL [expected PASS] test_remove_one_of_preload_scripts
  │   → AssertionError: assert equals failed
  │
  │ bidi_session = <webdriver.bidi.client.BidiSession object at 0x112ac7d90>
  │
  │     @pytest.mark.asyncio
  │     async def test_remove_one_of_preload_scripts(bidi_session):
  │         script_1 = await bidi_session.script.add_preload_script(
  │             function_declaration="() => { window.bar='foo'; }"
  │         )
  │         script_2 = await bidi_session.script.add_preload_script(
  │             function_declaration="() => { window.baz='bar'; }"
  │         )
  │
  │         # Remove one of the scripts
  │         await bidi_session.script.remove_preload_script(script=script_1)
  │
  │         new_tab = await bidi_session.browsing_context.create(type_hint="tab")
  │
  │         # Check that the first script didn't run
  │         result = await bidi_session.script.evaluate(
  │             expression="window.bar",
  │             target=ContextTarget(new_tab["context"]),
  │             await_promise=True,
  │         )
  │         assert result == {"type": "undefined"}
  │
  │         # Check that the second script still applied the changes to the window
  │         result_2 = await bidi_session.script.evaluate(
  │             expression="window.baz",
  │             target=ContextTarget(new_tab["context"]),
  │             await_promise=True,
  │         )
  │ >       assert result_2 == {"type": "string", "value": "bar"}
  │ E       AssertionError: assert equals failed
  │ E         -{'type': 'undefined'}   +{
  │ E                                 +  'type': 'string',
  │ E                                 +  'value': 'bar',
  │ E                                 +}
  │
  │ bidi_session = <webdriver.bidi.client.BidiSession object at 0x112ac7d90>
  │ new_tab    = {'children': [],
  │  'context': '2C79C2E49241962AD97EBA1D63E7F7A9',
  │  'parent': None,
  │  'url': 'about:blank'}
  │ result     = {'type': 'undefined'}
  │ result_2   = {'type': 'undefined'}
  │ script_1   = '431d2ab8-49f7-428f-8552-7c90f93b9b15'
  │ script_2   = 'd4f7a9de-930c-44c7-a445-1cb449c4d18b'
  │
  └ wpt/webdriver/tests/bidi/script/remove_preload_script/remove_preload_script.py:77: AssertionError

Ran 1 tests finished in 3.0 seconds.
  • 0 ran as expected. 0 tests skipped.
  • 1 tests had unexpected subtest results
```

This test is failing. It should not, as per the current diff. Please advise.

thiagowfx:
Actually, I repurposed this CL. Never mind.

src/bidiMapper/domains/context/PreloadScriptStorage.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+import {PreloadScriptStorage} from './PreloadScriptStorage';
+
+describe('PreloadScriptStorage', () => {
+  let preloadScriptStorage: PreloadScriptStorage;
+
+  beforeEach(() => {
+    preloadScriptStorage = new PreloadScriptStorage();
+  });
+
+  it('initial state', () => {
```
sadym-chromium, line 29: I assume you wanted to add some more tests

src/bidiMapper/domains/context/browsingContextProcessor.ts
```diff
@@ -187,6 +187,28 @@ export class BrowsingContextProcessor {
...
+        contextId: null,
+      });
+
+      // Add them all to the new context.
+      for (const script of scripts) {
+        const functionDeclaration = script.functionDeclaration;
+        const sandbox = script.sandbox;
+
+        // The spec provides a function, and CDP expects an evaluation.
+        const cdpPreloadScriptId = await context.cdpTarget.addPreloadScript(
```
sadym-chromium, line 202: There will be a race condition between [`CdpTarget.create` -> `void cdpTarget.#unblock()`](https://github.com/GoogleChromeLabs/chromium-bidi/blob/43882f46dca6e203b0bdbb12c4aa9d79a6a3895b/src/bidiMapper/domains/context/cdpTarget.ts#L55) and adding preload script. Adding preload script should be a part of [`async #unblock()`](https://github.com/GoogleChromeLabs/chromium-bidi/blob/43882f46dca6e203b0bdbb12c4aa9d79a6a3895b/src/bidiMapper/domains/context/cdpTarget.ts#L55).

sadym-chromium: The easiest way to achieve that is to add a parameter `preloadScripts` to [`CdpTarget.create`](https://github.com/GoogleChromeLabs/chromium-bidi/blob/43882f46dca6e203b0bdbb12c4aa9d79a6a3895b/src/bidiMapper/domains/context/cdpTarget.ts#L36)

thiagowfx: Do you really mean `create` of CdpTarget, not of browsingContextImpl?

sadym-chromium: it depends. From BiDi perspective it sticks to BrowserContext. And we tend to base structure on the BiDi. In other words, it can be both in Target and in BrowsingcontextProcessor. Up to you.

sadym-chromium: I'm thinking if we should add e2e test which can expose such a race condition. E.g. add a preloadScript, and create a new tab with console.log(), and verify the event order.

src/bidiMapper/domains/context/cdpTarget.ts
```diff
@@ -38,34 +41,40 @@ export class CdpTarget {
...
 
     LogManager.create(cdpTarget, realmStorage, eventManager);
 
     cdpTarget.#setEventListeners();
 
-    // No need in waiting. Deferred will be resolved when the target is unblocked.
-    void cdpTarget.#unblock();
+    // No need in waiting.
+    // Deferred will be resolved when the target is unblocked.
+    cdpTarget.#unblock();
```
sadym-chromium, line 61: please keep `void` if async function is not awaited intentionally

thiagowfx: Sure. In this case I removed it because `create` is not `async`, so `await` cannot be added anyway.

src/bidiMapper/domains/context/cdpTarget.ts
```diff
@@ -115,6 +124,9 @@ export class CdpTarget {
     });
 
     await this.#cdpClient.sendCommand('Runtime.runIfWaitingForDebugger');
+
+    await this.loadTopLevelPreloadScripts();
```
sadym-chromium, line 128: It's not always top-level. It can be preload scripts for it's parent context.

sadym-chromium: Please add e2e for such a case

sadym-chromium: You can move it to another PR though, but please add a failing e2e anyways in this PR.

sadym-chromium, line 128: We want to setup preload scripts BEFORE the target started living it's own life.
```suggestion
    await this.loadTopLevelPreloadScripts();

    await this.#cdpClient.sendCommand('Runtime.runIfWaitingForDebugger');
```

src/bidiMapper/domains/context/PreloadScriptStorage.spec.ts
thiagowfx: Only later on, once this class is stable. This is purposely just skeleton code for now.

thiagowfx: I included a TODO at the top.

src/bidiMapper/domains/context/cdpTarget.ts
thiagowfx: As discussed during live code review, we already have such a test in `test_preload_script.py`, which is currently skipped.

src/bidiMapper/domains/context/PreloadScriptStorage.ts
```diff
@@ -119,17 +132,40 @@ export class PreloadScriptStorage {
    * Keeps track of the given CDP preload script in the given BiDi preload
    * script.
    */
-  appendPreloadScript(
-    script: BidiPreloadScript,
+  appendCdpPreloadScript(
+    script: BidiPreloadScript, // XXX: id?
```
sadym-chromium, line 136: NIT: I don't quite understand the comment. Could you please extend it a bit?

src/bidiMapper/domains/context/browsingContextProcessor.ts
```diff
@@ -295,13 +277,13 @@ export class BrowsingContextProcessor {
...
+      throw new Error('add preload script arguments are not supported');
     }
 
+    // XXX: Stop using Set.
     const cdpTargets = new Set<CdpTarget>(
-      // TODO: flatten children and deduplicate.
+      // TODO: Flatten children.
       params.context === undefined || params.context === null
         ? this.#browsingContextStorage
             .getTopLevelContexts()
```
sadym-chromium, line 289: Again, the unique target can be in non-top-level browsing context. What we really need here is all the targets, so we can walk through all the contexts and collect their targets into the set.

src/bidiMapper/domains/context/PreloadScriptStorage.ts
```diff
@@ -119,17 +132,40 @@ export class PreloadScriptStorage {
...
   ) {
     script.cdpPreloadScripts.push(cdpPreloadScript);
   }
 
-  /** Deletes all entries that match the given filter. */
-  removePreloadScripts(filter?: BidiPreloadScriptFilter) {
+  /** Deletes all BiDi preload script entries that match the given filter. */
+  removeBiDiPreloadScripts(filter?: BidiPreloadScriptFilter) {
     for (const preloadScript of this.findPreloadScripts(filter)) {
       this.#scripts.delete(preloadScript);
```
sadym-chromium, line 145: We need to de-register them. Don't we?

src/bidiMapper/domains/context/PreloadScriptStorage.ts
thiagowfx: The idea was to do this:

```
    let preloadScript: BidiPreloadScript;

    if (typeof script === 'string') {
      preloadScript = this.getPreloadScript({id: script});
    } else {
      preloadScript = script;
    }

    preloadScript.cdpPreloadScripts.push(cdpPreloadScript);
  }
```

But now I realize we don't really need to use it. I'll just remove the comment.

src/bidiMapper/domains/context/cdpTarget.ts
```diff
@@ -147,6 +165,27 @@ export class CdpTarget {
     });
   }
 
+  /** Loads all top-level and parent preload scripts. */
+  async loadPreloadScripts() {
+    for (const script of this.#preloadScriptStorage.findPreloadScripts({
+      contextIds: [null, this.#parentTargetId],
```
sadym-chromium, line 171: +1

sadym-chromium (approved):
LGTM with a few NITs

src/bidiMapper/domains/context/browsingContextProcessor.ts
thiagowfx: Is it OK to do it in a follow up?

I am aware something needs to change here, hence the TODO.

I'll incorporate your comment in the TODO.

thiagowfx: Updated. Please let me know if you're OK with doing this in a follow-up PR.

src/bidiMapper/domains/context/browsingContextProcessor.ts
```diff
@@ -295,13 +277,13 @@ export class BrowsingContextProcessor {
...
   ): Promise<Script.AddPreloadScriptResult> {
     if (params.arguments !== undefined && params.arguments.length > 0) {
-      throw new Message.UnsupportedOperationException(
-        'add preload script arguments are not supported'
-      );
+      // TODO: Handle arguments.
+      throw new Error('add preload script arguments are not supported');
     }
 
+    // XXX: Stop using Set.
```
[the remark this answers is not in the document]
thiagowfx: Removed comment.

thiagowfx:
>  BTW, what are the failing WPT tests?

Hmm, last time I checked they were all passing. Let me check it again.

> And did you had a chance to try running all the disabled tests?

They still fail.

thiagowfx:
> Hmm, last time I checked they were all passing. Let me check it again.

Ah, they were from https://github.com/GoogleChromeLabs/chromium-bidi/commit/53f9c36b7bdf79d3719289acb222f375197fb0ba

thiagowfx:
Note: E2E failures are unrelated. They have to do with serialization.

src/bidiMapper/domains/context/browsingContextProcessor.ts
thiagowfx: Will carry-over to a checkbox in #293

PR #675, opened 2023-05-03
bidi server: close CDP connection before closing the browser

thiagowfx:
1. catch unhandled errors in src/bidiServer/mapperServer.ts (not sure if there is anything better to do with them?)
2. close mapperServer when the bidi client connection is gone
3. provide proper stack traces for errors
4. clean up callbacks after they are resolved or rejected

P.S. there is still an unintended Runtime.evaluate with the browserContext.getTree call pending when the test is over. Without #1 it crashes the server. We could not figure out where it comes from.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -60,6 +60,7 @@ jobs:
           path: logs
 
 env:
-  DEBUG: bidiServer:log
+  DEBUG: 'bidiServer:log,bidiMapper:mapperDebug:*'
   FORCE_COLOR: 3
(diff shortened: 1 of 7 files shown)


OrKoN (approved):
LGTM

thiagowfx:
E2E tests fail. Let me know if you think of a better approach.

sadym-chromium (approved):

thiagowfx:
FYI: I am still observing E2E failures occasionally. Unfortunately I do not think this has completely fixed the issue (although, it's a good PR anyway!).

PR #677, opened 2023-05-03
update WPT expectations for unimplemented commands

thiagowfx:


thiagowfx:
This will make WPT tests run faster and reduce the noise. If we don't do this, then e.g. in my print to pdf PR (https://github.com/GoogleChromeLabs/chromium-bidi/actions/runs/4873106610/jobs/8692210520?pr=672) we see unrelated noise
image

thiagowfx:
This screenshot makes it clearer: 
image

runWPT.sh
```diff
@@ -103,6 +104,9 @@ if [[ "$UPDATE_EXPECTATIONS" == "true" ]]; then
     "$WPT_REPORT"
 fi
 
+# Script status should match "wpt run" for CI.
+exit "$status"
```
sadym-chromium, line 108: +1

src/bidiMapper/domains/network/networkRequest.ts
```diff
@@ -23,9 +23,9 @@
 
 import Protocol from 'devtools-protocol';
 
-import {Deferred} from '../../../utils/deferred';
-import {IEventManager} from '../events/EventManager';
-import {Network} from '../../../protocol/protocol';
+import {Deferred} from '../../../utils/deferred.js';
```
sadym-chromium, line 26: NIT: ideally this should be in a separate PR to reduce noise

sadym-chromium (approved):

src/bidiMapper/domains/network/networkRequest.ts
thiagowfx: Agreed. This should have been reverted, it was accidentally merged. Removed.

thiagowfx:
This wasn't failing before. Let me see what's wrong.

thiagowfx:
> This wasn't failing before. Let me see what's wrong.

We're stuck in a loop. Some tests keep changing from PASS to FAIL and vice-versa. I am extracting the expectations from the artifacts, but whenever I update them, they change again in the next run.

thiagowfx:
> The latest of CfT is not the same as latest Chromium.

Are we talking about Dev <-> Canary, or Dev <-> Stable? Still, it doesn't explain the timeouts / unstability / flakiness. Very odd.

Pinning is an option, but since we spent so much time on this matter this week, let's defer that to another time.

PR #680, opened 2023-05-03
TS: ensure all local imports have a .js suffix

thiagowfx:


thiagowfx:
Alex, IIRC you mentioned it once in a past PR that this should be the case, but please correct me if I remember it wrongly.

OrKoN (approved):

thiagowfx:
Ah, this should have been merged against main, not against the parent branch. It’s really inconvenient to work with multiple children PRs on GitHub.

OrKoN:
Ah I overlooked it. I think this kind of change should not be made dependent on other branch. Should we revert-reland?

thiagowfx:
Not your fault, i could easily have done the same mistake. Yes we should revert. I can do that.

PR #681, opened 2023-05-03
prefactor browsing context classes and add all children method

thiagowfx:
The main purpose of this PR is to add `allChildren()` (flattened), but I did some prefactoring as well.

The main benefit now is that `BrowsingContextImpl` is now only stored in the storage. It's not duplicated in the `Impl` class itself via children. This makes it easier to maintain it / reason about it.

### DIFF
--- a/src/bidiMapper/domains/context/browsingContextImpl.ts
+++ b/src/bidiMapper/domains/context/browsingContextImpl.ts
@@ -83,8 +83,6 @@ export class BrowsingContextImpl {
     this.#eventManager = eventManager;
     this.#browsingContextStorage = browsingContextStorage;
     this.#logger = logger;
-
-    this.#initListeners();
   }
 
   static create(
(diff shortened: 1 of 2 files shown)


src/bidiMapper/domains/context/PreloadScriptStorage.ts
```diff
@@ -16,9 +16,9 @@
  */
 import {v4 as uuidv4} from 'uuid';
 
-import {CommonDataTypes, Script} from '../../../protocol/protocol';
+import {CommonDataTypes, Script} from '../../../protocol/protocol.js';
```
sadym-chromium, line 19: Can it be done once throughout the whole project in one dedicated PR?

thiagowfx: Done: https://github.com/GoogleChromeLabs/chromium-bidi/commit/28e665442640cf1db9b114b993cc40b3a74aae06

sadym-chromium (approved):
Keeping one source of truth make sense. LGTM

PR #682, opened 2023-05-04
TS: ensure all local imports have a .js suffix

thiagowfx:
Recreating it from https://github.com/GoogleChromeLabs/chromium-bidi/pull/680


thiagowfx:
Since this is exactly the same as https://github.com/GoogleChromeLabs/chromium-bidi/pull/680, which was already approved (but mistakenly merged in another branch), I'll go ahead with it to save time.

PR #683, opened 2023-05-04
Use Chromium instead of CfT

sadym-chromium:
Experiments showed that using Chromium instead of CfT increased e2e tests stability drastically: https://github.com/GoogleChromeLabs/chromium-bidi/actions/workflows/e2e.yml?query=branch%3Asadym%2Fe2e-experiment (disregard the last few runs, where wrong bin path was provided).

I suggest using Chromium until we fins out the root cause.


thiagowfx (approved):

OrKoN:
The problem is probably not CfT itself but the difference in versions. The latest of CfT is not the same as latest Chromium. Perhaps it makes sense to pin a particular version and update it daily/weekly.

thiagowfx:
> The latest of CfT is not the same as latest Chromium.

Are we talking about Dev <-> Canary, or Dev <-> Stable? Still, it doesn't explain the timeouts / unstability / flakiness. Very odd.

Pinning is an option, but since we spent so much time on this matter this week, let's defer that to another time.

OrKoN:
It's Beta for CfT and ToT for Chromium

PR #690, opened 2023-05-08
fix: don't throw error when encountering redirects

Lightning00Blade:
I have almost ready PR for Puppeteer to start using the network module. The problem is that the Mapper throws error if redirects are involved. That brakes the whole Mapper in unrecoverable way. Disabling the `throws` do not fix the problems with redirects but allows the mapper to continue functioning for the next tests.

Note: 
Redirects have the same RequestId though the chain - added TODO in place with info.


sadym-chromium (approved):
Makes sense. Thanks!

Issue #693, opened 2023-05-08
Use Chrome for Testing in CI instead of Chromium


thiagowfx:
This was originally landed on https://github.com/GoogleChromeLabs/chromium-bidi/pull/683 but had to be reverted on https://github.com/GoogleChromeLabs/chromium-bidi/pull/683 due to flakiness.

## Tracking

- [x] #697
- [x] @OrKoN's suggestion: pin browser version instead of `latest`

OrKoN:
This should be closed by https://github.com/GoogleChromeLabs/chromium-bidi/pull/703

PR #697, opened 2023-05-09
Make "Install Chromium" CI step failable

thiagowfx:
Bug: #693

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -27,7 +27,10 @@ jobs:
       - name: Install and build npm dependencies
         run: npm ci
       - name: Install Chromium
-        run: echo "chromium_binary=$(npx @puppeteer/browsers install chromium@latest | cut -f 2- -d' ')" >> $GITHUB_ENV
+        shell: bash # force -eo pipefail
+        run: |
+          chromium_binary="$(npx @puppeteer/browsers install chromium@latest | cut -f 2- -d' ')"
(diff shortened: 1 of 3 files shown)


thiagowfx:
`chromium@115` does not work. We need rev number.

OrKoN:
@thiagowfx that's why you should use CfT ;-)

.github/workflows/e2e.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       - name: Install and build npm dependencies
         run: npm ci
       - name: Install Chromium
-        run: echo "chromium_binary=$(npx @puppeteer/browsers install chromium@latest | cut -f 2- -d' ')" >> $GITHUB_ENV
+        run: echo "chromium_binary=$(npx @puppeteer/browsers install chromium@1141358 | cut -f 2- -d' ')" >> $GITHUB_ENV
```
OrKoN, line 30: let's add a file (e.g., `.browser`) to the root of the repository containing `chromium@1141358` and read that file in CI to get the pinned browser@version?

sadym-chromium (approved):
LGTM. Alex's suggestion would make it a bit smoother, but can be done later.

thiagowfx:
Great: I made a change to make `install chromium` fail. Previously, it was always succeeding, because of the pipe. `-o pipefail` will make it fail now.

image

thiagowfx:
Let me merge just one of the commits for now.

PR #700, opened 2023-05-10
miscellaneous ESLint and CI changes

thiagowfx:
c.f. individual commits for the full changes

### DIFF
--- a/.github/workflows/wpt.yml
+++ b/.github/workflows/wpt.yml
@@ -87,21 +87,18 @@ jobs:
           UPDATE_EXPECTATIONS: true
           WPT_REPORT: out/wptreport.json
       - name: Generate HTML test report
-        # Force run task even if the previous tasks failed.
         if: ${{ always() }}
         run: >
           node test-report/htmlWptReport.mjs
           out/wptreport.json
           out/wptreport.html
       - name: Upload expectations
-        # Force run task even if the previous tasks failed.
         if: ${{ always() }}
         uses: actions/upload-artifact@v3
         with:
           name: ${{ matrix.kind }}-${{ matrix.head }}-wpt-metadata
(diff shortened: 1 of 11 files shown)


OrKoN (approved):

package.json
```diff
@@ -44,7 +44,12 @@
       ]
     },
     "eslint": {
-      "command": "eslint --cache --ext .js,.ts --fix ."
+      "command": "eslint --cache --ext .js,.ts --fix .",
+      "files": [
+        ".eslintignore",
+        ".eslintrc.js",
+        "src/**/*.ts"
```
Lightning00Blade, line 50: Note: because of how `wireit` in watch mode this will run 2 times. 
I remember need to explicitly set this for VSCode to fix EsLint errors.
```
  "editor.codeActionsOnSave": {
    "source.fixAll.eslint": true
  },
  ```

Comment (thiagowfx) [reply package.json:50]: Thanks! This should be non-intrusive, though. I use `--watch`, but Maksim does not. If one does not use `--watch`, this is a no-op. We also have that snippet in our vscode config: https://github.com/GoogleChromeLabs/chromium-bidi/blob/a09dc1c3132a074710632decb05f6cbc8cac29b7/.vscode/settings.json#L6

File: .github/workflows/wpt.yml [hunk 2]
```diff
@@ -80,6 +84,13 @@ jobs:
           CHROMEDRIVER: true
           UPDATE_EXPECTATIONS: true
           WPT_REPORT: out/wptreport.json
+      - name: Commit WPT expectations
```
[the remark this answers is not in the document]
thiagowfx: We definitely only want it on PRs, main should never be touched by this. Good point. Will do.

Lightning00Blade (approved):

thiagowfx:
I am moving the auto-commit stuff to a separate PR in case we need to easily revert it.

PR #703, opened 2023-05-11
chore: pin Chrome

OrKoN:


install-browser.mjs
```diff
@@ -0,0 +1,44 @@
+/**
```
thiagowfx, line 1: Not a fan of adding this file here...shouldn't `@puppeteer/browsers` be responsible for this logic instead?

.browser
```diff
@@ -0,0 +1 @@
+chrome@115.0.5763.0
```
thiagowfx, line 1: Not a blocker:

Instead of creating an arbitrary `.browser`, let's stick to a widely used convention like `.envrc` or `.env` (or similar):

```
$ cat ~/.envrc:
BROWSER=chrome@115.0.5763.0
```

install-browser.mjs
OrKoN: It is responsible but requires an integration with GitHub Actions. Therefore a cross-platform Node script is preferrable to sh/bash.

.browser
OrKoN: this is not really an env variable so I am not sure it needs to be in an env file.

install-browser.mjs
```diff
@@ -0,0 +1,44 @@
...
+ * Copyright 2021 Google LLC.
```
thiagowfx, line 2: 2023

install-browser.mjs
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+import {install, computeExecutablePath} from '@puppeteer/browsers';
+import actions from '@actions/core';
+import fs from 'fs/promises';
+
+try {
+  const version = (await fs.readFile('.browser', 'utf-8')).trim();
+  const cacheDir = process.argv[2] || process.cwd();
+  const browser = version.split('@')[0];
```
thiagowfx, line 25:
```suggestion
  // Format: {chrome | chromium}
  const browser = version.split('@')[0];
```

install-browser.mjs
```diff
@@ -0,0 +1,44 @@
...
+  const buildId = version.split('@')[1];
```
thiagowfx, line 26: Could we rename:

- `version` -> `browserSpec` (or `browserAndVersion`, or similar)
- `buildId` -> `version`

install-browser.mjs
```diff
@@ -0,0 +1,44 @@
...
+    'executablePath',
+    computeExecutablePath({
+      cacheDir,
+      browser,
+      buildId,
+    })
+  );
+  console.log(result.path);
+} catch (err) {
+  actions.setFailed('Failed to download the browser: ' + err.message);
```
thiagowfx, line 40: Use `` (this wasn't detected by pre-commit because it is only configured to run eslint for `.ts`, not `.mjs`, but we should change that later

install-browser.mjs
thiagowfx: Still, it would be more ideal to e.g. have @puppeteer/browsers provide an action we can `uses:` here, with custom inputs (browser + version + cache dir).

I am OK to land this as an intermediate step, but long-term we should get rid of this file _in this repo_, IMHO.

install-browser.mjs
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+
+import {install, computeExecutablePath} from '@puppeteer/browsers';
+import actions from '@actions/core';
+import fs from 'fs/promises';
+
+try {
+  const version = (await fs.readFile('.browser', 'utf-8')).trim();
+  const cacheDir = process.argv[2] || process.cwd();
```
thiagowfx, line 24: Could you document this somewhere?

Maybe add a `@fileoverview` block at the top (after the copyright) with something like:

`Usage: node <script> <path/to/cache/dir>` + 1 or 2-sentence description of the purpose of this file

.github/workflows/wpt.yml
```diff
@@ -49,11 +49,14 @@ jobs:
...
         working-directory: wpt
-      - name: Install Chromium
-        shell: bash # force -eo pipefail
-        run: |
-          chromium_binary="$(npx @puppeteer/browsers install chromium@latest | cut -f 2- -d' ')"
-          echo "chromium_binary=$chromium_binary" >> $GITHUB_ENV
+      - name: Setup cache for browser binaries
+        uses: actions/cache@v3
+        with:
+          path: ~/.cache/chromium-bidi
```
thiagowfx, line 55: If the name of the action is browser agnostic, let's make the cache path browser agnostic as well.

`~/.cache/browsers` or similar

.github/workflows/generate-wpt-report.yml
```diff
@@ -43,16 +43,19 @@ jobs:
...
-          chromium_binary="$(npx @puppeteer/browsers install chromium@latest | cut -f 2- -d' ')"
-          echo "chromium_binary=$chromium_binary" >> $GITHUB_ENV
+      - name: Setup cache for browser binaries
+        uses: actions/cache@v3
+        with:
+          path: ~/.cache/chromium-bidi
+          key: ${{ runner.os }}-browsers-${{ hashFiles('.browser') }}) }}
+      - name: Install pinned browser
+        id: browser
+        run: node install-browser.mjs
```
thiagowfx, line 53: add cache path?

install-browser.mjs
OrKoN: buildId is not a version though. Depending on the browser it could be anything

.github/workflows/wpt.yml
OrKoN: `chromium-bidi` is the name of the repository though to avoid conflicts with other cashes.

install-browser.mjs
OrKoN: Sounds good to me!

install-browser.mjs
OrKoN: I didn't get the idea behind the suggestion, could you please elaborate?

sadym-chromium (approved):

sadym-chromium:
thanks!

install-browser.mjs
thiagowfx: The idea was to document the browser spec here. That said, with `@fileoverview`, this becomes unnecessary.

I'd add a snippet there in `@fileoverview` to document the format (`<browser>@<version>`) though.

install-browser.mjs
```diff
@@ -0,0 +1,52 @@
...
+  });
+  const executablePath = computeExecutablePath({
+    cacheDir,
+    browser,
+    buildId,
+  });
+  actions.setOutput('executablePath', executablePath);
+  console.log(executablePath);
+} catch (err) {
+  actions.setFailed('Failed to download the browser: ' + err.message);
```
thiagowfx, line 51:
```suggestion
  actions.setFailed(`Failed to download the browser: ${err.message}`);
```

thiagowfx (approved):

thiagowfx:
E2E failure seems unrelated: https://github.com/GoogleChromeLabs/chromium-bidi/actions/runs/4947469165/jobs/8846880913?pr=703

install-browser.mjs
OrKoN: Added a link to the source of the format.

OrKoN:
Are those WPT failures related?

PR #705, opened 2023-05-11
CI: automatically create commits with updated WPT expectations

thiagowfx:
Only commit WPT expectations if it is in a PR.

Docs:

- https://github.com/stefanzweifel/git-auto-commit-action
- https://stackoverflow.com/questions/55657835/is-there-a-way-to-push-changes-with-a-github-action


thiagowfx:
https://github.com/GoogleChromeLabs/chromium-bidi/pull/747 ==> Verified it does not commit when the branch name does not match.

thiagowfx:
I can't seem to make this work only for named branches that contain "auto-commit". Closing.

PR #706, opened 2023-05-11
chore: auto-update Chrome

OrKoN:


.github/workflows/update-browser-pin.yml
```diff
@@ -0,0 +1,32 @@
+# This workflow will update the version in .browser and open a PR
+
+name: 'Update pinned Chrome'
+
+on:
+  schedule:
+    # Run hourly at xx:05.
```
thiagowfx, line 7: Isn't `hourly` too much?

`weekly` (to match with dependabot that updates WPT) is better, IMHO.

.github/workflows/update-browser-pin.yml
```diff
@@ -0,0 +1,32 @@
...
+    - cron: '05 * * * *'
+  push:
+    branches:
+      - main
+  workflow_dispatch:
```
thiagowfx, line 12: What is `workflow_dispatch:` for?

.github/workflows/update-browser-pin.yml
```diff
@@ -0,0 +1,32 @@
...
+
+jobs:
+  unit:
```
thiagowfx, line 15: s/unit/update

.github/workflows/update-browser-pin.yml
```diff
@@ -0,0 +1,32 @@
...
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
+          cache: npm
+      - uses: google/wireit@setup-github-actions-caching/v1
+      - name: Install and build npm dependencies
+        run: npm ci
+      - name: Download the latest version
+        run: |
+          browser_spec="$(npx @puppeteer/browsers install chrome@canary --path /tmp/browsers | cut -f 1 -d' ')"
```
thiagowfx, line 29: Can we omit `--path` to just use the cwd?

.github/workflows/update-browser-pin.yml
```diff
@@ -0,0 +1,32 @@
...
+          echo "$browser_spec\n" > .browser
```
thiagowfx, line 30: `echo` emits `\n` by default. This will yield:

```
1 chrome@123
2
```

I believe you want only

```
1 chrome@123
```

...in which case, omit the `\n`.

.github/workflows/update-browser-pin.yml
thiagowfx, line 29: Not an AI for this repo: It would be nice if `browsers` could just do a "HEAD" request or similar to merely query the browser version, without downloading it.

Here we don't really need to download the browser.

It's a small optimization, and not in this repo anyway, so just FYI.

.github/workflows/update-browser-pin.yml
```diff
@@ -0,0 +1,32 @@
+# This workflow will update the version in .browser and open a PR
```
thiagowfx, line 1: nit: Can we rename name the file to `update-browser-version.yml`?

.github/workflows/update-browser-pin.yml
sadym-chromium: The more frequent the better, as it would allow us knowing about breaking changes early

OrKoN: The PR will only be opened/updated if there are changes. The frequency matches the CfT dashboard frequency.

.github/workflows/update-browser-pin.yml
OrKoN: it is for manually triggering the job

.github/workflows/update-browser-pin.yml
OrKoN: untracked files will be published in the PR and we don't want to commit the binaries.

PR #710, opened 2023-05-11
Fix more WPT preload script tests

thiagowfx:
Bug: #293

### DIFF
--- a/src/bidiMapper/domains/context/browsingContextProcessor.ts
+++ b/src/bidiMapper/domains/context/browsingContextProcessor.ts
@@ -155,7 +155,8 @@ export class BrowsingContextProcessor {
       sessionId,
       this.#realmStorage,
       this.#eventManager,
-      this.#preloadScriptStorage
+      this.#preloadScriptStorage,
+      this.#browsingContextStorage
     );
 
     if (maybeContext) {
--- a/src/bidiMapper/domains/context/browsingContextProcessor.ts
+++ b/src/bidiMapper/domains/context/browsingContextProcessor.ts
@@ -156,7 +156,8 @@ export class BrowsingContextProcessor {
       this.#realmStorage,
       this.#eventManager,
       this.#preloadScriptStorage,
-      this.#browsingContextStorage
+      this.#browsingContextStorage,
+      this.#logger
     );
 
     if (maybeContext) {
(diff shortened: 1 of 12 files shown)


tests/test_browsing_context.py
```diff
@@ -303,21 +303,33 @@ async def test_browsingContext_create_eventContextCreatedEmitted(
...
-        await read_JSON_message(websocket)
+        await read_JSON_message(websocket),
+        await read_JSON_message(websocket),
+        await read_JSON_message(websocket),
     ]
 
     messages.sort(key=lambda x: x["method"] if "method" in x else "")
     [context_created_event, dom_content_loaded_event, load_event] = messages
 
+    load_result = await read_JSON_message(websocket)
```
sadym-chromium, line 314: Are you sure this extra message should be here? What is it?

thiagowfx:
@sadym-chromium I am blocked here, but did some investigation yesterday and wanted to share my findings.

The main issue is that two `load` events are being sent. We only want one.

I tried to suppress the second load event in several ways - added a boolean to control it, added a deferred to control when preload scripts are added, tried to do a shallow navigation. All my attempts failed so far.

Another route I took was to do a single navigation, but then the preload scripts tests do not pass.

We should try to debug together next time.

sadym-chromium:
> Although this approach fixes the scenarios where new targets are created via BiDi commands, there are other ways to create targets that also need to be handled. For example, the following link creates a new target in a new tab: `<a href="#" target="_blank">new tab</a>`. The approach of splitting target creation into two steps would not allow handling this use case.

So the solution is relatively simple:
* In `loadPreloadScripts`, you need not only call `appendCdpPreloadScript`, but actually evaluate each of them BEFORE the `Runtime.runIfWaitingForDebugger` is called.

thiagowfx:
Thanks for looking into it! Let me patch https://github.com/GoogleChromeLabs/chromium-bidi/pull/743 locally and try it out.

thiagowfx:
@sadym: E2E to add:

```
# Browsing contexts:
# 1. Existing top-level.
# 2. [same-origin, cross-origin] iframe added to existingExisting top-level 
#    BrowsingContext.
# 3. New top level created via [browsingContext.create, target=_blank]
# 4. New [same-origin, cross-origin] as a part of a new top-level
#    BrowsingContext created via [browsingContext.create, target=_blank].
we should carefully check what of this is covered and what is not, and add the missing parts
```

1) `test_addPreloadScript_setGlobalVariable`
2) `test_addPreloadScript_loadedInNewIframes*`

sadym-chromium (approved):
LGTM. Nice amount of new passing WPT!

PR #713, opened 2023-05-11
chore: update pinning + browsers version

OrKoN:


.github/workflows/update-browser-version.yml
```diff
@@ -30,3 +27,8 @@ jobs:
           echo "$browser_spec" > .browser
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v5
+        with:
+          token: ${{secrets.BROWSER_AUTOMATION_BOT_TOKEN}}
```
thiagowfx, line 31:
```suggestion
          token: ${{ secrets.BROWSER_AUTOMATION_BOT_TOKEN }}
```

(for consistency)

.github/workflows/update-browser-version.yml
```diff
@@ -30,3 +27,8 @@ jobs:
           echo "$browser_spec" > .browser
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v5
+        with:
```
thiagowfx, line 30: Set `delete-branch: true`?

> Delete the branch when closing pull requests, and when undeleted after merging. 

Default is false

thiagowfx (approved):

.github/workflows/update-browser-version.yml
OrKoN: Should be okay since merging a PR deletes the branch automatically. Let's iterate on this detail later.

PR #720, opened 2023-05-11
chore: add concurrency options to workflows

OrKoN:
so that GitHub actions can cancel the runs for outdated commits.
See https://docs.github.com/en/actions/using-jobs/using-concurrency

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -7,6 +7,10 @@ on:
   push:
     branches: 'main'
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 jobs:
   e2e:
     strategy:
--- a/.github/workflows/generate-wpt-report.yml
+++ b/.github/workflows/generate-wpt-report.yml
@@ -13,6 +13,10 @@ permissions:
   pages: write
   id-token: write
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 jobs:
   wpt:
     environment:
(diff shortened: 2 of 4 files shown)


.github/workflows/e2e.yml
```diff
@@ -7,6 +7,10 @@ on:
   push:
     branches: 'main'
 
+concurrency:
+  group: e2e-${{ github.head_ref || github.run_id }}
```
thiagowfx, line 11: Can you use `${{ github.workflow }}` instead of hard-coding `e2e` here? Same to all other workflows.

Please double-check if `github.workflow` evals to "e2e" (or to "e2e.yml") though.

thiagowfx (approved):

sadym-chromium:
This is not always a wanted behaviour. Specifically, the long-lasting WPT tasks produces WPT meta data, which can be used to update WPT expectations. Cancelling previous workflows would increase already long waiting time before having access to WPT meta.

OrKoN:
@sadym-chromium could you explain how WPT metadata from outdated commits will be used?

sadym-chromium:
Those meta from the old commits is better then no meta at all, especially when getting results takes 20 minutes. In quite a lot of cases the new commit is a rebase, some merge or addressing some PR comment. In such a case I still want to get the WPT feedback ASAP to start investigating failures rather then waiting for another 20 minutes to get the fresh result.

sadym-chromium:
Another issue is it looks like the workflow cancels both new and old workflows

OrKoN:
@sadym-chromium it should only cancel the old ones. Regarding WPT, we could exclude WPT jobs but note that GitHub actions pool becomes exhausted when old jobs are not canceled, leading to longer wait times. In any case, if that does not work for you, please revert.

OrKoN:
Also consider triggering special jobs for getting WPT metadata by adding a label on a PR so that those jobs are persisted only when needed.

PR #725, opened 2023-05-16
chore: EventEmitter should return type this

Lightning00Blade:
Similar to how the NodeJS EventEmitter is type we should expect to return this, this make it easier to extend.


OrKoN (approved):

thiagowfx (approved):
There's a linter rule that should have done this, not sure why it didn't catch it.

Thanks!

PR #726, opened 2023-05-16
Deep serialization fix

sadym-chromium:
* Update schema + types
* Fix CDP parameter passing
* Added e2e

### DIFF
--- a/src/bidiMapper/domains/script/scriptEvaluator.ts
+++ b/src/bidiMapper/domains/script/scriptEvaluator.ts
@@ -94,7 +94,7 @@ export class ScriptEvaluator {
     resultOwnership: Script.ResultOwnership,
     serializationOptions: Script.SerializationOptions
   ): Promise<Script.ScriptResult> {
-    if (![0, undefined].includes(serializationOptions?.maxDomDepth))
+    if (![0, undefined, null].includes(serializationOptions.maxDomDepth))
       throw new Error('serializationOptions.maxDomDepth!=0 is not supported');
 
     const cdpEvaluateResult = await realm.cdpClient.sendCommand(
@@ -105,9 +105,10 @@ export class ScriptEvaluator {
         awaitPromise,
         serializationOptions: {
           serialization: 'deep',
-          ...(serializationOptions?.maxObjectDepth
+          ...(serializationOptions.maxObjectDepth === undefined ||
+          serializationOptions.maxObjectDepth === null
             ? {}
-            : {maxDepth: serializationOptions?.maxObjectDepth}),
+            : {maxDepth: serializationOptions.maxObjectDepth}),
         },
       }
     );
@@ -142,7 +143,7 @@ export class ScriptEvaluator {
     resultOwnership: Script.ResultOwnership,
     serializationOptions: Script.SerializationOptions
   ): Promise<Script.ScriptResult> {
-    if (![0, undefined].includes(serializationOptions?.maxDomDepth))
+    if (![0, null, undefined].includes(serializationOptions.maxDomDepth))
       throw new Error('serializationOptions.maxDomDepth!=0 is not supported');
 
     const callFunctionAndSerializeScript = `(...args)=>{ return _callFunction((\n${functionDeclaration}\n), args);
@@ -173,9 +174,10 @@ export class ScriptEvaluator {
           arguments: thisAndArgumentsList, // this, arguments.
           serializationOptions: {
             serialization: 'deep',
-            ...(serializationOptions?.maxObjectDepth
+            ...(serializationOptions.maxObjectDepth === undefined ||
+            serializationOptions.maxObjectDepth === null
               ? {}
-              : {maxDepth: serializationOptions?.maxObjectDepth}),
+              : {maxDepth: serializationOptions.maxObjectDepth}),
           },
           executionContextId: realm.executionContextId,
         }
--- a/src/protocol-parser/protocol-parser.ts
+++ b/src/protocol-parser/protocol-parser.ts
@@ -270,24 +270,27 @@ export namespace Script {
   //   ?maxObjectDepth: (js-uint / null) .default null,
   //   ?includeShadowTree: ("none" / "open" / "all") .default "none",
   // }
-  const SerializationOptions = zod.object({
-    maxDomDepth: zod.number().int().min(0).optional(),
-    maxObjectDepth: zod.number().int().min(0).max(MAX_INT).optional(),
+  const SerializationOptionsSchema = zod.object({
+    maxDomDepth: zod.union([zod.null(), zod.number().int().min(0)]).optional(),
+    maxObjectDepth: zod
+      .union([zod.null(), zod.number().int().min(0).max(MAX_INT)])
+      .optional(),
     includeShadowTree: zod.enum(['none', 'open', 'all']).optional(),
   });
 
-  // ScriptEvaluateParameters = {
-  //   expression: text;
-  //   target: Target;
-  //   ?awaitPromise: bool;
-  //   ?resultOwnership: ResultOwnership;
+  // script.EvaluateParameters = {
+  //   expression: text,
+  //   target: script.Target,
+  //   awaitPromise: bool,
+  //   ?resultOwnership: script.ResultOwnership,
+  //   ?serializationOptions: script.SerializationOptions,
   // }
   const EvaluateParametersSchema = zod.object({
     expression: zod.string(),
     awaitPromise: zod.boolean(),
     target: TargetSchema,
     resultOwnership: ResultOwnershipSchema.optional(),
-    // serializationOptions: SerializationOptions.optional(),
+    serializationOptions: SerializationOptionsSchema.optional(),
   });
 
   export function parseEvaluateParams(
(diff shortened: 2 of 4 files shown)


tests/test_serialization_options.py
```diff
@@ -0,0 +1,91 @@
...
+                "awaitPromise": False,
+                "target": {
+                    "context": context_id
+                },
+                "serializationOptions": {
+                    "maxObjectDepth": max_object_depth
+                }
+            }
+        })
+    assert result["result"] == expected_result
```
thiagowfx, line 91: From a unit test perspective, it would be cleaner to split `callFunction` into separate tests, IMHO.

There are two ways to do this: You could either introduce a boolean param + add an `if` (to execute either `callFunction` or `evaluate`), or you could split the method into two (`test_serializationOptions_script_evaluate` and `test_serializationOptions_script_call_function).

The former would reduce code duplication, but the latter would be more readable.

I'd suggest to adopt an approach like in the [`pytest` docs](https://docs.pytest.org/en/7.1.x/example/parametrize.html#different-options-for-test-ids), with a `TESTDATA = ` definition at the top:

```
testdata = [
    (datetime(2001, 12, 12), datetime(2001, 12, 11), timedelta(1)),
    (datetime(2001, 12, 11), datetime(2001, 12, 12), timedelta(-1)),
]


@pytest.mark.parametrize("a,b,expected", testdata)
def test_timedistance_v0(a, b, expected):
    diff = a - b
    assert diff == expected


@pytest.mark.parametrize("a,b,expected", testdata, ids=["forward", "backward"])
def test_timedistance_v1(a, b, expected):
    diff = a - b
    assert diff == expected
```

Then we can have the two separate methods for each script calling style and they would be quite readable.

src/protocol-parser/protocol-parser.ts
```diff
@@ -270,24 +270,27 @@ export namespace Script {
   //   ?maxObjectDepth: (js-uint / null) .default null,
   //   ?includeShadowTree: ("none" / "open" / "all") .default "none",
   // }
-  const SerializationOptions = zod.object({
-    maxDomDepth: zod.number().int().min(0).optional(),
-    maxObjectDepth: zod.number().int().min(0).max(MAX_INT).optional(),
+  const SerializationOptionsSchema = zod.object({
+    maxDomDepth: zod.union([zod.null(), zod.number().int().min(0)]).optional(),
```
thiagowfx, line 274: How about adding `.default(0)` here? https://w3c.github.io/webdriver-bidi/#type-script-SerializationOptions

src/protocol-parser/protocol-parser.ts
```diff
@@ -270,24 +270,27 @@ export namespace Script {
...
+    maxObjectDepth: zod
+      .union([zod.null(), zod.number().int().min(0).max(MAX_INT)])
+      .optional(),
```
thiagowfx, line 277: How about adding `.default(null)` here? https://w3c.github.io/webdriver-bidi/#type-script-SerializationOptions

src/protocol-parser/protocol-parser.ts
```diff
@@ -270,24 +270,27 @@ export namespace Script {
...
     includeShadowTree: zod.enum(['none', 'open', 'all']).optional(),
```
thiagowfx, line 278: Unrelated to this PR, but if you add `.default()` (which I do recommend!), ensure to also add it here, set to `'none'`.

src/bidiMapper/domains/script/scriptEvaluator.ts
```diff
@@ -142,7 +143,7 @@ export class ScriptEvaluator {
     resultOwnership: Script.ResultOwnership,
     serializationOptions: Script.SerializationOptions
   ): Promise<Script.ScriptResult> {
-    if (![0, undefined].includes(serializationOptions?.maxDomDepth))
+    if (![0, null, undefined].includes(serializationOptions.maxDomDepth))
       throw new Error('serializationOptions.maxDomDepth!=0 is not supported');
```
thiagowfx, line 147:
```suggestion
      throw new Error('serializationOptions.maxDomDepth other than 0 or null is not supported');
```

src/bidiMapper/domains/script/scriptEvaluator.ts
```diff
@@ -94,7 +94,7 @@ export class ScriptEvaluator {
     resultOwnership: Script.ResultOwnership,
     serializationOptions: Script.SerializationOptions
   ): Promise<Script.ScriptResult> {
-    if (![0, undefined].includes(serializationOptions?.maxDomDepth))
+    if (![0, undefined, null].includes(serializationOptions.maxDomDepth))
       throw new Error('serializationOptions.maxDomDepth!=0 is not supported');
```
thiagowfx, line 98:
```suggestion
      throw new Error('serializationOptions.maxDomDepth other than 0 or null is not supported');
```

src/bidiMapper/domains/script/scriptEvaluator.ts
```diff
@@ -94,7 +94,7 @@ export class ScriptEvaluator {
     resultOwnership: Script.ResultOwnership,
     serializationOptions: Script.SerializationOptions
   ): Promise<Script.ScriptResult> {
-    if (![0, undefined].includes(serializationOptions?.maxDomDepth))
+    if (![0, undefined, null].includes(serializationOptions.maxDomDepth))
```
thiagowfx, line 97:
```suggestion
    if (![0, null, undefined].includes(serializationOptions.maxDomDepth))
```

Let's use the same order as below for consistency (no particular preference about the order, you could also change it the other way if preferred)

tests/test_serialization_options.py
```diff
@@ -0,0 +1,91 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pytest
+from test_helpers import execute_command
+
+
+@pytest.mark.parametrize("eval, max_object_depth, expected_result", [
```
thiagowfx, line 20: What do you think about adding an `ids=` param here to make the test function name more readable, as WPT does? https://github.com/web-platform-tests/wpt/blob/6ba9288e50a8881a791b3d2f7814df799846658e/webdriver/tests/print/orientation.py#L17

Docs: https://docs.pytest.org/en/7.1.x/example/parametrize.html#different-options-for-test-ids

> If you leave it up to pytest to generate ids, you will see that most of them are pretty good, but others feel somewhat random. The reason being that non-primitive types such as dict, list, tuple or instances of classes are non trivial to translate into ids.

(source: https://hackebrot.github.io/pytest-tricks/param_id_func/)

src/protocol-parser/protocol-parser.ts
sadym-chromium: If we want to provide a BiDi-like API to Puppeteer, we have to manage defaults later on in `process_script_evaluate` and `process_script_callFunction`. Handling defaults in 2 places seems odd.

src/protocol-parser/protocol-parser.ts
sadym-chromium: https://github.com/GoogleChromeLabs/chromium-bidi/pull/726#discussion_r1195130999

tests/test_serialization_options.py
```diff
@@ -0,0 +1,91 @@
+# Copyright 2021 Google LLC.
```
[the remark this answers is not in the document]
sadym-chromium: done

tests/test_serialization_options.py
```diff
@@ -0,0 +1,91 @@
...
+                    "value": "baz"
+                }]]
+            }
+        ]],
+    }),
+])
+@pytest.mark.asyncio
+async def test_serializationOptions(websocket, context_id, eval,
+                                    max_object_depth, expected_result):
+    result = await execute_command(
```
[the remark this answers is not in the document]
sadym-chromium: done

src/protocol-parser/protocol-parser.ts
thiagowfx: OK. Then maybe we should change `browsingContext.print` to match and stop using `.default` here. But this is out of scope. Feel free to keep it as is then.

tests/test_serialization_options.py
sadym-chromium: good idea. Done

tests/test_serialization_options.py
sadym-chromium: done

thiagowfx (approved):
LGTM

tests/test_serialization_options.py
```diff
@@ -0,0 +1,105 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pytest
+from test_helpers import execute_command
+
+test_data = {
```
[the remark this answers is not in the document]
thiagowfx: I'll resolve this to unblock you, but consider doing it as a follow-up.

PR #728, opened 2023-05-16
fix: layering issue with Puppeteer

Lightning00Blade:
A few days ago we introduce the error handling when you close the page while Targets are being initialize.
We did this by using `error.message.includes('Target closed')`. Two problems have been detected since:
1) No only `Target closed` but `Session closed` error may be thrown when closing (both we want to suppress).
2) The `Target closed`  and  `Session closed` are strings coming from Puppeteer's implementation which is a violation as Mapper should be independent from Puppeteer and handle this error on its own.

On the CDPClient (as we inject our own from Puppeteer) we have a method that can tell us if the Error is cause because of Closing events.


src/bidiMapper/CdpConnection.ts
```diff
@@ -33,6 +33,7 @@ export interface CdpConnection {
 }
 
 export interface CdpClient extends EventEmitter<CdpEvents> {
+  isCloseError(err: any): boolean;
```
thiagowfx, line 36: prefer `unknown` over `any`: https://google.github.io/styleguide/tsguide.html#any

thiagowfx, line 36: Please document this method

src/cdp/cdpConnection.ts
```diff
@@ -27,6 +27,8 @@ interface CdpCallbacks {
   error: Error;
 }
 
+export class CloseError extends Error {}
```
thiagowfx, line 30: Please document this class

thiagowfx (approved):

PR #729, opened 2023-05-16
`browsingContext.create`: become spec compliant

thiagowfx:
Spec: https://w3c.github.io/webdriver-bidi/#command-browsingContext-create


thiagowfx:
The WPT tests are incorrect and violate the spec. I'll send a PR upstream.

thiagowfx:
Sent https://github.com/web-platform-tests/wpt/pull/40034

src/bidiMapper/domains/context/browsingContextProcessor.ts
```diff
@@ -230,10 +231,23 @@ export class BrowsingContextProcessor {
...
-      newWindow: params.type === 'window',
-    });
+    let result: Protocol.Target.CreateTargetResponse;
+
+    switch (params.type) {
+      case 'tab':
+        result = await browserCdpClient.sendCommand('Target.createTarget', {
+          url: 'about:blank',
+          newWindow: false,
+          browserContextId: referenceContext?.id ?? undefined,
```
sadym-chromium, line 241: CDP `browserContextId` is not the same as BiDi's `browsingContext`.

sadym-chromium (commented):
LGTM, but let's merge the WPT PR first

src/bidiMapper/domains/context/browsingContextProcessor.ts
thiagowfx: Should I revert this snippet? The spec says we should differentiate `tab` and `window` which we don't currently seem to do, but we should try to address it at another time.

thiagowfx:
```
// TODO: Pass correct browserContextId.
// browserContextId: referenceContext?.id ?? undefined,
```

Would this comment be appropriate?

sadym-chromium: In order to implement it properly, we need to:
1. Add `cdpBrowserContextId` to `CdpTarget`.
2. Use it: `browserContextId: referenceContext?.target.cdpBrowserContextId ?? undefined`

thiagowfx: Given your previous comment:

> Chrome doesn't support opening tab next to another one, so the parameter should be ignored. No need it TODO.

Omitting this for now.

sadym-chromium (approved):

PR #731, opened 2023-05-16
script.disown: remove obsolete WPT expectations

thiagowfx:


thiagowfx:
It turns out this is not obsolete at all, never mind.

PR #735, opened 2023-05-17
Handle browsing context reload `wait` param

thiagowfx:
Bug: #650
CDP: https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-reload
Spec: https://w3c.github.io/webdriver-bidi/#command-browsingContext-reload

### DIFF
--- a/src/bidiMapper/domains/context/browsingContextImpl.ts
+++ b/src/bidiMapper/domains/context/browsingContextImpl.ts
@@ -47,7 +47,7 @@ export class BrowsingContextImpl {
 
   readonly #browsingContextStorage: BrowsingContextStorage;
 
-  readonly #defers = {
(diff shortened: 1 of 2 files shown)


thiagowfx:
@sadym-chromium This one is also ready.

src/bidiMapper/domains/context/browsingContextImpl.ts
```diff
@@ -47,7 +47,7 @@ export class BrowsingContextImpl {
 
   readonly #browsingContextStorage: BrowsingContextStorage;
 
-  readonly #defers = {
+  readonly #deferred = {
```
sadym-chromium, line 50: `deferreds` to underline it's plural?

PR #737, opened 2023-05-17
Merge WPT and WPT Report jobs

thiagowfx:
Make the WPT Report job dependent on the WPT job.

Docs: https://docs.github.com/en/actions/using-jobs/using-jobs-in-a-workflow


thiagowfx:
This is working as expected! Example successful run: https://github.com/GoogleChromeLabs/chromium-bidi/actions/runs/5008502015

Note that the final build will fail because I am setting it to only be deployable from the main branch.

sadym-chromium (approved):

PR #739, opened 2023-05-18
feat: implement input module

jrandolf:


sadym-chromium:
Please provide detailed description and e2e tests

sadym-chromium:
Or convert it to draft, if it's not review-ready

src/bidiMapper/domains/context/browsingContextImpl.ts
```diff
@@ -190,6 +190,14 @@ export class BrowsingContextImpl {
     return this.#parentId === null;
   }
 
+  get top(): BrowsingContextImpl {
```
thiagowfx, line 193: BrowsingContextStorage has a `findTopLevelContext` method already. Please reuse it.

jrandolf: Can there be more than one top-level context? In theory, it could be a forest, no?

jrandolf: Nevermind, my interpretation of the function was wrong.

jrandolf: Hmm, I think I'd rather keep this method and remove `findTopLevelContext`. `findTopLevelContext` is not used anywhere and it's far slower than this `BrowsingContextImpl.top`.

thiagowfx:
In order to test against WPT, remove the `__dir__` files within `input`.

thiagowfx:
Are you intending to break this PR down into smaller pieces? http://go/small-cls

jrandolf:
@thiagowfx Depends. We could separate the schema and parser changes, but for the main implementation, do we want to have partial implementation?

thiagowfx:
The splits that you did recently are a good start.

It's OK to keep the main implementation in a single PR as it's unnecessary burden for you to try to split them.

I was mostly referring to the helper/aux files and to the protocol/schema.

PR #741, opened 2023-05-19
CfT: Reduce automatic updates to weekly instead of every half an hour

thiagowfx:
Align with dependabot weekly WPT update.


thiagowfx:
Every half an hour is creating too many PRs, which is noisy and distracting. I am open to using other values and/or a slightly increased frequency than weekly if desired.

.github/workflows/update-browser-version.yml
```diff
@@ -4,8 +4,8 @@ name: 'Update the pinned browser version'
 
 on:
   schedule:
-    # Run hourly at xx:30.
-    - cron: '30 * * * *'
+    # Run every Monday at midnight and noon: https://crontab.guru/#0_0,12_*_*_1.
+    - cron: '0 0,12 * * 1'
```
sadym-chromium, line 8: Are you sure we need to run it on Monday 2 times?

thiagowfx: I put this just as a fail-safe in case the first run fails, but it's not needed.

Let me revert it then.

thiagowfx: Done.

thiagowfx:
Ops, I pushed an extra file accidentally. Let me revert it.

sadym-chromium (approved):

OrKoN:
> Every half an hour is creating too many PRs, which is noisy and distracting. I am open to using other values and/or a slightly increased frequency than weekly if desired.

note that the bot creates a single PR and keep updating it when a new version is published. You are free to land whenever you want.

PR #742, opened 2023-05-19
`browsingContext.reload`: add ignore cache E2E test

thiagowfx:
Bug: #650


tests/test_browsing_context.py
```diff
@@ -879,6 +879,104 @@ async def test_browsingContext_reload_waitComplete(websocket, context_id,
...
+            "redirectCount": 0,
+            "request": ANY_DICT,
+            "response": ANY_DICT,
+            "timestamp": ANY_TIMESTAMP,
+        },
+    }
+
+    # Assert command done.
+    response = await read_JSON_message(websocket)
+    assert response["result"] == {}
```
sadym-chromium, line 930: NIT: check command `ID` as well please

tests/test_browsing_context.py
```diff
@@ -879,6 +879,110 @@ async def test_browsingContext_reload_waitComplete(websocket, context_id,
...
+    url = "https://example.com/"
+
+    await subscribe(websocket, [
+        "network.beforeRequestSent",
+        "network.responseCompleted",
+    ])
+
+    await goto_url(websocket, context_id, url)
+
+    await send_JSON_command(
```
sadym-chromium, line 892:
```suggestion
    command_id = await send_JSON_command(
```

tests/test_browsing_context.py
```diff
@@ -879,6 +879,110 @@ async def test_browsingContext_reload_waitComplete(websocket, context_id,
...
+            "request": ANY_DICT,
+            "response": ANY_DICT,
+            "timestamp": ANY_TIMESTAMP,
+        },
+    }
+
+    # Assert command done.
+    response = await read_JSON_message(websocket)
+    assert response == {
+        "id": ANY_INT,
```
sadym-chromium, line 931:
```suggestion
        "id": command_id,
```

thiagowfx:
For now, both `true` and `false` for `ignoreCache` pass. We need something else to ensure the behavior is proper. @sadym-chromium any other ideas? Network events do not seem to be enough to differentiate them.

sadym-chromium:
what about `fromCache` flag in the `network.responseCompleted` event?

thiagowfx:
Good idea. Both tests have `fromCache: False` though.

tests/test_browsing_context.py
```diff
@@ -871,6 +871,56 @@ async def test_browsingContext_reload_waitComplete(websocket, context_id,
...
+
+    response = await read_JSON_message(websocket)
+    assert response == {
+        "method": "network.responseCompleted",
+        "params": {
+            "context": context_id,
+            "navigation": ANY_STR,
+            "redirectCount": 0,
+            "request": ANY_DICT,
+            "response": AnyExtending({"fromCache": False}),
```
thiagowfx, line 916: One would expect the correct value here is `not ignoreCache`, however it fails.

thiagowfx: Upon discussing this in internal chat, it doesn't seem to be possible to detect this condition from `network.responseCompleted`.

PR #743, opened 2023-05-19
[PoC] preload script

sadym-chromium:
Proof of concept. Upon attaching to a new target, run preload scripts on each execution context before `Runtime.runIfWaitingForDebugger`.


thiagowfx:
Thanks! Incorporated in https://github.com/GoogleChromeLabs/chromium-bidi/pull/710.

PR #744, opened 2023-05-19
eslint: add promise plug-in

thiagowfx:
And fix a couple additional eslint issues.

This is meant to catch more potential promise errors.

### DIFF
--- a/package.json
+++ b/package.json
@@ -179,6 +179,7 @@
     "eslint-plugin-import": "2.27.5",
     "eslint-plugin-mocha": "10.1.0",
     "eslint-plugin-prettier": "4.2.1",
+    "eslint-plugin-promise": "6.1.1",
     "gts": "3.1.1",
     "mocha": "10.2.0",
     "nyc": "15.1.0",
--- a/src/bidiServer/index.ts
+++ b/src/bidiServer/index.ts
@@ -37,7 +37,12 @@ import mapperReader from './mapperReader.js';
 
 const log = debug('bidiServer:log');
 
-function parseArguments() {
+function parseArguments(): {
+  port: number;
+  channel: ChromeReleaseChannel;
+  headless: string;
+  verbose: boolean;
+} {
   const parser = new argparse.ArgumentParser({
     add_help: true,
     exit_on_error: true,
@@ -54,6 +59,7 @@ function parseArguments() {
       'If set, the given installed Chrome Release Channel will be used ' +
       'instead of one pointed by Puppeteer version',
     choices: Object.values(ChromeReleaseChannel),
+    default: ChromeReleaseChannel.DEV,
   });
 
   parser.add_argument('-hl', '--headless', {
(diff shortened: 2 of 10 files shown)


src/bidiTab/mapperTabPage.ts
```diff
@@ -36,16 +36,16 @@ function findOrCreateTypeLogContainer(logType: string): HTMLElement {
     return existingContainer;
   }
 
-  const debugElement = document.getElementById('details')!;
+  const debugElement = document.getElementById('details');
```
Lightning00Blade, line 39: I would highly suggest testing if empty and throwing an error as it may become harder find the issue in the feature (especially for headless).

src/utils/deferred.ts
```diff
@@ -19,8 +19,8 @@ export class Deferred<T> implements Promise<T> {
   #isFinished = false;
   #promise: Promise<T>;
 
-  #resolve: (value: T) => void = () => {};
-  #reject: (value: unknown) => void = () => {};
+  #resolve?: (value: T) => void;
+  #reject?: (value: unknown) => void;
```
Lightning00Blade, line 23: This are always present as they are set up in the constructor. The `?` should be able to be removed.

src/bidiServer/index.ts
```diff
@@ -75,7 +75,12 @@ function parseArguments() {
   try {
     log('Launching BiDi server');
 
-    const args = parseArguments();
+    const args: {
```
Lightning00Blade, line 78: I think having it on the function as return type is a little bit cleaner, WDYT?

package.json
```diff
@@ -179,6 +179,7 @@
     "eslint-plugin-import": "2.27.5",
     "eslint-plugin-mocha": "10.1.0",
     "eslint-plugin-prettier": "4.2.1",
+    "eslint-plugin-promise": "^6.1.1",
```
Lightning00Blade, line 182: Perf: pinned dep

src/bidiServer/index.ts
thiagowfx: Agreed, done.

src/utils/deferred.ts
thiagowfx:
```
🏃 [tsc] Running command "tsc --build src/tsconfig.json --pretty"
src/utils/deferred.ts:22:3 - error TS2564: Property '#resolve' has no initializer and is not definitely assigned in the constructor.

22   #resolve: (value: T) => void;
     ~~~~~~~~


Found 1 error.

❌ [tsc] Failed with exit status 1
```

thiagowfx: It's needed

Lightning00Blade: Yeah, I think it's because they are set up inside the `new Promise` :disappointed:

Lightning00Blade (approved):

PR #748, opened 2023-05-19
e2e: add windows-latest

thiagowfx:
Bug: #361


thiagowfx:
Amazing, this is very close to working (python, node are all OS-agnostics), but the path separators and the shell are issues. Maybe next time. Was just trying to see if it was a low-hanging fruit.

PR #761, opened 2023-05-23
[EXPERIMENTAL] generate TS based on CDDL

sadym-chromium:
Experiments generating typescript interfaces based on WebDriver BiDi CDDL


thiagowfx:
Later: Use https://github.com/webdriverio/webdriverio/pull/10447/ as reference.

PR #766, opened 2023-05-23
rollup: use nodePolyfills

thiagowfx:
This is necessary to address the following warning from rollup:

(!) Missing shims for Node.js built-ins
Creating a browser bundle that depends on "crypto". You might need to include https://github.com/FredKSchott/rollup-plugin-polyfill-node

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -46,6 +46,7 @@
         "prettier": "2.8.8",
         "rimraf": "5.0.0",
         "rollup": "3.21.3",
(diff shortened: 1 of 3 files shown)


thiagowfx:
Interestingly, this fails E2E tests.

I guess we'll live with the warning then - although it would be nice to suppress it.

OrKoN:
we should not need to import crypto at all. Let's find out why that didn't work instead?

thiagowfx:
I'll keep trying in the meantime, but currently I am out of ideas.

PR #768, opened 2023-05-24
Add workflow_dispatch to WPT workflow

thiagowfx:
This would allow us to manually trigger WPT jobs on-demand.

Potential future use cases:

- manually re-run, with verbose logging
- manually re-run, but only a subset of tests
- manually re-run, and auto-commit WPT expectations

Since this seems to only take effect when it's merged in main, also add one checkbox input which we can use for testing / iteration later on.

This is similar to how chrome for testing does it: https://github.com/GoogleChromeLabs/chrome-for-testing/blob/c0c46b46178cb4aa2d2897496ae902125e8e575b/.github/workflows/build.yml#L24

Bug: #361
Docs: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch

### DIFF
--- a/.github/workflows/wpt.yml
+++ b/.github/workflows/wpt.yml
@@ -7,6 +7,20 @@ on:
   pull_request:
   push:
     branches: 'main'
+  workflow_dispatch:
+    inputs:
+      # TODO(#361): Use.
+      auto-commit:
+        description: Whether to auto-commit WPT expectations
+        default: false
+        required: false
+        type: boolean
+      # TODO(#361): Use.
+      verbose:
+        description: Whether to enable verbose logging
+        default: false
+        required: false
+        type: boolean
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}


.github/workflows/wpt.yml
```diff
@@ -7,6 +7,20 @@ on:
   pull_request:
   push:
     branches: 'main'
+  workflow_dispatch:
+    inputs:
+      # TODO(#361): Use.
```
OrKoN, line 12: what does it mean? could you include a full link to the issue and expand the comment?

thiagowfx: It's just a placeholder because these things seem to only work when they are checked into main.

These will become checkboxes in the workflow run dialog.

The intent is to use these in the next PR.

thiagowfx: In other words: it's not worth to elaborate here because this will either becoming self-documenting in the next PR, or be removed if it ends up not working.

OrKoN (approved):

PR #769, opened 2023-05-24
npm: add `wpt` target and ensure build is up-to-date when running WPT

thiagowfx:
See commit history for changes.

Bug: #361

### DIFF
--- a/runWPT.sh
+++ b/runWPT.sh
@@ -87,6 +87,9 @@ else
 fi
 
 TEST="${*: -1}"
+if [ -z "${TEST}" ]; then
+  TEST="webdriver/tests/bidi/"
+fi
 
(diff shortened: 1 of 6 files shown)


runWPT.sh
```diff
@@ -87,6 +87,9 @@ else
 fi
 
 TEST="${*: -1}"
+if [ -z "${TEST}" ]; then
```
OrKoN, line 90: let's rewrite it as a node script :)

thiagowfx: It would be better indeed, as it would allow us to run it on Windows and finally enable Windows in CI.

thiagowfx: Although it's quite odd: This works on my machine but not in CI

PR #771, opened 2023-05-24
wpt: support optional verbose logging and custom tests

thiagowfx:
Bug: #361


thiagowfx:
Again, these only take effect after checked into `main`. The following will be updated:

image

thiagowfx:
Verified everything works as expected:

- verbose
- WPT subtests

OrKoN (approved):

PR #777, opened 2023-05-25
CDP refactoring, reuse types where applicable and reduce duplication

thiagowfx:


thiagowfx:
@OrKoN I wonder if I could also get rid of the `src/bidiMapper/bidiMapper.ts` file?

src/bidiMapper/bidiMapper.ts
```diff
@@ -20,7 +20,5 @@
  * Other modules should only access exports defined in this file.
  * XXX: eslint rule for this.
  */
-export {CdpConnection, CdpClient} from './CdpConnection.js';
-export {BidiServer} from './BidiServer.js';
-export {BidiTransport} from './BidiTransport.js';
-export {EventEmitter} from '../utils/EventEmitter.js';
+export {CdpConnection} from '../cdp/cdpConnection.js';
```
thiagowfx, line 23: Can I also delete this file?

thiagowfx: By other modules within `src/bidiMapper/`, correct?

OrKoN, line 23: Let's not add a dependency on other modules for this file as the mapper only needs minimal CDP interface to work and we don't need to bring in the entire CDP client implementation (that is not used by Puppeteer)

OrKoN (changes requested):

src/bidiMapper/bidiMapper.ts
```diff
@@ -20,7 +20,5 @@
  * Other modules should only access exports defined in this file.
  * XXX: eslint rule for this.
  */
-export {CdpConnection, CdpClient} from './CdpConnection.js';
-export {BidiServer} from './BidiServer.js';
-export {BidiTransport} from './BidiTransport.js';
```
OrKoN, line 25: these exports are used by Puppeteer

thiagowfx:
I tried to add https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-restricted-paths.md

Like this (several attempts):

```
    'import/no-restricted-paths': ['error', {
      zones: [
        {target: './src/bidiMapper', from: './src/bidiMapper.ts', message: 'BiDi mapper modules should not import from the bidiMapper entrypoint module in order to avoid circular dependencies.'},
        {target: './src/bidiMapper', from: './src/bidiMapper', message: 'BiDi mapper modules should not import from the bidiMapper entrypoint module in order to avoid circular dependencies.'},
        {target: './src/bidiMapper', from: './bidiMapper', message: 'BiDi mapper modules should not import from the bidiMapper entrypoint module in order to avoid circular dependencies.'},
      ],
    }],
```

....but it does not work.

thiagowfx:
Ah, I need to install the plug-in first. Never mind.

thiagowfx:
Even

```
    'import/no-restricted-paths': ['error', {
      zones: [
        {target: './src/bidiMapper/*', from: './src/bidiMapper', message: 'BiDi mapper modules should not import from the bidiMapper entrypoint module in order to avoid circular dependencies.'},
      ],
    }],
```

does not work. Odd.

src/bidiMapper/bidiMapper.ts
thiagowfx: Should I delete `export {CdpConnection} from '../cdp/cdpConnection.js';` and instead just import it directly?

OrKoN: I believe the src/bidiMapper should declare the interface it needs without the need to import the implementation interfaces

OrKoN: I think that still imports the entire file and its dependencies. I think the duplication here is fine.

thiagowfx: @Lightning00Blade made the good point offline that if we do `import type` then this should likely not be the case. Updated again, PTAL.

thiagowfx:
@OrKoN PTAL.

src/bidiMapper/BidiServer.ts
```diff
@@ -19,11 +19,11 @@ import {EventEmitter} from '../utils/EventEmitter.js';
 import {LogType, LoggerFn} from '../utils/log.js';
 import type {Message} from '../protocol/protocol.js';
 import {ProcessingQueue} from '../utils/processingQueue.js';
+import {CdpConnection} from '../cdp/cdpConnection.js';
```
OrKoN, line 22: the CDP implementation is not used in bidiMapper (it only needs the interface). Let's not import the implementation since it won't be needed by all consumers.

thiagowfx: Done

Issue #749, opened 2023-05-21
Implement `Input` module


jrandolf-2:
Spec: https://w3c.github.io/webdriver-bidi/#module-input

## Tracking

- [x] Implementation
  - [x] Initial implementation
    - [x] https://github.com/GoogleChromeLabs/chromium-bidi/pull/784 
    - [x] https://github.com/GoogleChromeLabs/chromium-bidi/pull/782
    - [x] https://github.com/GoogleChromeLabs/chromium-bidi/pull/781 
    - [x] https://github.com/GoogleChromeLabs/chromium-bidi/pull/780
    - [x] https://github.com/GoogleChromeLabs/chromium-bidi/pull/750
      - This breaks https://github.com/web-platform-tests/wpt/blob/5db1f8e9ad325bef573d36574b36c38640d70f40/webdriver/tests/bidi/input/perform_actions/invalid.py#L164
- [x] WebDriver
  - [x] https://github.com/w3c/webdriver/issues/1457
    - [x] https://github.com/web-platform-tests/wpt/pull/40992
  - [x] https://github.com/w3c/webdriver/issues/1741 (this is a noop)
  - [ ] https://github.com/w3c/webdriver-bidi/pull/666
  - [x] https://github.com/w3c/webdriver-bidi/pull/667
- [ ] WPT
  - [x] https://github.com/web-platform-tests/wpt/pull/40106
  - [x] https://github.com/web-platform-tests/wpt/pull/40101
  - [x] https://github.com/web-platform-tests/wpt/pull/40115
  - [ ] Missing tests
      - [x] /webdriver/tests/bidi/input/perform_actions/key.py/test_key_codepoint[\u0ba8\u0bbf]
         https://github.com/web-platform-tests/wpt/pull/46019
      - [x] /webdriver/tests/bidi/input/perform_actions/key.py/test_key_codepoint[\u1100\u1161\u11a8]
         https://github.com/web-platform-tests/wpt/pull/46019
      - [x] ~/webdriver/tests/bidi/input/perform_actions/pointer_mouse_multiclick.py/test_dblclick_at_coordinates[200-False]~
      - [x] /webdriver/tests/bidi/input/perform_actions/pointer_pen.py/test_pen_pointer_properties
         1. `pointermove` should not be emitted before the pen is detected (`pointerdown`)
         2. `pointerleave` should be emitted after `pointerup` in case of the pen.
         * https://g-issues.chromium.org/issues/40539182
[...]

PR #780, opened 2023-05-26
Update protocol with Input module types

jrandolf:

### DIFF
--- a/src/protocol/protocol.ts
+++ b/src/protocol/protocol.ts
@@ -33,7 +33,8 @@ export type BiDiCommand =
   | BrowsingContext.Command
   | CDP.Command
   | Script.Command
-  | Session.Command;
+  | Session.Command
+  | Input.Command;
 
 export namespace Message {
   export type OutgoingMessage =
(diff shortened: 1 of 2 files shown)


OrKoN (approved):
LGTM thanks

src/protocol/protocol.ts
```diff
@@ -1240,3 +1248,198 @@ export namespace Session {
...
+    Mouse = 'mouse',
+    Pen = 'pen',
+    Touch = 'touch',
+  }
+
+  export type PointerParameters = {
+    /**
+     * @defaultValue `"mouse"`
+     */
+    pointerType?: PointerType;
```
thiagowfx, line 1317: It's possible to do this in `protocol-parser.ts` by using zod's `.default()` method.

src/protocol/protocol.ts
```diff
@@ -1240,3 +1248,198 @@ export namespace Session {
...
+    type: ActionType.Scroll;
+    x: number;
+    y: number;
+    deltaX: number;
+    deltaY: number;
+    duration?: number;
+    /**
+     * @defaultValue `"viewport"`
+     */
+    origin?: Origin;
```
thiagowfx, line 1387: Ditto. And same for all other `defaultValue`.

PR #781, opened 2023-05-26
Update protocol parser with Input schemas

jrandolf:

### DIFF
--- a/src/protocol-parser/protocol-parser.ts
+++ b/src/protocol-parser/protocol-parser.ts
@@ -31,6 +31,7 @@ import {
   Session as SessionTypes,
   Network as NetworkTypes,
   CommonDataTypes as CommonDataTypesTypes,
+  Input as InputTypes,
 } from '../protocol/protocol.js';
 
 const MAX_INT = 9007199254740991 as const;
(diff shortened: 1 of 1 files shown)


src/protocol-parser/protocol-parser.ts
```diff
@@ -618,3 +625,291 @@ export namespace Session {
...
+  });
+
+  // input.AngleProperties = (
+  //   ? altitudeAngle: float .default 0.0,
+  //   ? azimuthAngle: float .default 0.0,
+  // )
+  const AnglePropertiesSchema = zod.object({
+    altitudeAngle: zod
+      .number()
+      .min(0.0)
```
OrKoN, line 690: Looks like the spec does not define min/max? should we accept any angle and do mod?

jrandolf: It's defined. See step (29) of https://w3c.github.io/webdriver/#dfn-process-a-pointer-up-or-pointer-down-action

OrKoN (approved):
Don't forget that auto-merge would land it into another branch instead of main.

src/protocol-parser/protocol-parser.ts
thiagowfx, line 690: In general, prefer `nonnegative()`.

PR #783, opened 2023-05-26
Remove `findTopLevelContext` in favor of class getter

jrandolf:
Placing `findTopLevelContext` in a class getter improves semantics since it will always exist if the initial context can be obtained.

### DIFF
--- a/src/bidiMapper/domains/context/browsingContextImpl.ts
+++ b/src/bidiMapper/domains/context/browsingContextImpl.ts
@@ -190,6 +190,17 @@ export class BrowsingContextImpl {
     return this.#parentId === null;
   }
 
+  get top(): BrowsingContextImpl {
+    // eslint-disable-next-line @typescript-eslint/no-this-alias
+    let topContext: BrowsingContextImpl = this;
+    let parent = topContext.parent;
+    while (parent) {
+      topContext = parent;
+      parent = topContext.parent;
+    }
+    return topContext;
+  }
+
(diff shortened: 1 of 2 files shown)


OrKoN (approved):

src/bidiMapper/domains/context/browsingContextImpl.ts
```diff
@@ -190,6 +190,17 @@ export class BrowsingContextImpl {
     return this.#parentId === null;
   }
 
+  get top(): BrowsingContextImpl {
```
thiagowfx, line 193: The previous method could return `null` or `undefined`.

What happens here if you call `<some context>.top().top()`? The parent of a top-level context should be `null` (c.f. `isTopLevelContext()`).

thiagowfx, line 193: nit: Either rename to `get topLevelContext()` or `getTopLevelContext()`, or add JSDoc (or maybe both).

src/bidiMapper/domains/context/browsingContextStorage.ts
```diff
@@ -81,17 +81,6 @@ export class BrowsingContextStorage {
     return this.findTopLevelContextId(parentId);
   }
 
-  /** Returns the top-level context of the given context, if any. */
-  findTopLevelContext(
```
thiagowfx, line 85: Maybe it also makes sense to remove `findTopLevelContextId`? (not sure)

PR #784, opened 2023-05-26
Initial Input module implementation

jrandolf:
Notes to the reviewer:

 - Mutex.ts is copied straight from https://github.com/ChromeDevTools/devtools-frontend
 - USKeyboardLayout.ts is a smaller version of https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer-core/src/common/USKeyboardLayout.ts


src/bidiMapper/domains/context/browsingContextProcessor.ts
```diff
@@ -427,15 +431,64 @@ export class BrowsingContextProcessor {
...
-    throw new Message.UnsupportedOperationException('Not implemented yet.');
+    const context = this.#browsingContextStorage.getContext(params.context);
+    const inputState = this.#inputStateManager.get(context.top);
+
+    const actionsByTick: ActionOption[][] = [];
+    for (const action of params.actions) {
+      switch (action.type) {
+        case Input.SourceActionsType.Pointer: {
+          action.parameters ??= {pointerType: Input.PointerType.Mouse};
+          action.parameters.pointerType ??= Input.PointerType.Mouse;
```
sadym-chromium, line 445: Note: Interesting way of handling defaults

OrKoN: See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing

src/bidiMapper/domains/context/browsingContextProcessor.ts
```diff
@@ -427,15 +431,64 @@ export class BrowsingContextProcessor {
...
+        default:
+          inputState.getOrCreate(action.id, action.type);
+      }
+      const actions: ActionOption[] = [];
+      for (const item of action.actions) {
+        actions.push({
+          id: action.id,
+          action: item,
+        });
+      }
```
sadym-chromium, line 468:
```suggestion
      const actions = action.actions.map(item => ({
          id: action.id,
          action: item,
        }));
```

src/bidiMapper/domains/input/InputStateManager.ts
```diff
@@ -0,0 +1,41 @@
...
+import {BrowsingContextImpl} from '../context/browsingContextImpl.js';
+
+import {InputState} from './InputState.js';
+
+export class InputStateManager {
+  #states = new WeakMap<BrowsingContextImpl, InputState>();
+
+  get(context: BrowsingContextImpl) {
+    if (!context.isTopLevelContext()) {
+      throw new Message.UnknownErrorException('Internal error');
```
OrKoN, line 28: Can we have the message say "Provided context is not top-level"?

src/bidiMapper/domains/context/browsingContextProcessor.ts
```diff
@@ -427,15 +431,64 @@ export class BrowsingContextProcessor {
...
-    _params: Input.PerformActionsParameters
+  async process_input_performActions(
+    params: Input.PerformActionsParameters
   ): Promise<Message.EmptyResult> {
-    throw new Message.UnsupportedOperationException('Not implemented yet.');
+    const context = this.#browsingContextStorage.getContext(params.context);
+    const inputState = this.#inputStateManager.get(context.top);
+
+    const actionsByTick: ActionOption[][] = [];
+    for (const action of params.actions) {
```
sadym-chromium, line 441: Can this logic be moved to a separate method? It would increase readability.

src/bidiMapper/domains/input/InputStateManager.ts
jrandolf: This should actually be an assertion. If this errors, we are calling this wrong.

src/bidiMapper/domains/input/ActionDispatcher.ts
```diff
@@ -0,0 +1,632 @@
...
+const CALCULATE_IN_VIEW_CENTER_PT_DECL = ((i: Element) => {
+  const t = i.getClientRects()[0] as DOMRect,
+    e = Math.max(0, Math.min(t.x, t.x + t.width)),
+    n = Math.min(window.innerWidth, Math.max(t.x, t.x + t.width)),
+    h = Math.max(0, Math.min(t.y, t.y + t.height)),
+    m = Math.min(window.innerHeight, Math.max(t.y, t.y + t.height));
+  return [e + ((n - e) >> 1), h + ((m - h) >> 1)];
+}).toString();
+
+async function getElementCenter(
```
sadym-chromium, line 38: Please make it a private static method

jrandolf: What's your reason? Generally you make something static if it's related to the creation, destruction, or modification of the class. It's also done for namespacing. I don't think this fits these criteria.

OrKoN:
Do we need to update some WPT expectations to see if the WPT tests pass? @sadym-chromium

tests/test_input.py
```diff
@@ -0,0 +1,323 @@
...
+                    'value': 1
+                }]]
+            }],
+            'handle': ANY_STR
+        },
+        'realm': ANY_STR
+    } == result
+
+
+# TODO(jrandolf): Fix scroll
```
OrKoN, line 278: what's with the scroll?

jrandolf: Scroll doesn't work by itself without some mouse input. I've updated the test to showcase this.

jrandolf:
@OrKoN Yea, we should update the WPT tests. I've added appended the headless tests, but should we do the others as well?

OrKoN (approved):
Code LGTM but I will leave the review of WPT test expectations/e2e to Maksim since I am not very familiar with the setup.

tests/test_input.py
jrandolf: Yea, the scroll is flaky. Probably because we don't use pixel precise wheel movements in CDP. I'm going to comment this test out.

OrKoN: scroll is on a different thread in the renderer, so that could be a reason for flakiness

jrandolf: That too. Scroll has a lot of problems :P

src/bidiMapper/domains/input/ActionDispatcher.ts
jrandolf: Resolving this for now. Reopen if needed.

src/bidiMapper/domains/input/InputSource.ts
```diff
@@ -0,0 +1,153 @@
...
+ */
+
+import {Input} from '../../../protocol/protocol.js';
+
+export import SourceType = Input.SourceActionsType;
+
+export class NoneSource {
+  type = SourceType.None as const;
+}
+export class KeySource {
```
sadym-chromium, line 25: It would be nice to have a comment here, where this constants came from

jrandolf: Which constants?

src/bidiMapper/domains/input/InputSource.ts
```diff
@@ -0,0 +1,153 @@
...
+      switch (button) {
+        case 0:
+          buttons |= 1;
+          break;
+        case 1:
+          buttons |= 4;
+          break;
+        case 2:
+          buttons |= 2;
+          break;
```
sadym-chromium, line 98: Add a comment for this as well please

src/bidiMapper/domains/input/InputSource.ts
```diff
@@ -0,0 +1,153 @@
...
+        case 4:
+          buttons |= 16;
+          break;
+      }
+    }
+    return buttons;
+  }
+
+  // --- Platform-specific state starts here ---
+  // This code should match https://source.chromium.org/chromium/chromium/src/+/refs/heads/main:ui/events/event.cc;l=479
```
sadym-chromium, line 111: Is it because of CDP considers a double click as one click, not as 2 separate clicks?

jrandolf: It's better to say CDP doesn't know what a double click is, but knows about a click count parameter that can trigger double clicks if it puts `2` in it.

src/bidiMapper/domains/input/InputSource.ts
sadym-chromium: return (this.#modifiers & 2) === 2;

src/bidiMapper/domains/input/InputSource.ts
sadym-chromium: Could you please add a comment about that in the code?

src/bidiMapper/domains/input/InputStateManager.ts
```diff
@@ -0,0 +1,39 @@
...
+ * limitations under the License.
+ */
+
+import {assert} from '../../../utils/assert.js';
+import {BrowsingContextImpl} from '../context/browsingContextImpl.js';
+
+import {InputState} from './InputState.js';
+
+export class InputStateManager {
+  #states = new WeakMap<BrowsingContextImpl, InputState>();
```
sadym-chromium, line 24: I'm confused on why is it `WeakMap`. Add a comment please

src/utils/assert.ts
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export function assert<T>(predicate: T): asserts predicate {
+  if (!predicate) {
+    throw new Error('Internal assertion failed.');
```
sadym-chromium, line 20:
```suggestion
    throw new UnknownErrorException('Internal assertion failed.');
```

src/bidiMapper/domains/input/InputStateManager.ts
jrandolf: This is what the spec specified we should do.

src/utils/assert.ts
jrandolf: Errors get converted to unknown errors. It'd probably be better to keep this as Error since utilities should not depend on other files.

sadym-chromium (approved):
LGTM after some comments will be added

tests/test_input.py
```diff
@@ -0,0 +1,385 @@
...
+            }],
+            'handle': ANY_STR
+        },
+        'realm': ANY_STR
+    } == result
+
+
+# TODO(jrandolf): Investigate mouse wheel flakiness. The deltaY sometimes
+# doubles and other times not. It is also not independent of the mouse.
+# @pytest.mark.asyncio
```
thiagowfx, line 280: Please use @pytest.mark.skip(reason="TODO: Implement") instead of submitting commented out code

tests/test_input.py
```diff
@@ -0,0 +1,385 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pytest
+from anys import ANY_STR
+from test_helpers import execute_command, goto_url
+
+SCRIPT_HTML = 'data:text/html,' + """
```
thiagowfx, line 20: Define the script without `data:text/html`. Then, in the test function, add the `html` fixture and use it as `html(SCRIPT)`

tests/test_input.py
```diff
@@ -0,0 +1,385 @@
...
+<div style="height: 2000px; width: 10px"></div>
+<script>
+    var allEvents = [];
```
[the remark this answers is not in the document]
jrandolf: We want access to the variable in evaluate calls.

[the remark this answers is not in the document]
thiagowfx: And `let` / `const` do not work for this purpose? (otherwise, I did not understand the comment)

PR #796, opened 2023-05-31
`browsingContext.captureScreenshot`: more spec compliance

thiagowfx:
`browsingContext.captureScreenshot`: more spec compliance

- Stop actively focusing the current tab.
- Use Page.getLayoutMetrics to retrieve the screenshot rect clip
- Pass sensible arguments to the CDP method.

Bug: #514

### DIFF
--- a/src/bidiMapper/domains/context/browsingContextImpl.ts
+++ b/src/bidiMapper/domains/context/browsingContextImpl.ts
@@ -579,15 +579,37 @@ export class BrowsingContextImpl {
     // This is needed because the screenshot gets blocked until the active tab gets focus.
     await this.#cdpTarget.cdpClient.sendCommand('Page.bringToFront');
 
-    const {cssContentSize: clip} = await this.#cdpTarget.cdpClient.sendCommand(
-      'Page.getLayoutMetrics'
-    );
+    let clip: Protocol.DOM.Rect;
+
+    if (this.isTopLevelContext()) {
+      clip = (
+        await this.#cdpTarget.cdpClient.sendCommand('Page.getLayoutMetrics')
+      ).cssContentSize;
+    } else {
+      const {
+        result: {value: iframeDocRect},
+      } = await this.#cdpTarget.cdpClient.sendCommand(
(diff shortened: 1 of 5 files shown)


thiagowfx:
@sadym-chromium This PR fixes a few issues w.r.t. screenshots, however the iframe capture is still not working, even with your previous suggestions :(

src/bidiMapper/domains/context/browsingContextImpl.ts
```diff
@@ -575,13 +575,21 @@ export class BrowsingContextImpl {
   }
 
   async captureScreenshot(): Promise<BrowsingContext.CaptureScreenshotResult> {
-    const [, result] = await Promise.all([
-      // XXX: Either make this a proposal in the BiDi spec, or focus the
-      // original tab right after the screenshot is taken.
-      // The screenshot command gets blocked until we focus the active tab.
-      this.#cdpTarget.cdpClient.sendCommand('Page.bringToFront'),
```
thiagowfx, line 582: Upon using layout metrics, this call isn't needed anymore!

thiagowfx: Actually: It's not needed in E2E, but it's still needed in WPT. Odd.

OrKoN (approved):

sadym-chromium:
> however the iframe capture is still not working, even with your previous suggestions :(

What exactly don't work? Which kind of screenshot is produced? Does it contain the iFrame content? Can the problem be in the framing?

thiagowfx:
> What exactly don't work? Which kind of screenshot is produced? Does it contain the iFrame content? Can the problem be in the framing?

Details here: https://chat.google.com/room/AAAAw7biRSY/SgGyjBGlkBc

TL;DR: Now it seemingly works. Previously it was taking a screenshot of the whole page (i.e. not the iframe only).

thiagowfx:
image

- left: actual screenshot of the iframe (with the current implementation)
- right: reference (the full page)

PR #797, opened 2023-06-01
browsingContext: add isOOPIF method

thiagowfx:
Reason about OOPIF cases in an abstracted manner.


sadym-chromium:
Shouldn't we just restrict screenshots of non-top level browsing contexts in the spec?

thiagowfx:
Closing in favor of changing the spec: https://github.com/w3c/webdriver-bidi/issues/441

PR #817, opened 2023-06-05
feat(browsingContext): implement `setViewport`

OrKoN:
Bug: #868
Spec: https://w3c.github.io/webdriver-bidi/#command-browsingContext-setViewport
Delegating edge-cases and extensive validation to WPT tests: https://github.com/web-platform-tests/wpt/pull/40584

### DIFF
--- a/tests/test_viewport.py
+++ b/tests/test_viewport.py
@@ -48,10 +48,9 @@ async def test_set_viewport(websocket, context_id):
         })
 
     assert [["width", {
-                "type": "number",
-                "value": 300
-            }], ["height", {
-                "type": "number",
-                "value": 300
-            }]
-    ] == result["result"]["value"]
+        "type": "number",
+        "value": 300
+    }], ["height", {
+        "type": "number",
+        "value": 300
+    }]] == result["result"]["value"]
--- a/tests/test_viewport.py
+++ b/tests/test_viewport.py
@@ -42,7 +42,7 @@ async def test_set_viewport(websocket, context_id):
                 "target": {
                     "context": context_id
                 },
-                "resultOwnership": "root",
+                "resultOwnership": "none",
                 "awaitPromise": True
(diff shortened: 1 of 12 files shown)


thiagowfx:
I created a new commit on top of this with a few adjustments. Implementation LGTM.

I just wonder if we should add more E2E tests.

thiagowfx:
Ah yes, we should add at least one more test, to exercise this part of the spec:

> If the width or height is outside of the implementation-specific or runtime constraints, return an [error](https://w3c.github.io/webdriver/#dfn-errors) with the [error code](https://w3c.github.io/webdriver/#dfn-error-code) [unsupported operation](https://w3c.github.io/webdriver/#dfn-unsupported-operation).

And maybe another one to test non-top-level contexts.

thiagowfx:
Synced off-line with @OrKoN: He will continue this PR. Edited the comment above with some TODOs once you continue it.

tests/test_viewport.py
```diff
@@ -0,0 +1,64 @@
...
+
+    result = await execute_command(
+        websocket, {
+            "method": "script.evaluate",
+            "params": {
+                "expression": "({'width': window.innerWidth, 'height': window.innerHeight})",
+                "target": {
+                    "context": context_id
+                },
+                "resultOwnership": "root",
```
sadym-chromium, line 45: NIT: In this way there will be no `handle` in the result.
```suggestion
                "resultOwnership": "none",
```

sadym-chromium (approved):
Consider adding (optional) changes to e2e

OrKoN:
I have overlooked the e2e test suggestions, will open a follow-up for the unsupported operation (I am not sure if we have a limit in Chrome) and the rest should be covered by WPT.

thiagowfx:
LGTM

PR #831, opened 2023-06-09
fix: network response fromCache

Lightning00Blade:


thiagowfx (approved):

thiagowfx (approved):

Lightning00Blade:
The test for `fromCache=True` seemed flaky so I disabled it for now as it's currently disable on main as well.

thiagowfx (approved):

PR #832, opened 2023-06-09
Extract Channel logic

sadym-chromium:
Preparation for the preload script channels. Extract the channel-related logic into a dedicated file.

### DIFF
--- a/src/bidiMapper/domains/script/channelProxy.ts
+++ b/src/bidiMapper/domains/script/channelProxy.ts
@@ -1,0 +1,189 @@
+/*
+ * Copyright 2023 Google LLC.
+ * Copyright (c) Microsoft Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+import {CommonDataTypes, Script} from '../../../protocol/protocol';
+import {Realm} from './realm';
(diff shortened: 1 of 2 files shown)


src/bidiMapper/domains/script/channelProxy.ts
```diff
@@ -0,0 +1,189 @@
...
+          }
+        ),
+        arguments: [{objectId: channelHandle}],
+        executionContextId: realm.executionContextId,
+        serializationOptions: {
+          serialization: 'deep',
+        },
+      }
+    );
+    return sendMessageArgResult.result.objectId!;
```
thiagowfx, line 142: I would inline these (i.e. delete the `const` declaration)

src/bidiMapper/domains/script/channelProxy.ts
```diff
@@ -0,0 +1,189 @@
...
+import Handle = CommonDataTypes.Handle;
+
+export class ChannelProxy {
+  readonly #channel: Script.ChannelProperties;
+  readonly #eventManager: IEventManager;
+  readonly #realm: Realm;
+  readonly #channelHandle: CommonDataTypes.Handle;
+  readonly #sendMessageHandle: CommonDataTypes.Handle;
+
+  get sendMessageHandle(): CommonDataTypes.Handle {
```
thiagowfx, line 31: nit: move this getter after the constructor

src/bidiMapper/domains/script/channelProxy.ts
```diff
@@ -0,0 +1,189 @@
...
+ * limitations under the License.
+ *
+ */
+
+import {CommonDataTypes, Script} from '../../../protocol/protocol';
+import {Realm} from './realm';
+import {IEventManager} from '../events/EventManager';
+import Handle = CommonDataTypes.Handle;
+
+export class ChannelProxy {
```
thiagowfx, line 24: jsdoc would be nice

thiagowfx (changes requested):

src/bidiMapper/domains/script/channelProxy.ts
sadym-chromium: IMO it would be way less readable then

thiagowfx: OK

PR #837, opened 2023-06-09
input: enable a few more WPT tests

thiagowfx:


thiagowfx:
Finally got it: `webdriver/tests/bidi/input/perform_actions/wheel.py.ini` is the flaky one. So, for now, we can enable all of them except the wheel.

cc @jrandolf FYI

PR #848, opened 2023-06-13
eslint: miscellaneous changes

thiagowfx:


src/bidiMapper/domains/network/networkRequest.ts
```diff
@@ -290,12 +290,12 @@ export class NetworkRequest {
...
           protocol: this.#responseReceivedEvent.response.protocol ?? '',
           status:
-            this.#responseReceivedExtraInfoEvent?.statusCode ||
+            this.#responseReceivedExtraInfoEvent?.statusCode ??
             this.#responseReceivedEvent.response.status,
           statusText: this.#responseReceivedEvent.response.statusText,
           fromCache:
-            this.#responseReceivedEvent.response.fromDiskCache ||
-            this.#responseReceivedEvent.response.fromPrefetchCache ||
+            this.#responseReceivedEvent.response.fromDiskCache ??
```
Lightning00Blade, line 297: I think this change is incorrect all values are boolean if one is true we want to set the value to true.

src/protocol/protocol.ts
```diff
@@ -423,9 +423,9 @@ export namespace CommonDataTypes {
     type: 'function';
   };
 
-  export type RegExpRemoteValue = RemoteReference & RegExpLocalValue;
+  export type RegExpRemoteValue = RegExpLocalValue & RemoteReference;
```
Lightning00Blade, line 426: I don't know how useful this sort rule is, looking at the code now it would be harder to check what extends what imo.

thiagowfx: I was on the fence about this one. Disabled it for unions, but for intersections it seemed OK. Let's disable it fully then.

src/bidiMapper/domains/network/networkRequest.ts
thiagowfx: Will revert.

Lightning00Blade (approved):

PR #849, opened 2023-06-14
Move preload script to a dedicated class

sadym-chromium:
Preparation for channels in preload scripts. Part 2.
* Move preload script logic to a dedicated class.
* Fix a couple of tests.
* Removed support of sandboxes, as [Page.createIsolatedWorld](https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-createIsolatedWorld) is needed for that. And after such a call it's too late.
* Don't wait for the initial run of the preload script to be finished.


tests/conftest.py
```diff
@@ -123,6 +123,22 @@ def url_all_origins(request):
     return request.param
 
 
+# Reads the given number of messages from the websocket, and returns them
```
thiagowfx, line 126: Use a docstring inside the `def` method. This integrates with `pytest --fixtures` and with python docs that are surfaced in IDEs.

tests/conftest.py
```diff
@@ -123,6 +123,22 @@ def url_all_origins(request):
...
+# in consistent order.
+@pytest.fixture
+def read_sorted_messages(websocket):
+
+    async def _(message_count):
```
thiagowfx, line 131: You can call the inner function `read_sorted_messages` as well (I've seen this pattern in WPT fixtures IIRC)

Lightning00Blade:
Can we maybe hold off merging this, as I want to make release after #850 is merged and a release is made. As I am not sure how it will affect Puppeteer.

src/bidiMapper/domains/context/PreloadScriptStorage.spec.ts
```diff
@@ -199,33 +99,23 @@ describe('PreloadScriptStorage', () => {
...
-        expect(preloadScriptStorage.findPreloadScripts(filter)).to.deep.equal([
-          {
-            id: MOCKED_UUID_1,
-            cdpPreloadScripts: [cdpPreloadScript1],
-            contextId: null,
-            functionDeclaration,
-            sandbox,
-          },
-        ]);
+        expect(preloadScriptStorage.findPreloadScripts(filter)[0]).to.equal(
```
thiagowfx, line 111: Could you keep the same pattern as it used to be previously, by comparing the whole array, instead of only `[0]`?

src/bidiMapper/domains/context/bidiPreloadScript.ts
```diff
@@ -0,0 +1,140 @@
...
+    this.#cdpPreloadScripts.push({
+      target: cdpTarget,
+      preloadScriptId: addCdpPreloadScriptResult.identifier,
+    });
+    this.#targetIds.add(cdpTarget.targetId);
+  }
+
+  /**
+   * Schedules the script to be run right after
+   * `Runtime.runIfWaitingForDebugger`, but does not wait for result.
```
thiagowfx, line 106: Add:

`XXX: Make it possible to evaluate in a specific sandbox / worldName` (or similar)

src/bidiMapper/domains/context/bidiPreloadScript.ts
```diff
@@ -0,0 +1,139 @@
...
+export class BidiPreloadScript {
+  /** BiDi ID, an automatically generated UUID. */
+  readonly #id: string = uuidv4();
+  /** CDP preload scripts. */
+  #cdpPreloadScripts: CdpPreloadScript[] = [];
+  /** The script itself, in a format expected by the spec i.e. a function. */
+  readonly #functionDeclaration: string;
+  /** Browsing context ID. */
+  readonly #contextId: CommonDataTypes.BrowsingContext | null;
+  readonly #targetIds: Set<string> = new Set();
```
thiagowfx, line 49: Document this

src/bidiMapper/domains/context/bidiPreloadScript.ts
```diff
@@ -0,0 +1,139 @@
...
+    if (params.arguments !== undefined && params.arguments.length > 0) {
+      // TODO: Handle arguments.
+      throw new Error('add preload script arguments are not supported');
+    }
+
+    this.#functionDeclaration = params.functionDeclaration;
+    this.#contextId = params.context ?? null;
+  }
+
+  async initInTargets(cdpTargets: Set<CdpTarget>) {
```
thiagowfx, line 81: I'd argue this can be less strict than `Set`.

`Iterable` is more forgiving. Or just a plain array.

It's totally fine to create a `Set` within this function for the sake of deduplication, but let's make the interface more forgiving

(i.e. basically https://en.wikipedia.org/wiki/Robustness_principle)

thiagowfx (changes requested):

src/bidiMapper/domains/context/bidiPreloadScript.ts
thiagowfx: True, it's redundant, I think it would just be clearer in terms of TODO for us, to remember what specific part of CDP we want to extend.

I am fine if you prefer to omit it.

src/bidiMapper/domains/context/bidiPreloadScript.ts
sadym-chromium: done

src/bidiMapper/domains/context/bidiPreloadScript.ts
sadym-chromium: done

src/bidiMapper/domains/context/PreloadScriptStorage.spec.ts
sadym-chromium: done

src/bidiMapper/domains/context/browsingContextProcessor.ts
```diff
@@ -367,11 +342,7 @@ export class BrowsingContextProcessor {
     }
 
     for (const script of scripts) {
-      for (const cdpPreloadScript of script.cdpPreloadScripts) {
-        const cdpTarget = cdpPreloadScript.target;
-        const cdpPreloadScriptId = cdpPreloadScript.preloadScriptId;
-        await cdpTarget.removePreloadScript(cdpPreloadScriptId);
-      }
+      await script.remove();
     }
```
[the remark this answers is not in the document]
sadym-chromium: done

tests/conftest.py
sadym-chromium: done

tests/conftest.py
sadym-chromium: done

thiagowfx (approved):

PR #850, opened 2023-06-14
feat: adds script.realmCreated

Lightning00Blade:


src/bidiMapper/bidiMapper.ts
```diff
@@ -26,6 +26,7 @@ export type {
   ICdpConnection as CdpConnection,
 } from '../cdp/cdpConnection.js';
 export type {ICdpClient, ICdpClient as CdpClient} from '../cdp/cdpClient.js';
+export {EventEmitter} from '../utils/EventEmitter.js';
```
Lightning00Blade, line 29: This is needed as Puppeteer EventEmitter is incomparable with BiDi Mapper EventEmitter.

tests/test_realm.py
```diff
@@ -0,0 +1,47 @@
...
+    # Wait for `script.realmCreated` event.
+    assert {
+        "method": "script.realmCreated",
+        "params": {
+            "type": "window",
+            "origin": "null",
+            "realm": ANY_STR,
+            "context": context_id,
+        }
+    } == response[0]
```
Lightning00Blade, line 47: Help needed here. The response is a tuple with 1 element.

src/bidiMapper/bidiMapper.ts
thiagowfx: Can this be `export type`, though?

src/protocol/protocol.ts
```diff
@@ -787,7 +787,18 @@ export namespace Script {
...
+  >;
+
+  export type RealmCreated = {
+    method: 'script.realmCreated';
+    params: RealmInfo;
+  };
+
   export enum EventNames {
+    RealmCreated = 'script.realmCreated',
     MessageEvent = 'script.message',
```
thiagowfx, line 802: Please sort the event names

Could also wrap `// keep-sorted` around here, but not absolutely needed as there are only two entries

tests/test_realm.py
```diff
@@ -0,0 +1,48 @@
+# Copyright 2023 Google LLC.
```
thiagowfx, line 1: Should this be in `test_browsing_context.py` instead? @sadym-chromium thoughts?

src/bidiMapper/bidiMapper.ts
Lightning00Blade: Types are different, I am happy if we align with something external - Node EventEmitter and we use that as a CommonInterface. 
Else we can't as you extend the interface CDPClient with the class and that brakes Typescript as its looking for `#private` which can't be provided.

tests/test_realm.py
```diff
@@ -0,0 +1,48 @@
...
+    # Wait for `script.realmCreated` event.
+    assert {
+        "method": "script.realmCreated",
+        "params": {
+            "type": "window",
+            "origin": "null",
+            "realm": ANY_STR,
+            "context": context_id,
+        }
+    } == response[0]
```
thiagowfx, line 48:
```suggestion
    # Wait for `script.realmCreated` event.
    assert ({
        "method": "script.realmCreated",
        "params": {
            "type": "window",
            "origin": "null",
            "realm": ANY_STR,
            "context": context_id,
        }
    },) == response
```

tests/test_realm.py
thiagowfx: See suggestion below

src/bidiMapper/bidiMapper.ts
thiagowfx: Sure, we could look into that in another PR.

sadym-chromium (approved):

PR #851, opened 2023-06-14
fix: screenshots failing when setViewport is used

Lightning00Blade:
The image was capturing empty parts if the content was overflowing.
Image has 200 boxes
image


thiagowfx (approved):

thiagowfx:
Would you like to add an E2E test for that?

Issue #858, opened 2023-06-16
Minified mapper contains non-ascii characters, which fails chromium presubmits when embedding it


thiagowfx:
How to test:

```
$ LC_ALL=C tr -d '[:print:]' < lib/cjs/bidiTab/bidiTab.js | tr -d '\n'
◂▸%
```

It causes the following issue:

- CL: https://chromium-review.googlesource.com/c/chromium/src/+/4615087
- Logs: https://ci.chromium.org/ui/p/chromium/builders/try/win-rel/233404/overview

```
Step compile (with patch) failed. Error logs are shown below:
[62/1772] ACTION //chrome/test/chromedriver:embed_bidimapper_in_cpp(//build/toolchain/win:win_clang_x64)
FAILED: gen/chrome/test/chromedriver/bidimapper/bidimapper.cc gen/chrome/test/chromedriver/bidimapper/bidimapper.h
C:/b/s/w/ir/cipd_bin_packages/cpython3/bin/python3.exe ../../chrome/test/chromedriver/embed_bidimapper_in_cpp.py --directory gen/chrome/test/chromedriver/bidimapper gen/third_party/bidimapper/mapper.js
Traceback (most recent call last):
File "../../chrome/test/chromedriver/embed_bidimapper_in_cpp.py", line 43, in <module>
sys.exit(main())
File "../../chrome/test/chromedriver/embed_bidimapper_in_cpp.py", line 38, in main
...The message was too long...
```

thiagowfx:
cc @mathiasbynens @Lightning00Blade @sadym-chromium FYI

thiagowfx:
I tried:

```
    terser({
      format: {
        ascii_only: true,
      },
    }),
```

...in `rollup.config.mjs` but it makes no difference.

I also tried:

```
    terser({
      output: {
        ascii_only: true,
      },
    }),
```

...which also makes no difference. `output` is deprecated though.

Will investigate other approaches.

thiagowfx:
https://github.com/mbostock/rollup-plugin-ascii also does not work:

```
% npm run clean; npm run build

> chromium-bidi@0.4.12 clean
> rimraf lib .eslintcache .wireit


> chromium-bidi@0.4.12 build
> wireit

🏃 [tsc] Running command "tsc --build src/tsconfig.json --pretty"
✅ [tsc] Executed successfully
🏃 [rollup] Running command "rollup -c"

[...]

PR #860, opened 2023-06-16
Restore non-ASCII characters where applicable, ensuring they get escaped during minification

mathiasbynens:
```
$ npm run build; npx is-ascii-safe-cli lib/iife/mapperTab.js

> chromium-bidi@0.4.12 build
> wireit

🏃 [tsc] Running command "tsc --build src/tsconfig.json --pretty"
✅ [tsc] Executed successfully
🏃 [rollup] Running command "rollup -c"

lib/cjs/bidiTab/bidiTab.js → lib/iife/mapperTab.js...
created lib/iife/mapperTab.js in 1.2s
✅ [rollup] Executed successfully
✅ [build] No command to execute
1 file(s) are ASCII-safe.
```

Bug: #858

### DIFF
--- a/rollup.config.mjs
+++ b/rollup.config.mjs
@@ -33,6 +33,10 @@ export default {
       // without webcrypto exposes globally.
       ignore: ['crypto'],
     }),
-    terser(),
+    terser({
+      format: {
+        ascii_only: true,
+      },
+    }),
   ],
 };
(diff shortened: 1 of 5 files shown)


thiagowfx (approved):

thiagowfx:
Let's merge it with the intent of rolling the mapper and see if it works as intended. Either way it seems to be a no-op in the case it doesn't work.

PR #865, opened 2023-06-19
feat: add fragmentNavigated event

Lightning00Blade:
Unblocks `waitForNavigation` in Puppeteer.

After talking with @sadym-chromium we should only give navigation to the initial request (document request).
But all CDP requests triggered by initial parsing of the document have the LoaderId set, so we relay on requestId matching the loaderId.

### DIFF
--- a/src/bidiMapper/domains/network/networkRequest.ts
+++ b/src/bidiMapper/domains/network/networkRequest.ts
@@ -24,7 +24,7 @@ import type Protocol from 'devtools-protocol';
 
 import {Deferred} from '../../../utils/deferred.js';
 import type {IEventManager} from '../events/EventManager.js';
-import {Network} from '../../../protocol/protocol.js';
+import {Network, type BrowsingContext} from '../../../protocol/protocol.js';
 
 export class NetworkRequest {
   static #unknown = 'UNKNOWN';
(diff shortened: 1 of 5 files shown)


src/bidiMapper/domains/network/networkRequest.ts
```diff
@@ -170,6 +170,20 @@ export class NetworkRequest {
     };
   }
 
+  #getNavigationId(): BrowsingContext.Navigation | null {
+    if (
+      !this.#requestWillBeSentEvent?.loaderId ||
+      // All request that are initiated by the navigation
+      // receive loaderId we want only the initial request
+      // to have the navigation set
```
thiagowfx, line 178: Could you rephrase this? It sounds a bit confusing.

thiagowfx (changes requested):
It would be nice to add at least one test that exercises the newly added event

(or augment an existing test)

tests/test_browsing_context.py
```diff
@@ -922,3 +922,37 @@ async def test_browsingContext_ignoreCache(websocket, context_id, ignoreCache):
...
         "result": {},
     }
+
+
+@pytest.mark.asyncio
+async def test_browsingContext_fragmentNavigated_event(websocket, context_id,
+                                                       html):
+    url = html()
+
+    await subscribe(websocket, ["browsingContext.fragmentNavigated"])
```
thiagowfx, line 932: Nit: when listening to a single element,  you can pass a string directly (instead of an array).

thiagowfx (approved):

tests/test_browsing_context.py
Lightning00Blade: ack

PR #866, opened 2023-06-19
chore: small fix

Lightning00Blade:
The `Target.targetInfoChanged` does not have an SessionId as it happens on the top level.
I am was looking into the NavigationStarted (BiDi implementation) event when I notice that.

### DIFF
--- a/src/bidiMapper/domains/context/browsingContextImpl.ts
+++ b/src/bidiMapper/domains/context/browsingContextImpl.ts
@@ -291,17 +291,11 @@ export class BrowsingContextImpl {
     };
   }
 
-  #initListeners() {
-    this.#cdpTarget.cdpClient.on(
-      'Target.targetInfoChanged',
-      (params: Protocol.Target.TargetInfoChangedEvent) => {
-        if (this.id !== params.targetInfo.targetId) {
-          return;
-        }
-        this.#url = params.targetInfo.url;
-      }
-    );
+  onTargetInfoChanged(params: Protocol.Target.TargetInfoChangedEvent) {
+    this.#url = params.targetInfo.url;
+  }
 
+  #initListeners() {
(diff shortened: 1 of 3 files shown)


src/bidiMapper/domains/context/browsingContextImpl.ts
```diff
@@ -291,17 +291,11 @@ export class BrowsingContextImpl {
...
-    this.#cdpTarget.cdpClient.on(
-      'Target.targetInfoChanged',
-      (params: Protocol.Target.TargetInfoChangedEvent) => {
-        if (this.id !== params.targetInfo.targetId) {
-          return;
-        }
-        this.#url = params.targetInfo.url;
-      }
-    );
+  targetInfoChanged(params: Protocol.Target.TargetInfoChangedEvent) {
```
thiagowfx, line 294: update this method name to a verb phrase, maybe "updateTargetInfoUrl". Otherwise it reads like an event

sadym-chromium:
Could you please add an e2e, which will fail without your change?

thiagowfx (approved):

Lightning00Blade:
@sadym-chromium I don't think we have a usecase atm. Also at the moment the this part of code
```
this.#cdpTarget.cdpClient.on(
      'Target.targetInfoChanged',
      (params: Protocol.Target.TargetInfoChangedEvent) => {
        if (this.id !== params.targetInfo.targetId) {
          return;
        }
        this.#url = params.targetInfo.url;
      }
    );
```
is never hit. I wanted to fix it as I was looking if the NavigationStarted event, and it seems this event is the one that can give us the correct URL for that event, so I kept it rather then remove it.
Once NavigationStarted is implemented then a test would fail if this is not added.

sadym-chromium (approved):

PR #872, opened 2023-06-20
eslint: add promise-function-async

thiagowfx:
https://typescript-eslint.io/rules/promise-function-async/


thiagowfx:
I believe this is a no-op that helps with readability. If it's not a no-op, then we must be careful.

Lightning00Blade:
I personally don't like the change, we can discuss a little bit offline, maybe there is something I am missing.

My opinion stems from that its the `async/await` pattern and no awaits are use. 
This makes it more confusing for me as I am expecting the await in the function.

thiagowfx:
> I personally don't like the change, we can discuss a little bit offline, maybe there is something I am missing.

I have no strong stance on this one and at this point would like to add only checks that are universally useful / agreeable upon. If you don't like it, then let's not add it.

Issue #765, opened 2023-05-23
Implement `network.responseStarted` event


thiagowfx:
Spec: https://w3c.github.io/webdriver-bidi/#event-network-responseStarted

## Tracking

- [x] ~~Ping(2023-05-31): Follow-up with upstream: https://github.com/w3c/webdriver-bidi/pull/204#issuecomment-1570324761~~
- [x] Ping(2023-05-31): Follow-up with upstream: https://github.com/w3c/webdriver-bidi/issues/435

thiagowfx:
See https://github.com/w3c/webdriver-bidi/pull/204#issuecomment-1570324761.

PR #875, opened 2023-06-20
network: populate response headers size

thiagowfx:
Note: In the absence of WPT tests that verify this field, it may be slightly off by a few bytes.

Bug: #765

### DIFF
--- a/src/bidiMapper/domains/network/networkRequest.ts
+++ b/src/bidiMapper/domains/network/networkRequest.ts
@@ -292,6 +292,10 @@ export class NetworkRequest {
       this.#responseReceivedExtraInfoEvent = undefined;
     }
 
+    const headers = NetworkRequest.#getHeaders(
+      this.#responseReceivedEvent.response.headers
+    );
+
     return {
       method: Network.EventNames.ResponseCompletedEvent,
       params: {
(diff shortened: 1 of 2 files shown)


src/bidiMapper/domains/network/networkRequest.ts
```diff
@@ -309,13 +313,10 @@ export class NetworkRequest {
...
             this.#servedFromCache,
-          headers: NetworkRequest.#getHeaders(
-            this.#responseReceivedEvent.response.headers
-          ),
+          headers,
           mimeType: this.#responseReceivedEvent.response.mimeType,
           bytesReceived: this.#responseReceivedEvent.response.encodedDataLength,
-          headersSize:
-            this.#responseReceivedExtraInfoEvent?.headersText?.length ?? 0,
+          headersSize: this.#computeResponseHeadersSize(headers),
```
sadym-chromium, line 319: Could you please add e2e for it?

thiagowfx: Done

tests/test_network.py
```diff
@@ -232,7 +232,7 @@ async def test_network_network_response_completed_event_emitted(
                 "headers": ANY_LIST,
                 "mimeType": "text/html",
                 "bytesReceived": ANY_NUMBER,
-                "headersSize": ANY_NUMBER,
+                "headersSize": AnyOr(338, 355, 360),
```
Lightning00Blade, line 235: This feels like it can be flaky, as it looks for 3 specific values, maybe just looking for a range between 300 and 400?

thiagowfx: Updated

PR #877, opened 2023-06-21
feat: add realmDestroyed event

Lightning00Blade:


src/bidiMapper/domains/script/channelProxy.ts
```diff
@@ -248,13 +248,12 @@ export class ChannelProxy {
...
             // `getEvalInWindowStr` will resolve the promise later.
             return new Promise((resolve) => (w[id] = resolve));
-          } else {
-            // The channelProxy is already created by `getEvalInWindowStr` and
-            // is set into window property. Return it.
-            const channelProxy = w[id];
-            delete w[id];
-            return channelProxy;
           }
+          // The channelProxy is already created by `getEvalInWindowStr` and
```
sadym-chromium, line 252: NIT: `else` was there to ease reading. Feel free to ignore this comment.

Lightning00Blade: This change was made by running `npm run format` so I am going to leave it. 
No preference on my side.

src/bidiMapper/domains/script/realmStorage.ts
```diff
@@ -117,6 +136,15 @@ export class RealmStorage {
   deleteRealms(filter: RealmFilter) {
     this.findRealms(filter).map((realm) => {
       this.#realmMap.delete(realm.realmId);
```
sadym-chromium, line 138: may it would be more logical to send event from the realm itself rather than from the storage?

Lightning00Blade: We should as it follows the same logic as the `BrowsingContextImpl`. Same method `.delete`

tests/test_realm.py
```diff
@@ -46,3 +48,111 @@ async def test_realm_realmCreated(websocket, context_id, html):
...
+    await subscribe(websocket, "script.realmCreated")
+
+    await send_JSON_command(
+        websocket, {
+            "id": 21,
+            "method": "script.evaluate",
+            "params": {
+                "target": {
+                    "context": context_id,
+                    "sandbox": "__test_sandbox__"
```
sadym-chromium, line 65: NIT: `SOME_SANDBOX`?

tests/test_realm.py
```diff
@@ -46,3 +48,111 @@ async def test_realm_realmCreated(websocket, context_id, html):
...
+            "id": 12,
+            "method": "browsingContext.close",
+            "params": {
+                "context": context_id,
+            }
+        })
+
+    response = await read_JSON_message(websocket),
+
+    # Wait for `script.realmCreated` event.
```
sadym-chromium, line 103: not related comment

tests/test_realm.py
```diff
@@ -46,3 +48,111 @@ async def test_realm_realmCreated(websocket, context_id, html):
...
+    response = await read_JSON_message(websocket),
+
+    assert ({
+        "method": "script.realmDestroyed",
+        "params": {
+            "realm": ANY_STR,
+        }
+    }, ) == response
+
+    # Waits for the comm
```
sadym-chromium, line 149: unfinished comment

tests/test_realm.py
```diff
@@ -46,3 +48,111 @@ async def test_realm_realmCreated(websocket, context_id, html):
...
+    await send_JSON_command(
+        websocket, {
+            "id": 22,
+            "method": "browsingContext.close",
+            "params": {
+                "context": context_id,
+            }
+        })
+
+    response = await read_JSON_message(websocket),
```
sadym-chromium, line 140: extra comma at the end of this line makes the `response` to be a tuple, which forces the assertion to have that weird wrapper 
```
({
    ...
}, ) == response
```

tests/test_realm.py
```diff
@@ -46,3 +48,111 @@ async def test_realm_realmCreated(websocket, context_id, html):
...
+    assert ({
+        "method": "script.realmDestroyed",
+        "params": {
+            "realm": ANY_STR,
+        }
+    }, ) == response
+
+    # Waits for the comm
+
+    response = await read_JSON_message(websocket),
```
sadym-chromium, line 151: same comma here

sadym-chromium (approved):
LGTM after the comments are addressed

tests/test_realm.py
```diff
@@ -26,6 +27,7 @@ async def test_realm_realmCreated(websocket, context_id, html):
 
     await send_JSON_command(
         websocket, {
+            "id": 21,
```
sadym-chromium, line 30: You can omit the id and it will be assigned automatically

src/bidiMapper/BidiServer.ts
```diff
@@ -36,8 +36,8 @@ export class BidiServer extends EventEmitter<BidiServerEvents> {
   #messageQueue: ProcessingQueue<OutgoingBidiMessage>;
   #transport: BidiTransport;
   #commandProcessor: CommandProcessor;
-  #browsingContextStorage: BrowsingContextStorage;
-  #realmStorage: RealmStorage;
```
sadym-chromium, line 40: Please refrain from reformat when doing functional changes unless it is required, as it can mask some issues.

tests/test_realm.py
sadym-chromium: you don't need it in this test, just FYI. `send_JSON_command` returns a command id, which can be used later on for assertions

tests/test_realm.py
```diff
@@ -34,15 +36,118 @@ async def test_realm_realmCreated(websocket, context_id, html):
...
+
+
+@pytest.mark.asyncio
+async def test_realm_realmDestroyed(websocket, context_id):
+
+    await subscribe(websocket, "script.realmDestroyed")
+
+    await send_JSON_command(
+        websocket, {
+            "id": 12,
```
sadym-chromium, line 92: same thong with `id` here

src/bidiMapper/BidiServer.ts
Lightning00Blade: Those 2 classes need to be initialize before all, as they are probed on some Class creation and result in errors if the order it wrong this insures the order is always correct.

thiagowfx (commented):
LGTM

PR #878, opened 2023-06-21
chore: use release-please for publishing

jrandolf:
Fixed: #880


.github/workflows/release-publish.yml
```diff
@@ -1,28 +1,48 @@
...
 
 jobs:
+  release-pr:
+    name: Release PR
+    runs-on: ubuntu-latest
+    outputs:
+      created: ${{ steps.release.outputs.release_created }}
+    steps:
+      - name: Use release-please
+        uses: google-github-actions/release-please-action@v3
```
Lightning00Blade, line 20: Use SHA with comment version

Lightning00Blade (approved):
LGTM+ comments

.github/workflows/publish.yml
```diff
@@ -1,28 +1,33 @@
...
+
+env:
+  FORCE_COLOR: 3
+  PIP_DISABLE_PIP_VERSION_CHECK: 1
 
 jobs:
   publish:
+    name: Publish
     runs-on: ubuntu-latest
     steps:
```
thiagowfx, line 16: FYI: I am intending to revert some of these changes in my next misc PR. Most of our YAML files look consistent, in terms of `@v3`, having `env` at the bottom of the file, the order of the steps, etc, and this change breaks some of this inconsistency.

In the future, if you make these changes, then at least try to be globally consistent/uniform.

OrKoN: Could you elaborate why env should be at the bottom of the file?

thiagowfx: It doesn't. My argument is in favor of consistency. `env` could equally be in the beginning, but if so, let's keep all files the same way.

OrKoN: I find the env at the top to be more readable as for other changes no preference :)

PR #881, opened 2023-06-22
feat: add `browsingContext.navigationStarted`

Lightning00Blade:
Still test need to be added, but wanted to see if the implementation makes sense.

### DIFF
--- a/src/bidiMapper/domains/context/browsingContextImpl.ts
+++ b/src/bidiMapper/domains/context/browsingContextImpl.ts
@@ -65,6 +65,7 @@ export class BrowsingContextImpl {
   #loaderId?: Protocol.Network.LoaderId;
   #cdpTarget: CdpTarget;
   #maybeDefaultRealm: Realm | undefined;
+  #isNavigating = false;
   readonly #logger?: LoggerFn;
 
   private constructor(
@@ -292,6 +293,24 @@ export class BrowsingContextImpl {
 
   onTargetInfoChanged(params: Protocol.Target.TargetInfoChangedEvent) {
     this.#url = params.targetInfo.url;
+
+    if (this.#isNavigating) {
+      this.#eventManager.registerEvent(
+        {
+          method: BrowsingContext.EventNames.NavigationStarted,
+          params: {
+            context: this.id,
+            // TODO: The network event is send before the CDP Page.frameStartedLoading
+            // It theory there should be a way to get the data.
+            navigation: null,
+            timestamp: BrowsingContextImpl.getTimestamp(),
+            url: this.#url,
+          },
+        },
+        this.id
+      );
+      this.#isNavigating = false;
+    }
   }
 
   #initListeners() {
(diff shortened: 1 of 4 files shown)


src/bidiMapper/domains/context/browsingContextImpl.ts
```diff
@@ -293,129 +294,141 @@ export class BrowsingContextImpl {
...
+          params: {
+            context: this.id,
+            navigation: null,
+            timestamp: BrowsingContextImpl.getTimestamp(),
+            url: this.#url,
+          },
+        },
+        this.id
+      );
+      this.#navigationStarted = false;
```
sadym-chromium, line 311: I guess there are other cases when the navigation stopped, like [`Page.frameStoppedLoading`](https://chromedevtools.github.io/devtools-protocol/tot/Page/#event-frameStoppedLoading).

sadym-chromium: please try to make sure there are no other events pointing the navigation stopped

src/bidiMapper/domains/context/browsingContextImpl.ts
```diff
@@ -293,129 +294,141 @@ export class BrowsingContextImpl {
...
+    }
   }
 
   #initListeners() {
-    this.#cdpTarget.cdpClient.on(
-      'Page.frameNavigated',
-      (params: Protocol.Page.FrameNavigatedEvent) => {
```
sadym-chromium, line 301: Even if the types are removed by auto formatting tool, this adds unnecessary noise to the PR, which can mask other changes

src/bidiMapper/domains/context/browsingContextImpl.ts
Lightning00Blade: It looks correct that the [Page.frameStoppedLoading](https://chromedevtools.github.io/devtools-protocol/tot/Page/#event-frameStoppedLoading) is the only event that signals that the navigation has stopped.

tests/test_browsing_context.py
```diff
@@ -956,3 +956,32 @@ async def test_browsingContext_fragmentNavigated_event(websocket, context_id,
...
     assert navigation['result']['navigation'] == response['params'][
         'navigation']
+
+
+@pytest.mark.asyncio
+async def test_browsingContext_navigationStarted_event(websocket, context_id,
+                                                       html):
+    url = html()
+
+    await subscribe(websocket, ["browsingContext.navigationStarted"])
```
thiagowfx, line 966: you can drop the brackets as we're passing a single event

thiagowfx (approved):

tests/test_browsing_context.py
Lightning00Blade: I use it as it more closely resembles BiDi, happy to change if there is already agreed way for this.

thiagowfx: then it would be nicer to drop the string option, in order to make it unambiguous w.r.t. style (but then it's out of scope of this PR)

PR #886, opened 2023-06-26
test: add a test for unsupported viewports

OrKoN:
Follow-up to https://github.com/GoogleChromeLabs/chromium-bidi/pull/817

### DIFF
--- a/src/bidiMapper/domains/context/browsingContextImpl.ts
+++ b/src/bidiMapper/domains/context/browsingContextImpl.ts
@@ -619,16 +619,30 @@ export class BrowsingContextImpl {
         'Emulation.clearDeviceMetricsOverride'
       );
     } else {
-      await this.#cdpTarget.cdpClient.sendCommand(
-        'Emulation.setDeviceMetricsOverride',
-        {
-          width: viewport.width,
-          height: viewport.height,
-          deviceScaleFactor: 0,
-          mobile: false,
-          dontSetVisibleSize: true,
+      try {
+        await this.#cdpTarget.cdpClient.sendCommand(
+          'Emulation.setDeviceMetricsOverride',
+          {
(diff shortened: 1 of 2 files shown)


tests/test_viewport.py
```diff
@@ -53,3 +55,27 @@ async def test_set_viewport(websocket, context_id):
...
+
+
+    with pytest.raises(Exception) as exception_info:
+      await execute_command(
+          websocket, {
+              "method": "browsingContext.setViewport",
+              "params": {
+                  "context": context_id,
+                  "viewport": {
+                      "width": 10000001,
```
Lightning00Blade, line 72: Can we have a link to the code that the max value is set. Atm the number seems to come from thin air.

Lightning00Blade (approved):

tests/test_viewport.py
OrKoN: There is a link to the code where the error is thrown.

OrKoN: Updated

src/bidiMapper/domains/context/browsingContextImpl.ts
```diff
@@ -619,16 +619,30 @@ export class BrowsingContextImpl {
...
+      } catch (err) {
+        if (
+          (err as Error).message.startsWith(
+            // https://crsrc.org/c/content/browser/devtools/protocol/emulation_handler.cc;l=257;drc=2f6eee84cf98d4227e7c41718dd71b82f26d90ff
+            'Width and height values must be positive'
+          )
+        ) {
+          throw new Message.UnsupportedOperationException(
+            'Provided viewport dimensions are not supported'
+          );
```
thiagowfx, line 642: I would print the provided dimensions in the error message as well

PR #891, opened 2023-06-26
chore: use common prefix for debug

Lightning00Blade:
As suggested by DEBUG we should have a common prefix for all our  statements, that also remove some clutter around 
`bidiMapper:mapperDebug` that has mapper 2 times.
in the feature we may want to change to something like `crbidi` compare to `bidi` to signify that it's chrome bidi over other bidi package.
CC @thiagowfx

### DIFF
--- a/src/bidiTab/bidiTab.ts
+++ b/src/bidiTab/bidiTab.ts
@@ -107,12 +107,7 @@ function createCdpConnection() {
     }
   }
 
-  return new CdpConnection(
-    new WindowCdpTransport(),
-    (...messages: unknown[]) => {
-      log(LogType.cdp, ...messages);
-    }
-  );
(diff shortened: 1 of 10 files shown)


OrKoN (approved):

src/bidiTab/bidiTab.ts
```diff
@@ -107,12 +107,7 @@ function createCdpConnection() {
     }
   }
 
-  return new CdpConnection(
-    new WindowCdpTransport(),
-    (...messages: unknown[]) => {
-      log(LogType.cdp, ...messages);
```
sadym-chromium, line 113: The first argument is used to define a container on the Mapper Debug page, where to put the log. This change split CDP and BiDi send and receive messages into different containers.

PR #892, opened 2023-06-26
feat: make the BiDi+ events less "chatty"

Lightning00Blade:
Currently we if we want to get the CDP event we can only pass the `cdp` param. 
That will make all the event from CDP pass though the WebDriver BiDi socket connection.
We want to send only the events that the user is interested for Puppeteer `Tracing.tracingComplete`.
for that I remove the send all `cdp` and replace it with individual `cdp.Tracing.tracingComplete`.
That way only this specific event is send. 

Ref Puppeteer https://github.com/puppeteer/puppeteer/pull/10453 and logs:
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,14 +30,14 @@ CdpSendCommandCommand = {
 }
 
 CdpSendCommandParameters = {
-   cdpMethod: text,
-   cdpParams: any,
-   cdpSession?: text,
+   method: text,
+   params: any,
+   cession?: text,
 }
 
 CdpSendCommandResult = {
    result: any,
-   cdpSession: text,
+   session: text,
 }
 ```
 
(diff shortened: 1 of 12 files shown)


src/bidiMapper/domains/context/cdpTarget.ts
```diff
@@ -154,14 +154,19 @@ export class CdpTarget {
   }
 
   #setEventListeners() {
-    this.#cdpClient.on('*', (cdpMethod, params) => {
+    this.#cdpClient.on('*', (event, params) => {
+      // We may encounter uses for EventEmitter
+      // other then CDP events we want to skip them
```
OrKoN, line 159: could you be more specific about what uses and why we use the event check for that?

OrKoN:
tests seem to be failing

src/bidiMapper/domains/context/cdpTarget.ts
Lightning00Blade: In Puppeteer we inject our own CDP implementation that uses their event emitter. On the connection part we send a  `this.emit(ConnectionEmittedEvents.Disconnected);` this is a Symbol that tries to be converted to String in the literal and fails. An another option would be to use `cdp.${String(event))`

OrKoN (approved):

tests/test_network.py
```diff
@@ -94,7 +94,7 @@ async def test_network_specific_context_subscription_does_not_enable_cdp_network
     new_context_id = await create_context()
 
     # Subscribe to all CDP events.
```
thiagowfx, line 96: Comment is outdated

README.md
```diff
@@ -30,14 +30,14 @@ CdpSendCommandCommand = {
 }
 
 CdpSendCommandParameters = {
-   cdpMethod: text,
-   cdpParams: any,
-   cdpSession?: text,
+   method: text,
+   params: any,
+   cession?: text,
```
thiagowfx, line 35: typo

thiagowfx:
LGTM overall, but one downgrade of this PR is that it's not possible to subscribe to all CDP events anymore.

thiagowfx:
I sent one follow up, https://github.com/GoogleChromeLabs/chromium-bidi/pull/899, which would have been comments in this PR had I reviewed it. PTAL.

Issue #868, opened 2023-06-20
Implement `browsingContext.setViewport`


thiagowfx:
Spec: https://w3c.github.io/webdriver-bidi/#command-browsingContext-setViewport

thiagowfx:
Spec PR was merged yesterday: https://github.com/w3c/webdriver-bidi/pull/425

PR #895, opened 2023-06-27
fix: expand viewport validation tests

thiagowfx:
These are not intended to be exhaustive; for that, we rely on WPT.

Bug: #868


OrKoN:
probably this should have been prefix with `test: `

thiagowfx:
yes, sorry :( - still learning https://www.conventionalcommits.org/

PR #900, opened 2023-06-28
ci(pre-commit): enforce conventionalcommit.org

thiagowfx:


jrandolf (approved):

thiagowfx:
I'll go ahead with this with one disclaimer: if it makes the workflow too annoying then it will be reverted.

OrKoN:
does it validate the PR title (final merge commit)?

thiagowfx:
This is validation for the local workflow only (commit-msg instead of pre-commit).
It is intended to remember people to prepend prefixes right away.

It knows nothing about GitHub PRs. pre-commit.com cannot help here.

I saw there's another presubmit check that was added recently - I suppose that one covers the PR title, at least when it's initially created?

OrKoN:
Yes, that covers the PR and strictly speaking any commit message can be used in the branches

PR #903, opened 2023-06-28
fix: use correct location for key events

jrandolf:


OrKoN (approved):

thiagowfx:
A tiny change that fixes so many tests :D

PR #905, opened 2023-06-28
chore: add script for running WPT on all modes

jrandolf:
This allows maintainers to update expectations across all modes without executing each mode manually.


OrKoN (approved):

thiagowfx:
Please wait for my review on this one. I have a local script that does something similar and we could benefit from merging it with yours

thiagowfx:
PTAL

thiagowfx (approved):

thiagowfx:
Ah, wait. We need a small change to make it work without arguments too.

runWPTAll.sh
```diff
@@ -0,0 +1,27 @@
...
+
+set -uo pipefail
+
+trap ctrl_c INT
+ctrl_c() {
+    exit 1
+}
+
+export LOG_FILE="logs/$(basename "$0").log"
+export UPDATE_EXPECTATIONS="${UPDATE_EXPECTATIONS:-false}"
```
Lightning00Blade, line 12: I have needed to use it yet so, @jrandolf do you think setting update expectation default to true is more in line with what this script will be used for?

jrandolf: @thiagowfx

thiagowfx: I tend to use both values. `false` during development, `true` when I am ready to regenerate the expectations. We can set this to `false`.

PR #911, opened 2023-06-28
chore: add shortcuts

jrandolf:


src/bidiMapper/domains/input/ActionDispatcher.ts
```diff
@@ -569,6 +584,38 @@ export class ActionDispatcher {
...
+          break;
+        case 'KeyC':
+          command = 'Copy';
+          break;
+        case 'KeyV':
+          if (source.shift) {
+            command = 'PasteAndMatchStyle';
+          } else {
+            command = 'Paste';
+          }
```
thiagowfx, line 604: optional:

command = source.shift ? '...' : '...'

ditto below

src/bidiMapper/domains/input/ActionDispatcher.ts
```diff
@@ -69,14 +73,25 @@ async function getElementCenter(
   return {x: x as number, y: y as number};
 }
 
+async function isMacOS(context: BrowsingContextImpl) {
+  const {result} = await (
+    await context.getOrCreateSandbox(undefined)
+  ).callFunction(IS_MAC_DECL, {type: 'undefined'}, [], false, 'none', {});
+  assert(result.type !== 'exception');
+  assert(result.result.type === 'boolean');
```
thiagowfx, line 81: What is the side effect of this assert during runtime? Would the mapper tab crash?

src/bidiMapper/domains/input/ActionDispatcher.ts
```diff
@@ -39,6 +39,10 @@ const CALCULATE_IN_VIEW_CENTER_PT_DECL = ((i: Element) => {
   return [e + ((n - e) >> 1), h + ((m - h) >> 1)];
 }).toString();
 
+const IS_MAC_DECL = (() => {
+  return navigator.platform.toLowerCase().includes('mac');
+}).toString();
```
thiagowfx, line 44: I'd just inline this. If you want to preserve syntax highlighting in text editors, then do:

```
String(() => {
  return ...
})
```

We use this pattern elsewhere

src/bidiMapper/domains/input/ActionDispatcher.ts
jrandolf: It throws an unknown error.

src/bidiMapper/domains/input/ActionDispatcher.ts
jrandolf: Inlining this causes the formatter to modify the declaration's call site drastically. If allowed, I'd prefer to keep it this way consistent with the above decl.

src/bidiMapper/domains/input/ActionDispatcher.ts
```diff
@@ -569,6 +584,38 @@ export class ActionDispatcher {
...
+          command = 'Cut';
+          break;
+        case 'KeyZ':
+          if (source.shift) {
+            command = 'Redo';
+          } else {
+            command = 'Undo';
+          }
+          break;
+        default:
```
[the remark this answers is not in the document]
jrandolf: It won't be exhaustive then and the linter will throw.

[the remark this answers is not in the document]
thiagowfx: Then could you add a comment below?

`// Intentionally left empty` or similar

src/bidiMapper/domains/input/ActionDispatcher.ts
thiagowfx: OK

src/bidiMapper/domains/input/ActionDispatcher.ts
```diff
@@ -69,14 +73,25 @@ async function getElementCenter(
...
 export class ActionDispatcher {
   #tickStart = 0;
   #tickDuration = 0;
   #inputState: InputState;
   #context: BrowsingContextImpl;
+  #isMacOS: Promise<boolean>;
   constructor(inputState: InputState, context: BrowsingContextImpl) {
     this.#inputState = inputState;
     this.#context = context;
+    this.#isMacOS = isMacOS(this.#context);
```
OrKoN, line 94: here an unhandled promise rejection might occur: please add a catch

jrandolf: In what situation?

OrKoN: when the async calls inside isMacOS throw (for example, target shutting down or any other reason really)

PR #913, opened 2023-06-28
chore: miscellaneous changes

thiagowfx:


.github/workflows/publish.yml
```diff
@@ -1,33 +1,33 @@
...
     name: Publish
     runs-on: ubuntu-latest
     steps:
-      - name: Check out repository
-        uses: actions/checkout@96f53100ba2a5449eb71d2e6604bbcd94b9449b5
-      - name: Set up Node.js
-        uses: actions/setup-node@698d50532ea4cc0490274810a45fe52318110a1c
+      - name: Check out repo
+        uses: actions/checkout@v3
+      - name: Set up node.js
```
OrKoN, line 19: node.js is correctly spelled as 'Node.js'

.github/workflows/publish.yml
```diff
@@ -1,33 +1,33 @@
...
   publish:
     name: Publish
     runs-on: ubuntu-latest
     steps:
-      - name: Check out repository
-        uses: actions/checkout@96f53100ba2a5449eb71d2e6604bbcd94b9449b5
-      - name: Set up Node.js
-        uses: actions/setup-node@698d50532ea4cc0490274810a45fe52318110a1c
+      - name: Check out repo
+        uses: actions/checkout@v3
```
OrKoN, line 18: using hashes is a security recommendation. Put the version as a comment next to the hash.

thiagowfx: Better to address this one separately. Lots of manual work to get each individual hash. @jrandolf is there a way to automate this process?

OrKoN: we could copy the hashes from this workflow (or perhaps from Puppeteer) and then let dependabot do the updates

OrKoN: e.g., https://github.com/puppeteer/puppeteer/blob/main/.github/workflows/publish.yml#L19

OrKoN: btw I counted only four unique packages so updating manually (find & replace) should not be too problematic

thiagowfx: There are way more than four, updating them is quite tedious

OrKoN (approved):

PR #912, opened 2023-06-28
test: compare similarity of print to pdf output

thiagowfx:
Bug: #518

PR #914, opened 2023-06-28
test: print to pdf: convert to PNG and compare to golden

thiagowfx:
This is a better approach than https://github.com/GoogleChromeLabs/chromium-bidi/pull/912.

## Tracking

- [x] distinguish headless x headful

Bug: #518


thiagowfx:
This approach works well. We just need to find a way to distinguish headless from headful in pytest, so that we can check for an exception in the headless case.

thiagowfx:
Actually, I thought of a workaround:

We could always try/catch, for the sake of silencing the error in headless.

It's not very elegant, but it would work for the purposes of this test.

OrKoN (approved):

PR #919, opened 2023-06-29
feat: `addScriptToEvaluateOnNewDocument`: run immediately

thiagowfx:
Bug: #293
Doc: [go/chrome-devtools:preload-script-world-name](http://go/chrome-devtools:preload-script-world-name)

## Tracking

- [x] Blocked on https://github.com/GoogleChromeLabs/chromium-bidi/pull/885

### DIFF
--- a/src/bidiMapper/domains/context/bidiPreloadScript.ts
+++ b/src/bidiMapper/domains/context/bidiPreloadScript.ts
@@ -141,10 +141,8 @@ export class BidiPreloadScript {
     }
   }
 
-  /**
-   * Removes the provided cdp target from the list of cdp preload scripts.
-   */
-  cdpTargetIsGone(cdpTargetId: string) {
+  /** Removes the provided cdp target from the list of cdp preload scripts. */
+  dispose(cdpTargetId: string) {
     this.#cdpPreloadScripts = this.#cdpPreloadScripts.filter(
       (cdpPreloadScript) => cdpPreloadScript.target?.targetId !== cdpTargetId
     );
--- a/src/bidiMapper/domains/context/browsingContextProcessor.ts
+++ b/src/bidiMapper/domains/context/browsingContextProcessor.ts
@@ -196,7 +196,7 @@ export class BrowsingContextProcessor {
 
     this.#preloadScriptStorage
       .findPreloadScripts({targetId: contextId})
-      .map((preloadScript) => preloadScript.cdpTargetIsGone(contextId));
+      .map((preloadScript) => preloadScript.dispose(contextId));
   }
 
   #handleTargetInfoChangedEvent(


src/bidiMapper/domains/context/bidiPreloadScript.ts
```diff
@@ -116,6 +116,7 @@ export class BidiPreloadScript {
       'Page.addScriptToEvaluateOnNewDocument',
       {
         source: this.#getEvaluateString(),
+        runImmediately: true,
```
OrKoN, line 119: should not some `Runtime.evaluate` calls be removed with this? Also, not sure if the pinned browser already supports that?

thiagowfx:
Please correct me if I am wrong, but I believe the newly added tests should pass, however they do not:

```
=================================== FAILURES ===================================
__________ test_preloadScript_addGlobally_runImmediately[with reload] __________

websocket = <websockets.legacy.client.WebSocketClientProtocol object at 0x1020d7e10>
html = <function html.<locals>.html at 0x10282dbc0>
context_id = '8FEBE7B1B62C33DEC6BB5996EF2C762B', with_reload = True

    @pytest.mark.asyncio
    @pytest.mark.parametrize("with_reload", [True, False],
                             ids=["with reload", "without reload"])
    async def test_preloadScript_addGlobally_runImmediately(
            websocket, html, context_id, with_reload):
        await execute_command(
            websocket, {
                "method": "browsingContext.navigate",
                "params": {
                    "url": html(),
                    "wait": "complete",
                    "context": context_id
                }
            })

        await execute_command(
            websocket, {
                "params": {
                    "functionDeclaration": "() => { window.foo='bar'; }",
                }
            })

        result = await execute_command(
            websocket, {
                "method": "script.evaluate",
                "params": {
                    "expression": "window.foo",
                    "target": {
                        "context": context_id
                    },
                    "awaitPromise": True,
                    "resultOwnership": "root"
                }
            })
>       assert result["result"] == {"type": "string", "value": 'bar'}
E       AssertionError: assert equals failed
E         {'type': 'undefined'}   {
E                                   'type': 'string',
E                                   'value': 'bar',
E                                 }

tests/test_preload_script.py:527: AssertionError
________ test_preloadScript_addGlobally_runImmediately[without reload] _________

websocket = <websockets.legacy.client.WebSocketClientProtocol object at 0x1014e0ad0>
html = <function html.<locals>.html at 0x1027d9e40>
context_id = 'F8155C92CA2B8C9EF7C0B666E1035A32', with_reload = False

    @pytest.mark.asyncio
    @pytest.mark.parametrize("with_reload", [True, False],
                             ids=["with reload", "without reload"])
    async def test_preloadScript_addGlobally_runImmediately(
            websocket, html, context_id, with_reload):
        await execute_command(
            websocket, {
                "method": "browsingContext.navigate",
                "params": {
                    "url": html(),
                    "wait": "complete",
                    "context": context_id
                }
            })

        await execute_command(
            websocket, {
                "params": {
                    "functionDeclaration": "() => { window.foo='bar'; }",
                }
            })

        result = await execute_command(
            websocket, {
                "method": "script.evaluate",
                "params": {
                    "expression": "window.foo",
                    "target": {
                        "context": context_id
                    },
                    "awaitPromise": True,
                    "resultOwnership": "root"
                }
            })
>       assert result["result"] == {"type": "string", "value": 'bar'}
E       AssertionError: assert equals failed
E         {'type': 'undefined'}   {
E                                   'type': 'string',
E                                   'value': 'bar',
E                                 }

tests/test_preload_script.py:527: AssertionError
=========================== short test summary info ============================
FAILED tests/test_preload_script.py::test_preloadScript_addGlobally_runImmediately[with reload] - AssertionError: assert equals failed
FAILED tests/test_preload_script.py::test_preloadScript_addGlobally_runImmediately[without reload] - AssertionError: assert equals failed
======================= 2 failed, 32 deselected in 0.65s =======================
```

src/bidiMapper/domains/context/bidiPreloadScript.ts
thiagowfx: > Also, not sure if the pinned browser already supports that?

Ah, of course. We will probably need to wait for the roll up.

thiagowfx: > should not some Runtime.evaluate calls be removed with this?

Yes. Done. Was writing a test first

OrKoN (approved):

thiagowfx:
@OrKoN at least one of the WPT tests fail (`  │ FAIL [expected PASS] test_add_preload_script[tab]`), which suggests `runImmediately` doesn't cover all scenarios

OrKoN:
@thiagowfx it looks like the failing test checks that the current context should not have preload script in it (which means that runImmediately is probably set to true in a context where it should be false). Please check the mapper implementation.

src/bidiMapper/domains/context/bidiPreloadScript.ts
OrKoN, line 119: it appears that this runs for existing targets too whereas the spec probably says that it should only apply to the future targets (probably requires checking if this runs for a target that is actually waiting for a debugger)

thiagowfx: Actually, I got it. We run immediately for new targets, but not for pre-existing ones. Will push soon.

thiagowfx: Fixed

thiagowfx:
@OrKoN I had enabled auto-merge but thought it would wait for your review, not sure why it didn't (since I had re-requested it). Anyway, please take a look when you can.

OrKoN:
@thiagowfx no problem, lgtm

PR #922, opened 2023-06-30
test: fix margin.py wpt

thiagowfx:
Bug: #518


thiagowfx:
Comparing with the Firefox implementation, it turns out our implementation is correct. There must be some other error. Closing.

PR #937, opened 2023-07-02
test: update wpt submodules

thiagowfx:
Intended to absorb the commit that fixes a flaky browsingContext.reload test.


Lightning00Blade (approved):

thiagowfx:
Reload tests are still flaky, sadly. Will add expectations to mark them as flaky, accordingly.

We may need to revisit the WPT tests to investigate (or it could just be a problem with our own implementation).

PR #939, opened 2023-07-03
fix: use non-force close for BrowsingContext.close

Lightning00Blade:
BiDi defines that we should run the https://html.spec.whatwg.org/multipage/document-sequences.html#close-a-top-level-traversable, which tells us to run the unload events.
Should we also provide a way to `force close` in BiDi spec?

### DIFF
--- a/src/bidiMapper/domains/context/browsingContextImpl.ts
+++ b/src/bidiMapper/domains/context/browsingContextImpl.ts
@@ -353,7 +353,6 @@ export class BrowsingContextImpl {
         this.#url = params.url;
         this.#deferreds.Page.navigatedWithinDocument.resolve(params);
 
-        // TODO: Remove this once History event for BiDi are added
         this.#eventManager.registerEvent(
           {
             method: BrowsingContext.EventNames.FragmentNavigated,
@@ -827,5 +826,9 @@ export class BrowsingContextImpl {
         data: result.data,
       },
     };
+  }
+
+  async close(): Promise<void> {
+    await this.#cdpTarget.cdpClient.sendCommand('Page.close');
   }
 }
--- a/src/bidiMapper/domains/context/browsingContextProcessor.ts
+++ b/src/bidiMapper/domains/context/browsingContextProcessor.ts
@@ -510,37 +510,51 @@ export class BrowsingContextProcessor {
   async process_browsingContext_close(
     commandParams: BrowsingContext.CloseParameters
   ): Promise<Message.EmptyResult> {
-    const browserCdpClient = this.#cdpConnection.browserClient();
-
     const context = this.#browsingContextStorage.getContext(
       commandParams.context
     );
+
     if (!context.isTopLevelContext()) {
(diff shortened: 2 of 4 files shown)


src/bidiMapper/domains/context/browsingContextImpl.ts
```diff
@@ -790,4 +790,13 @@ export class BrowsingContextImpl {
       },
     };
   }
+
+  async close(): Promise<void> {
+    if (!this.isTopLevelContext()) {
+      throw new Message.InvalidArgumentException(
+        'A top-level browsing context cannot be closed.'
```
OrKoN, line 797:
```suggestion
        'Non top-level browsing context ${context.id} cannot be closed.'
```

OrKoN (approved):

src/bidiMapper/domains/context/browsingContextImpl.ts
thiagowfx: This is resolved but it wasn't push. I just want to make sure it's addressed

Lightning00Blade: Ops forced pushed over it.

src/bidiMapper/domains/context/browsingContextProcessor.ts
```diff
@@ -502,37 +502,11 @@ export class BrowsingContextProcessor {
...
-
-    await browserCdpClient.sendCommand('Target.closeTarget', {
-      targetId: commandParams.context,
-    });
 
-    // Sometimes CDP command finishes before `detachedFromTarget` event,
-    // sometimes after. Wait for the CDP command to be finished, and then wait
-    // for `detachedFromTarget` if it hasn't emitted.
-    await detachedFromTargetPromise;
+    await context.close();
```
thiagowfx, line 509: Whoa, what a big simplification!

thiagowfx (approved):

OrKoN (approved):

src/bidiMapper/domains/context/browsingContextImpl.ts
```diff
@@ -828,4 +828,21 @@ export class BrowsingContextImpl {
       },
     };
   }
+
+  async close(): Promise<void> {
+    try {
+      await this.#cdpTarget.cdpClient.sendCommand('Page.close');
+    } catch (error: any) {
+      // Swallow error that arise from the page being destroyed
+      // Example is navigating to faulty SSL certificate
```
thiagowfx, line 837: Is it feasible to add a test for this?

Lightning00Blade: We don't have our own server we can control at this time, we can open an issue and label `local-server` to them once we have the infastructure.

thiagowfx: I started with 2 items, you can add a third one there (maybe as a subitem of the second item)

PR #752, opened 2023-05-22
Avoid deadlock in Target initialization

thiagowfx:
When initialising new tab, don't wait for the result of the commands can be blocked by debugger.

Bug: #293

PR #951, opened 2023-07-03
test: preload scripts: add e2e test with user gesture

thiagowfx:
Bug: #293
Co-authored-by: Maksim Sadym <sadym@chromium.org>
Reland-of: #752 (revert: #879)

### DIFF
--- a/tests/test_preload_script.py
+++ b/tests/test_preload_script.py
@@ -603,6 +603,68 @@ async def test_preloadScript_add_runImmediately(websocket, html, context_id,
 
 
 @pytest.mark.asyncio
+async def test_preloadScript_addGlobally_withBlankTargetLink(
+        websocket, context_id, html, read_sorted_messages, get_cdp_session_id):
+    LINK_WITH_BLANK_TARGET = html(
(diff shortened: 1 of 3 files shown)


tests/test_preload_script.py
```diff
@@ -602,6 +602,68 @@ async def test_preloadScript_add_runImmediately(websocket, html, context_id,
...
+            }
+        })
+
+    await subscribe(websocket, ["log.entryAdded"])
+
+    session_id = await get_cdp_session_id(context_id)
+
+    command_id = await send_JSON_command(
+        websocket, {
+            "method": "cdp.sendCommand",
```
OrKoN, line 637: why do we send a CDP command instead of a BiDi command? Is it for the gesture flag? perhaps a todo pointing to https://github.com/w3c/webdriver-bidi/issues/359 would be worthwhile

thiagowfx:
This is the last piece to close preload scripts, but it will be hard to submit it as is. Let me try to split it.

tests/test_preload_script.py
thiagowfx: Yes! Will do

PR #956, opened 2023-07-04
test: pytest adjustments to deal with timeouts

thiagowfx:
Docs: https://pypi.org/project/pytest-timeout/

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -6,4 +6,4 @@ testpaths = tests
 
 # https://pypi.org/project/pytest-timeout/
 # timeout in seconds
-timeout = 5
+timeout = 10
--- a/pytest.ini
+++ b/pytest.ini
@@ -3,3 +3,7 @@
 addopts = --doctest-modules
 minversion = 7.0
 testpaths = tests
+
+# https://pypi.org/project/pytest-timeout/
+# timeout in seconds
+timeout = 5
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -4,5 +4,4 @@ pillow==10.0.0
 pytest==7.4.0
 pytest-asyncio==0.21.0
 pytest-icdiff==0.6
-pytest-sugar==0.9.7
 websockets==11.0.3
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -4,4 +4,5 @@ pillow==10.0.0
 pytest==7.4.0
 pytest-asyncio==0.21.0
 pytest-icdiff==0.6
+pytest-timeout==2.1.0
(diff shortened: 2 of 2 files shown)


thiagowfx:
Motivation: https://github.com/GoogleChromeLabs/chromium-bidi/pull/885

E2E takes the full 20 minutes set by the github action

pytest.ini
```diff
@@ -3,3 +3,6 @@
 addopts = --doctest-modules
 minversion = 7.0
 testpaths = tests
+
+# pytest-timeout
+timeout = 60
```
OrKoN, line 8: could you please add the units as a comment

OrKoN: if it is seconds, let's use a value like 5?

thiagowfx: Yes, a smaller value makes sense. I was thinking of 10 or so. It can also be 5. The only reason I started with 60 is to be conservative and not disrupt the CI too much in case it proved to be too low.

I'll decrease it to 10 then.

OrKoN: I would start with 5 and increase if needed

OrKoN (approved):

thiagowfx:
5.0s keeps failing `test_input_performActionsEmitsPointerEvents`.

We could increase the timeout only for that test, but let me instead increase the global timeout to 10 for now.

thiagowfx:
Note that previously there was no timeout at all (at the pytest level).

src/bidiMapper/domains/events/EventManager.ts
```diff
@@ -210,8 +209,8 @@ export class EventManager implements IEventManager {
     eventName: Session.SubscriptionRequestEvent,
     contextId: CommonDataTypes.BrowsingContext | null
   ) {
-    // Enable network domain if user subscribed to any of network events.
-    if (eventName.startsWith(EventManager.#NETWORK_DOMAIN_PREFIX)) {
+    // Enable network domain if user subscribed to any network event.
```
OrKoN, line 212: why these changes are in this PR? (looks like added by mistake)

OrKoN:
`test_input_performActionsEmitsPointerEvents`: judging by the name it probably should not take too long. Perhaps it relies on high fixed timeouts?

OrKoN:
`test_input_performActionsEmitsPointerEvents`: judging by the name it probably should not take too long. Perhaps it relies on high fixed timeouts?

src/bidiMapper/domains/events/EventManager.ts
thiagowfx: ooops, committed to the wrong branch

PR #963, opened 2023-07-05
chore: remove non-standard `context` attribute

jrandolf:


Lightning00Blade (approved):

tests/test_preload_script.py
```diff
@@ -19,27 +19,6 @@
                           send_JSON_command, subscribe)
 
 
-@pytest.mark.asyncio
-async def test_preloadScript_add_nonExistingContext_exceptionReturned(
```
thiagowfx, line 23: `context_id` should also be removed from preload script calls in other tests in this file

PR #964, opened 2023-07-05
fix: Network Module clogging Processing Queue

Lightning00Blade:
Recently we changed to use `Page.close` (#939) which does not destroy the target immediately.
In puppeteer some times we test timeouts of `1ms` which the command is send and failed on Puppeteer end (never awaited). The problem is we try close the page, but not forcing lets the navigation starting and some NetworkEvents to be sent though CDP, but due to being deferred they never resolve and the ProcessingQueue stops working.

### DIFF
--- a/src/bidiMapper/domains/context/browsingContextImpl.ts
+++ b/src/bidiMapper/domains/context/browsingContextImpl.ts
@@ -140,7 +140,7 @@ export class BrowsingContextImpl {
     return this.#loaderId;
   }
 
-  delete() {
+  dispose() {
     this.#deleteAllChildren();
 
     this.#realmStorage.deleteRealms({
@@ -217,7 +217,7 @@ export class BrowsingContextImpl {
   }
 
   #deleteAllChildren() {
-    this.directChildren.map((child) => child.delete());
+    this.directChildren.map((child) => child.dispose());
   }
 
   get #defaultRealm(): Realm {
--- a/src/bidiMapper/domains/network/networkRequest.ts
+++ b/src/bidiMapper/domains/network/networkRequest.ts
@@ -40,7 +40,7 @@ export class NetworkRequest {
    * The identifier for a request resulting from a redirect matches that of the
    * request that initiated it.
    */
-  requestId: Network.Request;
+  readonly requestId: Network.Request;
 
   #servedFromCache = false;
   #redirectCount = 0;
@@ -155,6 +155,12 @@ export class NetworkRequest {
       },
       this.#requestWillBeSentEvent?.frameId ?? null
     );
+  }
+
+  dispose() {
+    const error = new Error('Network processor detached');
+    this.#responseReceivedDeferred.reject(error);
+    this.#beforeRequestSentDeferred.reject(error);
   }
 
   #getBaseEventParams(): Network.BaseParameters {
--- a/src/cdp/cdpClient.ts
+++ b/src/cdp/cdpClient.ts
@@ -30,6 +30,16 @@ export class CloseError extends Error {}
 
 export interface ICdpClient extends EventEmitter<CdpEvents> {
   /**
+   * Unique session identifier.
+   */
+  sessionId: string | undefined;
+
+  /**
+   * Get the default browser client (no sessionId)
+   */
+  browserClient(): ICdpClient;
+
+  /**
    * Provides an unique way to detect if an error was caused by the closure of a
    * Target or Session.
    *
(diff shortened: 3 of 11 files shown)


src/bidiMapper/domains/context/browsingContextImpl.ts
```diff
@@ -320,12 +301,26 @@ export class BrowsingContextImpl {
...
         }
+        const timestamp = BrowsingContextImpl.getTimestamp();
         this.#url = params.frame.url + (params.frame.urlFragment ?? '');
 
         // At the point the page is initialized, all the nested iframes from the
         // previous page are detached and realms are destroyed.
         // Remove children from context.
         this.#deleteAllChildren();
+
+        this.#eventManager.registerEvent(
```
Lightning00Blade, line 312: After some CDP log dumpster diving. It seems this event is emitted before the `Page.frameStartedLoading`

src/bidiMapper/domains/context/cdpTarget.ts
```diff
@@ -146,7 +152,11 @@ export class CdpTarget {
   async enableNetworkDomain() {
     if (!this.#networkDomainEnabled) {
       this.#networkDomainEnabled = true;
-      await NetworkProcessor.create(this.cdpClient, this.#eventManager);
+      await NetworkProcessor.create(
+        this.#cdpConnection,
+        this.cdpClient,
```
OrKoN, line 157: let's add a connection getter on the cdpClient?

src/bidiMapper/domains/network/networkRequest.ts
```diff
@@ -153,6 +153,12 @@ export class NetworkRequest {
     );
   }
 
+  delete() {
```
OrKoN, line 156: let's call it `dispose`?

Lightning00Blade: I did it as there was precedent already for delete rather then dispose. OK to change. Should I rename the `BrowsingContextImpl.delete` as well? @thiagowfx

src/cdp/cdpClient.ts
```diff
@@ -62,6 +67,10 @@ export class CdpClient extends EventEmitter<CdpEvents> implements ICdpClient {
     this.#sessionId = sessionId;
   }
 
+  get sessionId(): string | undefined {
```
thiagowfx, line 70: sessionId is a public variable of the interface.

Here we are creating a private method in the class, setting it, and returning it. Then we don't use the public one.

Could you please reconcile it in some way? Either use the public one, or remove it.

src/cdp/cdpClient.ts
```diff
@@ -29,6 +29,11 @@ export type CdpEvents = {
 export class CloseError extends Error {}
 
 export interface ICdpClient extends EventEmitter<CdpEvents> {
+  /**
+   * Unique session identifier
+   */
```
thiagowfx, line 34:
```suggestion
   * Unique session identifier.
   */
```

(consistency)

src/bidiMapper/domains/script/channelProxy.ts
```diff
@@ -16,13 +16,13 @@
...
 
-import {CommonDataTypes, Script} from '../../../protocol/protocol.js';
+import {type CommonDataTypes, Script} from '../../../protocol/protocol.js';
 import type {IEventManager} from '../events/EventManager.js';
 import {uuidv4} from '../../../utils/uuid';
 
 import type {Realm} from './realm.js';
 
-import Handle = CommonDataTypes.Handle;
+type Handle = CommonDataTypes.Handle;
```
thiagowfx, line 25: What change does this yield? Is it a no-op?

src/bidiMapper/domains/network/networkRequest.ts
thiagowfx: Not very opinionated on this. Just please be globally consistent.

thiagowfx (changes requested):

src/cdp/cdpClient.ts
Lightning00Blade: This is not a method it's a getter so that it can't be modified, the interface is correctly implemented this way as well.

src/bidiMapper/domains/script/channelProxy.ts
Lightning00Blade: It's a no-op, brakes some building (esbuild) and does nothing but add alias. To be honest don't know how it works at all. But it's just a short name, should be in the import statement, will update.

OrKoN (approved):

PR #965, opened 2023-07-05
chore: add fast builds option

Lightning00Blade:
Useful for me when I am prototyping or fixing some Puppeteer issue as it ignores types.

### DIFF
--- a/esbuild.mjs
+++ b/esbuild.mjs
@@ -1,0 +1,12 @@
+import {build} from 'esbuild';
+import {sync} from 'glob';
+
+build({
+  entryPoints: sync('./src/**/*.ts'),
+  outbase: './src',
+  outdir: './lib/cjs',
+  platform: 'browser',
+  sourcemap: true,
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -35,6 +36,9 @@
         "tsc"
       ]
     },
+    "build:fast": {
+      "command": "node ./esbuild.mjs"
```
thiagowfx, line 40: Why add it to wire it if it's a simple command with no deps? Is it because of --watch?

thiagowfx (approved):

esbuild.mjs
```diff
@@ -0,0 +1,12 @@
+import {build} from 'esbuild';
```
thiagowfx, line 1: missing copyright

Lightning00Blade:
After some thought. I think I can come for a solution to will workaround from Puppeteer side better.
That way we will avoid adding another dependency for a single use case.

Issue #874, opened 2023-06-20
Adopt `cddl2ts` and make BiDi protocol sync to upstream in a type-safe fashion


thiagowfx:
Precedent: https://github.com/webdriverio/webdriverio/pull/10447
Previous try: https://github.com/GoogleChromeLabs/chromium-bidi/pull/761
~~Tool: https://github.com/christian-bromann/cddl2ts~~

## Tracking

- [x] ~~Introduce a submodule for https://github.com/w3c/webdriver-bidi~~
- [x] Adopt ~~`cddl2ts`~~ `cddlconv`: https://github.com/GoogleChromeLabs/chromium-bidi/pull/961
- [x] Make AST adjustments to integrate with specifics of our codebase, based on https://github.com/webdriverio/webdriverio/pull/10447: https://github.com/GoogleChromeLabs/chromium-bidi/pull/969
- [x] Add a github action to periodically regenerate the protocol based on the upstream CDDL: https://github.com/GoogleChromeLabs/chromium-bidi/commit/d69091d3a27935c57126bf172519869c087ac4b4
- [x] Auto-generate `zod` parsing: https://github.com/GoogleChromeLabs/chromium-bidi/pull/971

thiagowfx:
https://github.com/GoogleChromeLabs/chromium-bidi/pull/961

PR #968, opened 2023-07-06
ci: add WebDriverBidi type generation action

jrandolf:
Bug: #874


.github/workflows/update-bidi-types.yml
```diff
@@ -0,0 +1,67 @@
...
+    # Run hourly at xx:30.
+    - cron: '30 * * * *'
+  workflow_dispatch:
+
+jobs:
+  build:
+    name: Build WebDriverBidi types
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository
```
thiagowfx, line 16: I was initially thinking to use a submodule here, just like how we do with `wpt/`, for pinning purposes.

And then have the type generation be a pre-requisite step of e.g. `npm run build` (via `wireit`).

Have you considered this approach?

.github/workflows/update-bidi-types.yml
```diff
@@ -0,0 +1,67 @@
...
+      - name: Set up Rust
+        run: rustup update stable && rustup default stable
+      - name: Install cddlconv
+        run: cargo install --git https://github.com/jrandolf/cddlconv.git cddlconv
+      - name: Generate CDDL types
+        run: ./scripts/test.sh
+      - name: Generate TypeScript types
+        run: cddlconv all.cddl > webdriver-bidi.ts
+      - name: Upload WebDriverBidi types
+        uses: actions/upload-artifact@65d862660abb392b8c4a3d1195a2108db131dd05
```
thiagowfx, line 29: add version in the comment

.github/workflows/update-bidi-types.yml
```diff
@@ -0,0 +1,67 @@
...
+  pr:
+    name: Send PR
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
+      - name: Remove old WebDriverBidi types
+        run: rm src/protocol/webdriver-bidi.ts
+      - name: Download WebDriverBidi types
+        uses: actions/download-artifact@e9ef242655d12993efdcda9058dee2db83a2cb9b
```
thiagowfx, line 42: add version

.github/workflows/update-bidi-types.yml
```diff
@@ -0,0 +1,67 @@
...
+      - name: Generate CDDL types
+        run: ./scripts/test.sh
+      - name: Generate TypeScript types
+        run: cddlconv all.cddl > webdriver-bidi.ts
+      - name: Upload WebDriverBidi types
+        uses: actions/upload-artifact@65d862660abb392b8c4a3d1195a2108db131dd05
+        with:
+          name: webdriver-bidi-ts
+          path: webdriver-bidi.ts
+  pr:
```
thiagowfx, line 33: Are you missing a `needs : build` here? Otherwise how are the jobs connected?

thiagowfx (changes requested):

.github/workflows/update-bidi-types.yml
```diff
@@ -0,0 +1,67 @@
...
+        with:
+          repository: w3c/webdriver-bidi
+      - name: Set up Rust
+        run: rustup update stable && rustup default stable
+      - name: Install cddlconv
+        run: cargo install --git https://github.com/jrandolf/cddlconv.git cddlconv
+      - name: Generate CDDL types
+        run: ./scripts/test.sh
+      - name: Generate TypeScript types
+        run: cddlconv all.cddl > webdriver-bidi.ts
```
thiagowfx, line 27: It would be nice to have the ability to generate types locally (for debugging, and as an escape hatch).

Consider moving the `cddlconv` and `test.sh` commands to a script in `package.json`.

.github/workflows/update-bidi-types.yml
jrandolf: I've considered this approach, but submoduling will low-key encourage modifying the spec locally which is something we don't want done. We can go either way, but we can settle with this solution for now and switch later if necessary.

thiagowfx: OK. I am fine with this solution for the time being but IMHO at first glance the submodule approach seems better. I could change my mind over time though, let's see.

Not intending to block this approach, but would like to see at least another supporter.

cc @OrKoN @Lightning00Blade

.github/workflows/update-bidi-types.yml
jrandolf: This is related to the above. See the comment.

thiagowfx: I think it's independent-ish (it creates more work as in one needs to manually clone the upstream repository, but having the ability to regenerate the types is still useful).

OrKoN (approved):

.github/workflows/update-bidi-types.yml
```diff
@@ -0,0 +1,68 @@
...
+      - name: Set up Node.js
+        uses: actions/setup-node@64ed1c7eab4cce3362f8c340dee64e5eaeef8f7c # v3.6.0
+        with:
+          node-version: 18
+          cache: npm
+      - uses: google/wireit@f3a3c79c553122e2fe5829eeac7d815326502903 # setup-github-actions-caching/v1
+      - name: Install and build npm dependencies
+        run: npm ci
+      - name: Run formatter
+        run: npm run format
```
Lightning00Blade, line 56: I think this exits with non 0 when there are changes please verify.

.github/workflows/update-bidi-types.yml
```diff
@@ -0,0 +1,68 @@
+# This workflow will update the WebdriverBidi types.
+
+name: Update WebdriverBidi types
+
+on:
+  schedule:
+    # Run hourly at xx:30.
+    - cron: '30 * * * *'
```
Lightning00Blade, line 8: Too spammy, types change infrequently and we have a manual trigger so I would put it at 1 time per day.

Lightning00Blade (approved):
LGTM + The only thing missing is Local generation -> either README update or a script, but happy to review that in follow up PR.

.github/workflows/update-bidi-types.yml
thiagowfx: Once per day seems reasonable, it would align with WPT. Set it to midnight, maybe.

.github/workflows/update-bidi-types.yml
```diff
@@ -0,0 +1,65 @@
+# This workflow will update the WebdriverBidi types.
+
+name: Update WebdriverBidi types
```
thiagowfx, line 3:
```suggestion
name: Update WebDriverBidi types
```

(very small nit - but to align with your own addition below)

thiagowfx (approved):

PR #969, opened 2023-07-06
chore: add local type generation

jrandolf:
Bug: #874


OrKoN (changes requested):
let's just accept a path to a local cddl file instead of adding submodules?

OrKoN:
E.g., `npm run bidi-types ~/src/webdriver-bidi/all.cddl`

src/protocol/webdriver-bidi.ts
```diff
@@ -51,7 +53,9 @@ export type EventData =
   | LogEvent
   | NetworkEvent
   | ScriptEvent;
-export type Extensible = Record<string, unknown>;
+export type Extensible = {
+  [key: string]: any;
```
thiagowfx, line 57: If possible (if compilation doesn't fail), `unknown` would be preferred.

https://google.github.io/styleguide/tsguide.html#any

jrandolf: Yea, I'll update the parser to use unknown.

tools/generate-bidi-types.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+outfile=src/protocol/webdriver-bidi.ts
```
thiagowfx, line 3:
```suggestion
readonly OUTFILE="src/protocol/webdriver-bidi.ts"
```

https://google.github.io/styleguide/shellguide.html#constants-and-environment-variable-names

Please update `outfile -> OUTFILE` below as well.

tools/generate-bidi-types.sh
```diff
@@ -0,0 +1,34 @@
...
+
+cat <<EOF > $outfile
```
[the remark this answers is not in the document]
thiagowfx: Same below.

tools/generate-bidi-types.sh
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * THIS FILE IS AUTOGENERATED. Run \`npm run bidi-types\` to regenerate.
+ * @see https://github.com/w3c/webdriver-bidi/blob/master/index.bs
+ */
+
+EOF
+
+cddlconv "$1" >> $outfile
+
+# Format twice because the first format doesn't get everything.
```
thiagowfx, line 32: Should we call `npm run format` instead, to decrease coupling between tools?

thiagowfx: Also note, there used to be a `npm run prettier` that was removed. It's fine to reintroduce it in case you dislike `npm run format` (which invokes all formatters).

package.json
```diff
@@ -29,6 +30,12 @@
     ]
   },
   "wireit": {
+    "bidi-types": {
+      "command": "tools/generate-bidi-types.sh",
```
thiagowfx, line 34: If we specify the input file in `files`, then this would work with `--watch`. Maybe it's overkill and only useful when one is iterating on the CDDL, commenting for completeness anyway.

thiagowfx (commented):
This is my final review.

tools/generate-bidi-types.sh
jrandolf: Using `npm run format` is overkill since it runs all the precommits. We only need prettier. I think it's safe to leave it like this; we don't want to expose prettier in `package.json` because that may encourage use of it.

package.json
jrandolf: @OrKoN wants this command to just be a utility command for easily regenerating on our end. The anticipated workflow is:

 - You have `w3c/webdriver-bidi` cloned somewhere.
 - You generate the CDDL document there.
 - You run `npm run bidi-types -- path/to/all.cddl`

This decouples generation with building and allows us to use a custom CDDL when needed.

tools/generate-bidi-types.sh
thiagowfx: Here's the problem: The way it is currently, it forces the user to have `prettier` locally installed and in `$PATH`.

If you do it via `npm`, then there is no need to have it locally installed, nor in the `$PATH`. `npm` properly includes `node_modules/.bin`.

jrandolf: Hmm, then let me use `npx`

thiagowfx: That's also fine. And it will guarantee we use the same version as in `package.json`.

thiagowfx (approved):

OrKoN:
what's happening in this PR?

Issue #876, opened 2023-06-21
Make tooling cross-platform (Linux, macOS, Windows)


OrKoN:
Currently, tools like runWPT.sh are not cross-platform as they use tools (e.g., nproc) that is not available everywhere by default. I suggest we re-write tools in JavaScript to make them work cross-platform.

### Platforms

- [x] Linux
- [x] macOS
- [ ] Windows: https://github.com/GoogleChromeLabs/chromium-bidi/pull/1043

### Convert scripts from bash to node (`js`) or `ts`

- [x] `runBiDiServer.sh`: https://github.com/GoogleChromeLabs/chromium-bidi/pull/1039
- [x] `runWPT.sh`: https://github.com/GoogleChromeLabs/chromium-bidi/pull/1039
- [ ] `runWPTAll.sh`

thiagowfx:
That's reasonable

thiagowfx:
Updated the first comment to reflect tracking.

PR #970, opened 2023-07-06
ci: npm run e2e runs local CfT

Lightning00Blade:
Bug: #876

### DIFF
--- a/install-browser.mjs
+++ b/install-browser.mjs
@@ -28,14 +28,23 @@
  *  - `node install-browser.mjs /tmp/cache`
  */
 
-import fs from 'fs/promises';
+import {readFile} from 'fs/promises';
+import {homedir} from 'os';
+import {resolve} from 'path';
 
 import {install, computeExecutablePath} from '@puppeteer/browsers';
 import actions from '@actions/core';
 
+const SHELL_ARG = '--shell';
+
 try {
-  const browserSpec = (await fs.readFile('.browser', 'utf-8')).trim();
-  const cacheDir = process.argv[2] || process.cwd();
+  const browserSpec = (await readFile('.browser', 'utf-8')).trim();
+
+  let cacheDir = resolve(homedir(), '.cache/chromium-bidi');
+  if (process.argv[2] && process.argv[2] !== SHELL_ARG) {
+    cacheDir = process.argv[2];
+  }
+
   // See .browser for the format.
   const browser = browserSpec.split('@')[0];
   const buildId = browserSpec.split('@')[1];
@@ -49,7 +58,9 @@ try {
     browser,
     buildId,
   });
-  actions.setOutput('executablePath', executablePath);
+  if (!process.argv.includes(SHELL_ARG)) {
+    actions.setOutput('executablePath', executablePath);
+  }
   console.log(executablePath);
 } catch (err) {
   actions.setFailed(`Failed to download the browser: ${err.message}`);
--- a/runBiDiServer.sh
+++ b/runBiDiServer.sh
@@ -19,7 +19,14 @@ if [[ $# -gt 0 && ("$1" == "-h" || "$1" == "--help") ]]; then
 fi
 
 # The chrome release channel to use: `stable`, `beta`, `canary`, `dev`.
-readonly CHANNEL="${CHANNEL:-dev}"
+# `local` means to use the browser version specified in `.browser`.
+CHANNEL="${CHANNEL:-local}"
+
+if [ "$CHANNEL" == "local" ]; then
+  export BROWSER_BIN="$(node install-browser.mjs --shell)"
+  # Need to pass valid CHANNEL to the command below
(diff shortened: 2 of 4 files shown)


runBiDiServer.sh
```diff
@@ -19,7 +19,13 @@ if [[ $# -gt 0 && ("$1" == "-h" || "$1" == "--help") ]]; then
 fi
 
 # The chrome release channel to use: `stable`, `beta`, `canary`, `dev`.
```
thiagowfx, line 21: The comment is now outdated. Add:

```
# ... `dev`, `local`.
# `local` means to use the browser version specified in `.browser`.
```

runBiDiServer.sh
```diff
@@ -19,7 +19,13 @@ if [[ $# -gt 0 && ("$1" == "-h" || "$1" == "--help") ]]; then
...
-readonly CHANNEL="${CHANNEL:-dev}"
+CHANNEL="${CHANNEL:-local}"
+
+# Run the pinned CfT Cannary by default
```
thiagowfx, line 24: Given the suggestion above, this comment can be removed.

runBiDiServer.sh
```diff
@@ -19,7 +19,13 @@ if [[ $# -gt 0 && ("$1" == "-h" || "$1" == "--help") ]]; then
...
+if [ $CHANNEL == "local" ]; then
```
thiagowfx, line 25:
```suggestion
if [ "$CHANNEL" == "local" ]; then
```

runBiDiServer.sh
```diff
@@ -19,7 +19,13 @@ if [[ $# -gt 0 && ("$1" == "-h" || "$1" == "--help") ]]; then
...
+  export BROWSER_BIN=$(node install-browser.mjs ~/.cache/chromium-bidi --shell)
```
thiagowfx, line 26:
```suggestion
  export BROWSER_BIN="$(node install-browser.mjs $HOME/.cache/chromium-bidi --shell)"
```

runBiDiServer.sh
```diff
@@ -19,7 +19,13 @@ if [[ $# -gt 0 && ("$1" == "-h" || "$1" == "--help") ]]; then
...
+  CHANNEL='canary'
```
thiagowfx, line 27:
```suggestion
  CHANNEL='dev'
```

Since canary does not exist on Linux, please keep defaulting to `dev` here.

runBiDiServer.sh
thiagowfx, line 26: It would be cleaner if we did not hard-code any cache in this file. Can we omit it?

Instead, either keep using `cwd`, or update `install-browser.mjs` to use `.cache`, that's where it should belong.

runBiDiServer.sh
Lightning00Blade: This change is correct we pass it down into some parser that expects the only `stable`, `dev` etc.
So `canary` is the closes correct one.

runBiDiServer.sh
Lightning00Blade: OK i will take a stab, but it will require some more changes.

install-browser.mjs
```diff
@@ -49,7 +49,9 @@ try {
     browser,
     buildId,
   });
-  actions.setOutput('executablePath', executablePath);
+  if (process.argv.indexOf('--shell') === -1) {
```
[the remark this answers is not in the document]
Lightning00Blade: Sure, no problem for me, note this is `mjs` and not `ts` files :smile:

runBiDiServer.sh
thiagowfx: OK, I see what you mean. Sorry, disregard.

Ideally I would eliminate `CHANNEL` but let's keep this out of scope of this PR.

install-browser.mjs
```diff
@@ -28,14 +28,23 @@
...
 import actions from '@actions/core';
 
+const SHELL_ARG = '--shell';
+
 try {
-  const browserSpec = (await fs.readFile('.browser', 'utf-8')).trim();
-  const cacheDir = process.argv[2] || process.cwd();
+  const browserSpec = (await readFile('.browser', 'utf-8')).trim();
+
+  let cacheDir = resolve(homedir(), '.cache/chromium-bidi');
```
thiagowfx, line 43: Maybe add this comment:

```
XXX: Consider https://www.npmjs.com/package/cache-or-tmp-directory or similar for portability.
```

This is for https://github.com/GoogleChromeLabs/chromium-bidi/issues/876. Windows is not a fan of directories that start with a period AFAIK.

No need to change anything here, code is great as is.

thiagowfx (approved):

install-browser.mjs
Lightning00Blade: We should be OK about this one Git creates `.git` and that seems to work fine.

PR #975, opened 2023-07-06
ci: run install before building types

jrandolf:


.github/workflows/update-bidi-types.yml
```diff
@@ -31,24 +31,32 @@ jobs:
     steps:
       - name: Check out repository
         uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
+      - name: Set up Python
+        uses: actions/setup-python@bd6b4b6205c4dbad673328db7b31b7fab9e241c0 # v4.6.1
+        with:
+          python-version: '3.11'
+          cache: pip
+      - name: Install pre-commit
+        uses: pre-commit/action@v3.0.0
```
thiagowfx, line 40: Add a comment here because it's not obvious:

```
# Needed for `npm run format`.
```

.github/workflows/update-bidi-types.yml
```diff
@@ -31,24 +31,32 @@ jobs:
...
-          name: all-cddl
-      - name: Generate TypeScript types
-        run: npm run bidi-types -- all.cddl
       - name: Set up Node.js
         uses: actions/setup-node@64ed1c7eab4cce3362f8c340dee64e5eaeef8f7c # v3.6.0
         with:
           node-version: 18
           cache: npm
-      - uses: google/wireit@f3a3c79c553122e2fe5829eeac7d815326502903 # setup-github-actions-caching/v1
+      - name: Set up cache
```
thiagowfx, line 50:
```suggestion
```

All other actions do not provide a name for the wireit cache action.

Just remove this one - but if you're really strong about keeping it then please add it everywhere where `google/wireit` is used. Consistency ;-)

.github/workflows/update-bidi-types.yml
```diff
@@ -31,24 +31,32 @@ jobs:
     steps:
       - name: Check out repository
         uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
+      - name: Set up Python
+        uses: actions/setup-python@bd6b4b6205c4dbad673328db7b31b7fab9e241c0 # v4.6.1
```
thiagowfx, line 35: I understand you added python because of `npm run format` which runs `pre-commit` which runs `yapf`. But ultimately this workflow doesn't care about python at all. Adding this dependency here feels wrong.

There are multiple ways to fix this. If you're not keen on fixing it now then please add a comment above:

```
XXX: Remove python dependency (used indirectly by `yapf` via `npm run format`.
```

thiagowfx (approved):

.github/workflows/update-bidi-types.yml
jrandolf: `pre-commit` also comes directly from Python.

PR #980, opened 2023-07-07
ci: install pre-commit after npm packages

jrandolf:


OrKoN (approved):

.github/workflows/update-bidi-types.yml
```diff
@@ -31,14 +31,11 @@ jobs:
     steps:
       - name: Check out repository
         uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
-        # XXX: Remove python dependency (used indirectly by `yapf` via `npm run format`).
```
thiagowfx, line 34: > pre-commit also comes directly from Python.

(just noting because it's not obvious)

PR #984, opened 2023-07-07
chore: miscellaneous refactoring

thiagowfx:


tools/generate-bidi-types.sh
```diff
@@ -1,7 +1,10 @@
 #!/bin/bash
 
-TYPES_FILE=src/protocol/webdriver-bidi.ts
-ZOD_FILE=src/protocol-parser/webdriver-bidi.ts
+set -euo pipefail
+
+readonly TYPES_FILE="src/protocol/webdriver-bidi.ts"
```
jrandolf, line 5: I'd rather keep these at the bottom to be consistent

thiagowfx: Done

PR #991, opened 2023-07-10
chore: remove network interception

jrandolf:


thiagowfx (approved):

thiagowfx:
Apparently this is introducing some breaking changes to `print`

thiagowfx:
You could copy the validation implementation we previously had with zod, for printing

Issue #949, opened 2023-07-03
Support new headless in nodeJS runner


thiagowfx:
Currently, we NodeJS mapper runner supports only support old headless.

sadym-chromium:
`--HEADLESS=[false, new, old]`

PR #997, opened 2023-07-10
ci(headless): use new headless instead of old headless

thiagowfx:
Closes: #949

I chose to completely replace old headless. Is there any use for it anymore?

Alternatively we could support all of them. That would require changing more files. And would pose the question whether we would want another CI config (new headless x old headless).

### DIFF
--- a/runWPT.sh
+++ b/runWPT.sh
@@ -63,7 +63,7 @@ if [ -z ${WPT_METADATA+x} ]; then
 fi
 
 if [[ "$HEADLESS" == "true" ]]; then
-  log "Running WPT in headless mode..."
+  log "Running WPT in new headless mode..."
 else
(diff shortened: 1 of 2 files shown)


thiagowfx:
WPT failures are legitimate.

OrKoN:
I would say testing the new headless would make us better match the headful. I think we can just replace it but I wonder what level of support would we need for old headless and WebDriver BiDi.

thiagowfx:
> `E       webdriver.bidi.error.UnknownErrorException: unknown error ('Page.printToPDF' wasn't found)`

thiagowfx:
This will not be merged anytime soon. Closing (pingback added to https://github.com/GoogleChromeLabs/chromium-bidi/issues/949).

PR #998, opened 2023-07-10
test(screenshot): update wpt expectations

thiagowfx:
Bug: #514


thiagowfx:
Meh. I had sent this because it passed in a previous PR but now it fails again. There's a race condition with the automation bar. Never mind.

PR #1002, opened 2023-07-11
feat: implement browsingContext.activate

OrKoN:
WPT: https://github.com/web-platform-tests/wpt/pull/40964
Spec: https://github.com/w3c/webdriver-bidi/pull/453

### DIFF
--- a/src/bidiMapper/domains/context/browsingContextProcessor.ts
+++ b/src/bidiMapper/domains/context/browsingContextProcessor.ts
@@ -300,6 +300,19 @@ export class BrowsingContextProcessor {
     );
   }
 
+  async process_browsingContext_activate(
+    params: BrowsingContext.ActivateParameters
+  ): Promise<EmptyResult> {
+    const context = this.#browsingContextStorage.getContext(params.context);
+    if (!context.isTopLevelContext()) {
+      throw new InvalidArgumentException(
+        'Activation is only supported on the top-level context'
+      );
+    }
+    await context.activate();
+    return {};
+  }
+
   async process_browsingContext_captureScreenshot(
     params: BrowsingContext.CaptureScreenshotParameters
   ): Promise<BrowsingContext.CaptureScreenshotResult> {
--- a/tests/test_activate.py
+++ b/tests/test_activate.py
@@ -1,0 +1,44 @@
+# Copyright 2023 Google LLC.
+# Copyright (c) Microsoft Corporation.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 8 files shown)


src/bidiMapper/domains/context/browsingContextProcessor.ts
```diff
@@ -300,6 +300,19 @@ export class BrowsingContextProcessor {
     );
   }
 
+  async process_browsingContext_activate(
+    params: BrowsingContext.ActivateParameters
+  ): Promise<EmptyResult> {
+    const context = this.#browsingContextStorage.getContext(params.context);
+    if (!context.isTopLevelContext()) {
+      throw new InvalidArgumentException(
+        'Emulating viewport is only supported on the top-level context'
```
thiagowfx, line 309: nit: I know it's still a draft, but just FYI, this error message is too specific. It should not mention viewport (e.g. we could be calling `activate` for screenshot purposes).

OrKoN: yeah, I just copy pasted it from the other command, still in progress.

thiagowfx:
TODO for you (either in this PR or in the next one, but keep it in mind): Update

https://github.com/GoogleChromeLabs/chromium-bidi/blob/cfa01d1b2ec867e2b47d49f20f932f477631a411/src/bidiMapper/domains/context/browsingContextImpl.ts#L719

...to use the bidi API instead of calling the CDP command directly.

thiagowfx:
LGTM (adding one test would be nice).

jrandolf (approved):
+1 to adding a test

OrKoN:
Added a basic test (more extensive test coverage will be done by WPT)

tests/test_activate.py
thiagowfx, line 1: I would have suggested to add this to the browsing context test file.

The top-level directory for tests/* becomes a bit messy if we put everything here.

Alternatively, we could mirror WPT and create a top-level browsing_context dir, and move this therein.

(Feel free to re-request review before submitting, I tend to review fast and would have suggested this in advance.)

OrKoN: That sounds like a separate issue. Putting everything into the single browsing context test file is messy too.

thiagowfx: Do you think it's worth to create many subdirectories, like WPT? We could start to slowly migrate our codebase to a similar pattern.

OrKoN: I think subdirectories is the way to go where the directory name is the module name. I think we should not create a directory per command yet (a file per command/event should be okay)

PR #1017, opened 2023-07-12
chore: enable mouse wheel tests

jrandolf:
Tests are passing now because we activate the main tab before running input.


OrKoN (changes requested):
chromium binary committed (pls remove)

OrKoN (approved):

PR #1025, opened 2023-07-13
ci: only use "wpt run --processes" in CI

thiagowfx:
When regenerating WPT tests locally, `--processes` is quite flaky. Connections keep closing and stack traces keep being printed.

Local builds should default to not running in parallel, which is better for debugging and stability.

CI will keep using the flag for speed/performance reasons, as the job currently takes 10min on average instead of 20min.

Bug: #876

### DIFF
--- a/runWPT.sh
+++ b/runWPT.sh
@@ -86,7 +86,7 @@ if [[ "$VERBOSE" == "true" ]]; then
     --log-mach -
     --log-mach-level info
   )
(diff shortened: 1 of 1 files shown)


Lightning00Blade (approved):

jrandolf:
Closed in favor of https://github.com/GoogleChromeLabs/chromium-bidi/pull/1039

PR #1036, opened 2023-07-13
ci: wpt: exclude tentative tests from report

thiagowfx:
"Tentative" means a WPT test was written even though consensus was never
reached in the spec whether it will actually take place.

Tentative tests will usually never pass and they are not final.
Since these are not spec finalized, they only add noise to the report.

Once (if ever) consensus is reached upon upstream, the tentative suffix
is removed from the WPT test file, then the tests appear in the report
as usual.

chore: misc changes
test: add unit test
docs: add jsdoc

### DIFF
--- a/tools/wpt-report-builder/builder.mjs
+++ b/tools/wpt-report-builder/builder.mjs
@@ -14,7 +14,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 import fs from 'fs';
 
 import {generateReport} from './formatter.mjs';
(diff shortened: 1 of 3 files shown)


Lightning00Blade:
Test fail + leave a comment what these test are, and why exclude them.

thiagowfx:
> Test fail + leave a comment what these test are, and why exclude them.

I did but it was buried in the second commit message. Copied it to the PR.

thiagowfx:
Example that will be removed from the report:

image

Lightning00Blade (approved):

PR #1039, opened 2023-07-14
chore: use JavaScript for scripts

jrandolf:


package.json
```diff
@@ -223,6 +220,7 @@
     "websocket": "1.0.34",
     "wireit": "0.10.0",
     "ws": "8.13.0",
+    "yargs": "^17.7.2",
```
Lightning00Blade, line 223: Pinned

.github/workflows/wpt.yml
```diff
@@ -114,7 +114,7 @@ jobs:
       - name: Set up Node.js
         uses: actions/setup-node@64ed1c7eab4cce3362f8c340dee64e5eaeef8f7c # v3.6.0
         with:
-          node-version: 16
+          node-version: 18
```
Lightning00Blade: Make it use latest LTS with this
```suggestion
          node-version: lts/*
```

package.json
```diff
@@ -109,9 +106,9 @@
       ]
     },
     "server-headless": {
-      "command": "./runBiDiServer.sh --headless=true",
+      "command": "./tools/run-bidi-server.mjs --headless=true",
       "files": [
-        "runBiDiServer.sh"
+        "tools/run-bidi-server.mjs"
```
Lightning00Blade, line 111: Be consistent here or down.
```suggestion
        "./tools/run-bidi-server.mjs"
```

tools/run-wpt.mjs
```diff
@@ -0,0 +1,184 @@
...
+}
+
+function usage() {
+  log(
+    `Usage: [CHROMEDRIVER=<true | false>] [HEADLESS=<true | false>] [UPDATE_EXPECTATIONS=<true | false>] ${process.argv[1]} [webdriver/tests/bidi/[...]]`
+  );
+}
+
+if (
+  process.argv.length > 2 &&
```
Lightning00Blade, line 38: It would be better to look at all args rather then only the second.

tools/run-wpt.mjs
```diff
@@ -0,0 +1,184 @@
...
+) {
+  usage();
+  process.exit(0);
+}
+
+const isLinux = process.platform === 'linux';
+const isMacOS = process.platform === 'darwin';
+
+// The path to the browser binary.
+const BROWSER_BIN =
```
Lightning00Blade, line 49: @thiagowfx should we run this with the pinned version if not specified as well?

tools/run-wpt.mjs
```diff
@@ -0,0 +1,184 @@
...
+
+if (VERBOSE === 'true') {
+  wptRunArgs.push(
+    '--debug-test',
+    '--log-mach',
+    '-',
+    '--log-mach-level',
+    'info'
+  );
+} else if (HEADLESS === 'true') {
```
Lightning00Blade, line 129: Please update to use the part from this PR https://github.com/GoogleChromeLabs/chromium-bidi/pull/1025

.github/workflows/wpt.yml
jrandolf: Obsolete.

tools/run-wpt.mjs
jrandolf: Considering this only takes one argument right now, we can migrate to yargs for this script later.

Lightning00Blade: It's an `.includes` statement. No need for yargs

package.json
```diff
@@ -95,10 +95,7 @@
       ]
     },
     "server-headful": {
-      "command": "./runBiDiServer.sh --headless=false",
-      "files": [
-        "runBiDiServer.sh"
-      ],
+      "command": "npm run server-headless -- --headless=false",
```
thiagowfx, line 98: If it works, `npm run server --headless=false` seems less odd.

package.json
```diff
@@ -31,7 +31,7 @@
   },
   "wireit": {
     "bidi-types": {
-      "command": "tools/generate-bidi-types.sh",
+      "command": "./tools/generate-bidi-types.sh",
```
thiagowfx, line 34: There's no need for `./` here, as it was previously. Same for other commands under `tools/`.

tools/run-bidi-server.mjs
```diff
@@ -0,0 +1,107 @@
...
+
+process.chdir(dirname(dirname(process.argv[1])));
+
+function log(message) {
+  // eslint-disable-next-line no-console
+  console.log(`(${basename(process.argv[1])}) ${message}`);
+}
+
+const argv = yargs(hideBin(process.argv))
+  .usage(
```
thiagowfx, line 35: Can `yargs` generate the `usage` string automatically? It was always a chore to maintain this.

Although, I'm skeptical because these are env vars, not command switches.

tools/run-bidi-server.mjs
```diff
@@ -0,0 +1,107 @@
...
+    `[CHANNEL=<stable | beta | canary | dev>] [DEBUG=*] [DEBUG_COLORS=<yes | no>] [HEADLESS=<true | false>] [LOG_DIR=logs] [NODE_OPTIONS=--unhandled-rejections=strict] [PORT=8080] $0`
+  )
+  .option('headless', {
+    describe:
+      'The --headless flag takes precedence over the HEADLESS environment variable.',
```
thiagowfx, line 40: The doc should be:

"# Whether to start the server in headless or headful mode."

As it was previously. You can prepend the string above to the current one.

tools/run-bidi-server.mjs
```diff
@@ -0,0 +1,107 @@
...
+    type: 'boolean',
+    default: process.env.HEADLESS === 'true',
+  })
+  .parse();
+
+let CHANNEL = process.env.CHANNEL || 'local';
+let BROWSER_BIN = process.env.BROWSER_BIN;
```
thiagowfx, line 47: Move BROWSER_BIN one line above, so that it isn't in the middle of the CHANNEL logic.

tools/run-bidi-server.mjs
```diff
@@ -0,0 +1,107 @@
...
+if (CHANNEL === 'local') {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  BROWSER_BIN = spawnSync('node', ['tools/install-browser.mjs', '--shell'])
+    .stdout.toString()
+    .trim();
+  // Need to pass valid CHANNEL to the command below
+  CHANNEL = 'canary';
+}
+
+const DEBUG = process.env.DEBUG || '*';
```
thiagowfx, line 58: This `DEBUG` statement is incorrect. In bash, note that it is `${DEBUG-*}` instead of `${DEBUG:-*}`. This means that we must allow DEBUG to be empty.

I believe the correct node change is to replace `||` with `??`.

tools/run-bidi-server.mjs
```diff
@@ -0,0 +1,107 @@
...
+  [
+    'lib/cjs/bidiServer/index.js',
+    `--channel`,
+    CHANNEL,
+    `--headless`,
+    argv.headless,
+    ...process.argv.slice(2),
+  ],
+  {
+    stdio: ['inherit', 'pipe', 'pipe'],
```
thiagowfx, line 83: two pipes? What does this mean?

tools/run-bidi-server.mjs
```diff
@@ -0,0 +1,107 @@
...
+    env: {
+      ...process.env,
+      DEBUG,
+      DEBUG_COLORS,
+      NODE_OPTIONS,
+      PORT,
+      BROWSER_BIN,
```
thiagowfx, line 90: :sort

thiagowfx (changes requested):
First batch of comments, for "run bidi server". Still need to review the WPT file.

package.json
jrandolf: It doesn't work ;)

tools/run-bidi-server.mjs
jrandolf: Yea, because of the env, we can't ;/

OrKoN: I would be in favour of converting to proper args in the future (but not in this PR)

OrKoN: yargs does generate it automatically

tools/run-bidi-server.mjs
jrandolf: The above is `in`, `out`, `err`. We want to pipe `out` and `err`

package.json
jrandolf: Other places use this as well. I'm not sure if we need to change it right now. I'm going to add Windows CI so we can validate it then.

tools/run-bidi-server.mjs
```diff
@@ -0,0 +1,110 @@
...
+if (CHANNEL === 'local') {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  BROWSER_BIN = spawnSync('node', ['tools/install-browser.mjs', '--shell'])
+    .stdout.toString()
+    .trim();
+  // Need to pass valid CHANNEL to the command below
+  CHANNEL = 'canary';
+}
+
+const DEBUG = process.env.DEBUG ?? '*';
```
thiagowfx, line 59: Maybe preserve the comment from the bash script here, otherwise there's the risk future refactoring will break it.

```
// DEBUG= (empty string) is allowed.
```

OrKoN (approved):

tools/run-bidi-server.mjs
OrKoN: why do we need an empty string?

jrandolf, line 59:
```suggestion
// DEBUG = (empty string) is allowed.
const DEBUG = process.env.DEBUG ?? '*';
```

thiagowfx: DEBUG is set to '*' by default. If you do not want any DEBUG, you set it to empty.

tools/run-wpt.mjs
thiagowfx: Yes, I think it's a reasonable default to set.

thiagowfx:
LGTM

Issue #562, opened 2023-03-21
Handle CDP serialization (WebDriver classic collections, Node and `internalId`)


sadym-chromium:
- Upstream: https://github.com/w3c/webdriver-bidi/issues/318
- Spec: https://w3c.github.io/webdriver-bidi/#data-types-protocolValue-RemoteValue-serialization
- Umbrella chromium issue: https://bugs.chromium.org/p/chromium/issues/detail?id=1420968

## Tracking

- [x] Array and String iterators (needs a change in V8): https://bugs.chromium.org/p/v8/issues/detail?id=14206
- [x] Array-like
  - [x] `NodeList`
  - [x] `HTMLCollection`
- [x] `internalId`: Add support for `internalId` in serialization to handle circles and duplicated references.
- [x] Node serialization (https://github.com/w3c/webdriver-bidi/pull/335) failed the following tests: `/webdriver/tests/bidi/script/call_function/result_node.py.ini` and `/webdriver/tests/bidi/script/evaluate/result_node.py.ini`

sadym-chromium:
closed as implemented

PR #1042, opened 2023-07-14
feat: support iterator serialization

thiagowfx:
Bug: #562
Tested: Works for Map and Set iterators. Array and String iterators will need a separate change in V8.

### DIFF
--- a/src/bidiMapper/domains/script/realm.ts
+++ b/src/bidiMapper/domains/script/realm.ts
@@ -110,13 +110,16 @@ export class Realm {
       }
     }
 
-    if (
-      cdpValue.result.type === 'object' &&
-      cdpValue.result.subtype === 'generator'
-    ) {
-      return {
-        type: 'generator',
-      };
+    if (cdpValue.result.type === 'object') {
+      switch (cdpValue.result.subtype) {
+        case 'generator':
+        case 'iterator':
+          return {
+            type: cdpValue.result.subtype,
+          };
+        default:
+        // Intentionally left blank.
+      }
     }
 
     return bidiValue;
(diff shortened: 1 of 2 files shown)


OrKoN (approved):

src/bidiMapper/domains/script/realm.ts
```diff
@@ -110,13 +110,16 @@ export class Realm {
...
-      cdpValue.result.type === 'object' &&
-      cdpValue.result.subtype === 'generator'
-    ) {
-      return {
-        type: 'generator',
-      };
+    if (cdpValue.result.type === 'object') {
+      switch (cdpValue.result.subtype) {
+        case 'generator':
+        case 'iterator':
```
Lightning00Blade, line 116: Nit: Add TODO about Array/String iterator

thiagowfx: This code is going to remain the same though. The TODO should be tracked in the bug (or in V8).

What I could do is add an unit test that is disabled and has a TODO.

thiagowfx: Added tests + TODO

src/bidiMapper/domains/script/realm.ts
```diff
@@ -110,13 +110,16 @@ export class Realm {
...
     }
 
-    if (
-      cdpValue.result.type === 'object' &&
-      cdpValue.result.subtype === 'generator'
-    ) {
-      return {
-        type: 'generator',
-      };
+    if (cdpValue.result.type === 'object') {
```
sadym-chromium, line 113: This check is  done only for the top-level objects, but not for the nested one. Traversing nested data is done in [`#deepSerializedToBiDi`](https://github.com/GoogleChromeLabs/chromium-bidi/pull/1042/files#diff-d7af9adf42c7a4f4040462b9ff23cae43df9de6a355ee3197df6d9462485c1c3R93), but there is no `type` and `subtype` data.

PR #1043, opened 2023-07-14
ci: add windows

jrandolf:
Bug: #876

### DIFF
--- a/tools/run-wpt.mjs
+++ b/tools/run-wpt.mjs
@@ -19,6 +19,8 @@
 
 import {spawnSync} from 'child_process';
 import {cpus} from 'os';
+import {join, resolve, sep} from 'path';
+
 import {packageDirectorySync} from 'pkg-dir';
 
 // Changing the current work directory to the package directory.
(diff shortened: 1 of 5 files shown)


tools/run-wpt.mjs
```diff
@@ -43,17 +45,35 @@ if (
...
-  (isLinux
-    ? '/usr/bin/google-chrome-unstable'
-    : isMacOS
-    ? '/Applications/Google Chrome Dev.app/Contents/MacOS/Google Chrome Dev'
-    : '');
+let BROWSER_BIN = process.env.BROWSER_BIN;
+if (!BROWSER_BIN) {
+  switch (process.platform) {
+    case 'linux':
+      BROWSER_BIN = join(sep, 'usr', 'bin', 'google-chrome-unstable');
```
thiagowfx, line 52: More readable, no?

      BROWSER_BIN = '/usr/bin/google-chrome-unstable';

thiagowfx: Same for macOS.

thiagowfx: It's just odd to use both `join` and `sep` at the same time. The fact you put `sep` at the beginning paramounts to hard-coding the fact Unix absolute paths start with `/`. So you might as well completely remove `sep`.

thiagowfx (approved):

tools/run-wpt.mjs
jrandolf: This is more programmatic which is why I chose it. Since this is more versatile, I'll just keep it as it.

thiagowfx:
Did you force-merge? Windows E2E is failing.

jrandolf:
Windows E2E is not required so it merged.

thiagowfx:
If they do not work, it's better to remove them for now. Even if they are non-blocking, they make the overall PR status be "X" (fail / red), which is misleading.

Issue #644, opened 2023-04-25
Specify and Implement `Network` interception


sadym-chromium:
- Upstream: https://github.com/w3c/webdriver-bidi/issues/66
- Roadmap: https://github.com/w3c/webdriver-bidi/blob/master/roadmap.md#replacing-resources-with-test-data

### Prototyping and scaffolding by Thiago
- [x] #2034

### Productionisation of Network Interception by Nikolay
- [ ] #2035

Lightning00Blade:
Closing to be tracked in separate (specific) bug.

PR #1050, opened 2023-07-17
feat: scaffold network request interception

thiagowfx:
Supersedes #957.

Bug: #644
DevTools: https://source.chromium.org/chromium/chromium/src/+/main:content/browser/devtools/protocol/fetch_handler.cc
Doc: [go/webdriver-bidi:network-request-interception-proposal](http://go/webdriver-bidi:network-request-interception-proposal)
Spec: https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/429.html#network-intercepts
Puppeteer: https://github.com/puppeteer/puppeteer/blob/main/test/src/requestinterception.spec.ts
Upstream: https://github.com/w3c/webdriver-bidi/pull/429


thiagowfx:
Next: Create a Network Storage class.

thiagowfx:
Next: Scaffold "Add" and "Remove" intercepts, proxying to `Fetch.enable` and `Fetch.disable`. Beware: enable is not additive, we'll need to keep track of previous additions.

Next: Scaffold unit / E2E tests for both of these operations.

src/bidiMapper/domains/network/NetworkProcessor.ts
```diff
@@ -0,0 +1,68 @@
...
+
+import {NetworkStorage} from './NetworkStorage.js';
+
+export class NetworkProcessor {
+  #networkStorage: NetworkStorage = new NetworkStorage();
+
+  process_network_addIntercept(
+    _params: Network.AddInterceptParameters
+  ): Network.AddInterceptResult {
+    return {
```
Lightning00Blade, line 27: Throw error would be better that way it are spec complaint, then we can do thing steps one by one.

src/bidiMapper/domains/network/NetworkProcessor.ts
```diff
@@ -0,0 +1,68 @@
...
+ * limitations under the License.
+ */
+import type {Network, EmptyResult} from '../../../protocol/protocol.js';
+
+import {NetworkStorage} from './NetworkStorage.js';
+
+export class NetworkProcessor {
+  #networkStorage: NetworkStorage = new NetworkStorage();
+
+  process_network_addIntercept(
```
Lightning00Blade, line 24: Drop the process_network_ we are in the networkProccessor its doubling the text. Also I already did that for the other Processors so we should be consistent.

src/protocol-parser/webdriver-bidi.ts
```diff
@@ -108,7 +108,9 @@ export const ErrorCodeSchema = z.lazy(() =>
     'no such element',
```
Lightning00Blade, line 108: Like discussed offline I think we should create a separate file next to this one with the types generated from the CDDL for the Network Interception PR.

src/bidiMapper/domains/network/NetworkStorage.ts
```diff
@@ -0,0 +1,47 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import type {Network} from '../../../protocol/protocol.js';
+
+export class NetworkStorage {
+  /** A map to define the properties of active network intercepts. */
```
Lightning00Blade, line 20: When implementing this further see how it ties back to the #1058

src/bidiMapper/domains/network/NetworkManager.ts
```diff
@@ -111,40 +100,40 @@ export class NetworkManager {
...
       }
     );
 
-    return networkProcessor;
+    return networkManager;
   }
 
   dispose() {
-    for (const request of this.#requestMap.values()) {
+    for (const request of this.#networkStorage.requestMap.values()) {
```
Lightning00Blade, line 121: There should be a method I on the networkStorage that does the dispose and here we should just call that.

thiagowfx: Fair. Done.

thiagowfx:
@Lightning00Blade FYI: Two methods are partially implemented: add intercept and remove intercept.

src/bidiMapper/domains/network/NetworkProcessor.ts
```diff
@@ -0,0 +1,94 @@
...
+    const urlPatterns: string[] = params.urlPatterns ?? [];
+    const parsedPatterns: string[] = [];
+    for (const urlPattern of urlPatterns) {
+      const parsed = urlPattern; // TODO: Parse the pattern.
+      parsedPatterns.push(parsed);
+    }
+
+    // TODO: call CDP `Fetch.enable`.
+
+    this.#networkStorage.interceptMap.set(intercept, {
```
OrKoN, line 48: let's not expose maps directly, it would be really beneficial if we wrap map access

OrKoN (approved):

thiagowfx:
There's one part still missing: To have `addIntercept` call `Fetch.enable`.

src/bidiMapper/domains/network/NetworkStorage.ts
thiagowfx: Not being used yet, let's consider it for future PRs.

thiagowfx:
This is now ready to go!

thiagowfx:
I am going to write a couple more unit tests before submitting, but will not add new features to the implementation in this PR.

OrKoN (approved):

src/bidiMapper/domains/context/cdpTarget.ts
```diff
@@ -92,13 +98,33 @@ export class CdpTarget {
...
+    );
+  }
+
+  /** Calls `Fetch.disable`. */
+  async fetchDisable() {
+    await this.#cdpClient.sendCommand('Fetch.disable');
+  }
+
+  /**
+   * Calls `Fetch.disable` followed by `Fetch.enable`.
```
Lightning00Blade, line 120: Can we have a line explaining why we need this.

src/protocol-parser/webdriver-bidi.ts
thiagowfx: Not applicable anymore, given that the upstream PR was merged.

src/protocol/webdriver-bidi.ts
```diff
@@ -592,20 +594,58 @@ export namespace BrowsingContext {
...
-export type NetworkResult = Record<string, never>;
+export type NetworkCommand =
+  | Network.AddIntercept
+  | Network.ContinueRequest
+  | Network.ContinueResponse
+  | Network.ContinueWithAuth
+  | Network.FailRequest
+  | Network.ProvideResponse
+  | Network.RemoveIntercept;
+export type NetworkResult = {};
```
Lightning00Blade, line 605: Can you manually change this to `Record<string,never>` this will cause issue with typings else . The latest CDDL parser has that change.

src/bidiMapper/domains/network/NetworkManager.ts
```diff
@@ -22,129 +22,118 @@
...
-    cdpClient
+    cdpTarget.cdpClient
       .browserClient()
       .on(
         'Target.detachedFromTarget',
         (params: Protocol.Target.DetachedFromTargetEvent) => {
-          if (cdpClient.sessionId === params.sessionId) {
-            networkProcessor.dispose();
+          if (cdpTarget.cdpClient.sessionId === params.sessionId) {
+            networkManager.#networkStorage.disposeRequestMap();
```
Lightning00Blade, line 58: This I am not sure as from my understanding it will clear all requests. This will work if a single page is open but what if we open 2 pages? Happy to merge if the test pass but lets add comment or issue to investigate further.

src/protocol/webdriver-bidi.ts
thiagowfx: Not applicable anymore, given:

```
export type NetworkResult = Network.AddInterceptResult;
```

src/bidiMapper/domains/network/NetworkManager.ts
thiagowfx: This diff is a no-op. I just switched to cdpTarget.cdpClient instead of cdpClient; and to use the method within network storage, otherwise nothing else changed.

I can add a comment anyway.

src/bidiMapper/domains/context/cdpTarget.ts
thiagowfx: Done

Lightning00Blade (approved):

thiagowfx:
For the sake of unblocking PRs such as https://github.com/GoogleChromeLabs/chromium-bidi/pull/1177 and https://github.com/GoogleChromeLabs/chromium-bidi/pull/1178, this should be merged ASAP.

I believe I addressed the E2E and WPT failures. Setting auto merge.

Note: Url pattern parsing (+tests) will be deferred to another PR.

PR #1058, opened 2023-07-18
fix: don't hold finished requests in memory

Lightning00Blade:


src/bidiMapper/domains/network/networkProcessor.ts
```diff
@@ -125,6 +133,10 @@ export class NetworkProcessor {
     return this.#requestMap.get(requestId);
   }
 
+  #forgetRequest(requestId: Network.Request): void {
+    this.#requestMap.delete(requestId);
```
Lightning00Blade, line 137: Needs to dispose the request first...

OrKoN (approved):

PR #1059, opened 2023-07-18
chore: extract Script commands into separate processor

Lightning00Blade:
As per our offline discussion we should split all the logic not to be handled by the browingContextProcessor and have one processor per Module.


OrKoN (approved):

thiagowfx:
LGTM (I made a few tiny style changes that I will send soon in a PR)

Issue #1080, opened 2023-07-24
Network and Fetch domain activation


thiagowfx:
Currently, the network domain is incorrectly enabled.

The network domain should be enabled via the CDP client/target, only when the browsing context has...

- ...subscriptions to `network.` events, and/or
- ..._global_ subscription to the network namespace

It's important to not do `Network.enable` globally, otherwise we'd emit too many unnecessary CDP events, when the client may not be interested in them.

Edit(2023-09-11): @Lightning00Blade: It seems that if we never enable interception we always call `Fetch.disable` on target create, I think it would be good it we omit that that way its a little bit faster for WPT and other test.

cc @Lightning00Blade

sadym-chromium:
> cc @sadym-chromium FYI as this slightly relates to supporting multiple BiDi/CDP sessions.

ack

thiagowfx:
`Fetch` is partially resolved via #1409.

Lightning00Blade:
This should be resolved and we should be enabling both `Network` and `Fetch` only when needed.

PR #1082, opened 2023-07-24
chore: globally enable the Network domain per CDP Target

thiagowfx:
Bug: #644, #1080


src/bidiMapper/domains/context/cdpTarget.ts
```diff
@@ -102,22 +101,21 @@ export class CdpTarget {
    */
   async #unblock() {
     try {
+      NetworkProcessor.create(this.cdpClient, this.#eventManager);
```
thiagowfx, line 104: This seems out of place now.

thiagowfx: Ideally it should be moved to CommandProcessor, however once that is done all network tests start to fail.

tests/test_network.py
```diff
@@ -88,36 +88,6 @@ async def test_network_global_subscription_enabled_in_new_context(
     })
 
 
-@pytest.mark.asyncio
-async def test_network_specific_context_subscription_does_not_enable_cdp_network_globally(
```
Lightning00Blade, line 92: Don't remove but skip and link issue.

src/bidiMapper/domains/context/cdpTarget.ts
Lightning00Blade: I think the class name is misleading we should call it NetworkManager. Similar example is the logManager, where we have no Processor, as there are no commands.

Lightning00Blade (approved):

src/bidiMapper/domains/context/cdpTarget.ts
thiagowfx: Done

PR #1086, opened 2023-07-24
chore: consistently rename files with UpperCamelCase when appropriate

thiagowfx:


thiagowfx:
I borked this PR. Will need to recreate it.

PR #1092, opened 2023-07-25
chore: use Protocol.Target.SessionID/TargetID instead of string for CDP

thiagowfx:


thiagowfx:
There are also some occurrences of `selfTargetId`. Not sure if I should replace those. Thoughts?

thiagowfx:
> I have no context around selfTargetId so can't judge.

Then I'll be conservative and not touch it.

PR #1097, opened 2023-07-25
test: write an unit test for exception classes

thiagowfx:


src/protocol/webdriver-bidi.ts
```diff
@@ -71,7 +71,7 @@ export type JsInt = number;
  * Must be between `0` and `9007199254740991`, inclusive.
  */
 export type JsUint = number;
-export const enum ErrorCode {
+export enum ErrorCode {
```
thiagowfx, line 74: The test caught several issues: 2 missing entries and 1 incomplete entry.

There is one problem: It only works with non-const enums. And this file is auto-generated.

Do you know how I can write the same test with a `const enum`?

thiagowfx: Error message with a const enum:

```
src/protocol/ErrorResponse.spec.ts:23:15 - error TS2475: 'const' enums can only be used in property or index access expressions or the right hand side of an import declaration or export assignment or type query.

23   Object.keys(ErrorCode).forEach((errorCodeKey: string) => {
                 ~~~~~~~~~

src/protocol/ErrorResponse.spec.ts:24:33 - error TS2476: A const enum member can only be accessed using a string literal.

24     const errorCode = ErrorCode[errorCodeKey as keyof typeof ErrorCode];
                                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

src/protocol/ErrorResponse.spec.ts:30:15 - error TS7006: Parameter 'word' implicitly has an 'any' type.

30         .map((word) => word.charAt(0).toUpperCase() + word.slice(1))
                 ~~~~


Found 3 errors.
```

thiagowfx: Or maybe we could just make the generator emit non-`const`?

Lightning00Blade: You can't do it directly as const enums are replaced when building with the value they represent.
What you can do is have a structure 
```ts
const errors: {
  [key in ErrorCode]: string;
} = {
  [ErrorCode.NoElement]: '',
};
```
Typescript will then complain that the list is not exhaustive.
Then just do the same loop with a small change.
```ts
Object.keys(errors).forEach((errorCode) => {
  it(`should have an exception class for ErrorCode.${errorCode}`, () => {
    // Exception class name should be in PascalCase, e.g. InvalidArgumentException
    const exceptionClassName = `${errorCode
      .split(' ')
      .map((word) => word.charAt(0).toUpperCase() + word.slice(1))
      .join('')}Exception`;

    expect(exceptionClasses).to.have.property(exceptionClassName);
    const ExceptionClass = (exceptionClasses as any)[exceptionClassName];

    const instance = new ExceptionClass('my message', 'my stacktrace');
    expect(instance).to.have.property('error', errorCode);
    expect(instance).to.have.property('message', 'my message');
    expect(instance).to.have.property('stacktrace', 'my stacktrace');
  });
});
```

thiagowfx:
x-ref: https://github.com/microsoft/TypeScript/issues/41122

thiagowfx:
@Lightning00Blade PTAL

jrandolf (approved):

PR #1113, opened 2023-07-28
ci: add Puppeteer label

Lightning00Blade:
Similar to the Devtools one we created we should have one for Puppeteer here.

### DIFF
--- a/.github/workflows/puppeteer.yml
+++ b/.github/workflows/puppeteer.yml
@@ -1,0 +1,81 @@
+name: Puppeteer CI
+
+# Declare default permissions as read only.
+permissions: read-all
+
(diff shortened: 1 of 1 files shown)


jrandolf (approved):

jrandolf:
Nice!

thiagowfx:
Does this mean that if we add the puppeteer label to a PR then it will run the puppeteer test suite?

If so: That's cool!

sadym-chromium:
Would you mind to update the readme, or add more details to this CI step explaining what does it do and how to use it?

Lightning00Blade:
Create #1330 to address that.

PR #1115, opened 2023-07-28
build(chrome): fix CfT roll

Lightning00Blade:
Auto-roller overrides the changes every 30 min, that should be fixed in #1114.
Opening these too understand why test are failing.


jrandolf (approved):

jrandolf:
lgtm % passing tests

PR #1116, opened 2023-07-28
feat: implement `browser.close`

Lightning00Blade:


jrandolf (approved):

tests/browser/test_browser_close.py
thiagowfx, line 1: Suggest to rename it to `test_close.py`, as the directory name already contains `browser`. This follows the current pattern in the rest of `tests/`.

tests/browser/test_browser_close.py
```diff
@@ -0,0 +1,23 @@
...
+# limitations under the License.
+
+import pytest
+from test_helpers import execute_command
+
+
+@pytest.mark.asyncio
+async def test_await_close_response(websocket):
+
+    await execute_command(websocket, {"method": "browser.close", "params": {}})
```
thiagowfx, line 23: This is just testing the command is successful. It would be nice to check the response is empty.

(Given the `setTimeout` workaround, this would only make sense iff this assertion does not create flakiness.)

Alternatively: If the above is not feasible, then just add a comment to make it explicit why we're not adding a response assertion (currently the method name suggests it would check the response).

PS. Why `await` in the test name? We almost always run `execute_command` with `await`; it seems redundant.

src/bidiMapper/domains/browser/BrowserProcessor.ts
```diff
@@ -0,0 +1,37 @@
...
+  #cdpConnection: ICdpConnection;
+
+  constructor(cdpConnection: ICdpConnection) {
+    this.#cdpConnection = cdpConnection;
+  }
+
+  close(): EmptyResult {
+    const client = this.#cdpConnection.browserClient();
+
+    // Insure that its put at the last place in the event loop
```
thiagowfx, line 31: typo, suggested fix:

```
// Ensure that it is put at the end of the event loop.
// This way we send back the response before closing the tab.
```

tests/browser/test_browser_close.py
thiagowfx: N/A, apparently pytest cannot have two `test_close.py`, even if they belong to different directories (source: Nik).

PR #1119, opened 2023-07-31
build(wpt): use pinned browser by default

Lightning00Blade:


tools/run-wpt.mjs
```diff
@@ -47,32 +47,16 @@ if (
...
-    default:
-      BROWSER_BIN = '';
-  }
+  const path = execSync(
+    `node ${join('tools', 'install-browser.mjs')} '--shell'`
+  )
+    .toString()
+    .split(sep);
+  // We only need the path but return value
+  // contains the executable as well
```
Lightning00Blade, line 56: @jrandolf can you verify this works on Mac, on Linux Python fails if I give it the path + executable.

jrandolf (approved):

PR #1136, opened 2023-08-02
chore: extract Modules from Event enums

Lightning00Blade:
As per https://w3c.github.io/webdriver-bidi/#obtain-a-set-of-event-names we should consider this as Module rather then All events that way if we need to use an Module name we have the correct reference. 

Note I will need it to fix the network module from always being enabled.


src/protocol/chromium-bidi.ts
```diff
@@ -24,13 +24,22 @@ export type EventNames =
...
-  | Script.EventNames;
+  | Script.EventNames
+  | BiDiModule;
 // keep-sorted end
 
+export enum BiDiModule {
+  // keep-sorted start
+  BrowsingContext = 'browsingContext',
+  Script = 'script',
+  Log = 'log',
```
thiagowfx, line 35: how come Script comes before Log, even with keep sorted?

thiagowfx (approved):

PR #1139, opened 2023-08-03
chore: fix test with same name issue + typo

Lightning00Blade:


.pre-commit-config.yaml
```diff
@@ -105,6 +105,7 @@ repos:
           .* limitations under the License.
         language: pygrep
         files: \.js$|\.cjs$|\.mjs$|\.ts$|\.py$
+        exclude: \__init__.py$
```
thiagowfx, line 108: Why is the backslash needed? `exclude: __init__.py$` should suffice, as (AFAIK) underscore is not a special character in RE2

thiagowfx: Note: Move the backslash to before the period

thiagowfx (approved):

PR #1140, opened 2023-08-03
chore: rename logging

Lightning00Blade:
While trying it use it I found it hard to understand all the different types
also having them with capital letters needs more typing.
Spaces in the log name also behave strange paired with filter.

### DIFF
--- a/src/bidiMapper/domains/context/browsingContextImpl.ts
+++ b/src/bidiMapper/domains/context/browsingContextImpl.ts
@@ -37,6 +37,8 @@ import type {BrowsingContextStorage} from './browsingContextStorage.js';
 import type {CdpTarget} from './cdpTarget.js';
 
 export class BrowsingContextImpl {
+  static loggerPrefix = `${LogType.debug}:browsingContext` as const;
+
   /** The ID of this browsing context. */
   readonly #id: BrowsingContext.BrowsingContext;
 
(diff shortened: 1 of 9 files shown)


src/bidiMapper/domains/context/browsingContextImpl.ts
```diff
@@ -37,6 +37,8 @@ import type {BrowsingContextStorage} from './browsingContextStorage.js';
 import type {CdpTarget} from './cdpTarget.js';
 
 export class BrowsingContextImpl {
+  static loggerPrefix = `${LogType.debug}:browsingContext` as const;
```
thiagowfx, line 40: LOGGER_PREFIX

https://google.github.io/styleguide/tsguide.html#identifiers-constants

Also, maybe make it `static readonly` as well?

thiagowfx (approved):

src/bidiMapper/domains/context/browsingContextImpl.ts
```diff
@@ -37,6 +37,8 @@ import type {BrowsingContextStorage} from './browsingContextStorage.js';
 import type {CdpTarget} from './cdpTarget.js';
 
 export class BrowsingContextImpl {
+  static LOGGER_PREFIX = `${LogType.debug}:browsingContext` as const;
```
thiagowfx, line 40: Assuming not adding `readonly` is intentional. Just curious why?

Lightning00Blade: Its OK to have it, and i forgot, but  that being said when you have `as const` value it will only work if someone reassigns is the exact same string.

Lightning00Blade: image

thiagowfx (approved):

PR #1145, opened 2023-08-03
chore: fail channel loop gracefully

jrandolf:
This PR fails the channel loop when CDP errors as these errors are permanent. The loop will restart from `startListenerFromWindow` in these cases.


src/bidiMapper/domains/script/channelProxy.ts
```diff
@@ -171,64 +171,64 @@ export class ChannelProxy {
...
             },
           },
-        },
-        realm.browsingContextId
-      );
-    }
+          realm.browsingContextId
+        );
+      }
+    } catch {}
```
thiagowfx, line 231: Something is missing here, either some or all of those:

- Rationale for the PR (what problem does it resolve?) --> could just augment the commit message; alternatively (or as well), add a code comment here
- Log the `catch`?
- Why was this comment removed?

```
 // TODO(#294): Remove this loop after the realm is destroyed.
 // Rely on the CDP throwing exception in such a case.
```

thiagowfx:
Adding @Lightning00Blade as I'll be OOO. Don't wait for my  re-review.

src/bidiMapper/domains/script/channelProxy.ts
```diff
@@ -171,63 +174,66 @@ export class ChannelProxy {
...
           },
-        },
-        realm.browsingContextId
-      );
+          realm.browsingContextId
+        );
+      } catch (error) {
+        // If an error is thrown, then the channel is permanently broken, so we
+        // exit the loop.
+        this.#logger?.(LogType.cdp, error);
```
Lightning00Blade, line 234:
```suggestion
        this.#logger?.(LogType.debug, error);
```

src/bidiMapper/domains/script/channelProxy.ts
```diff
@@ -31,8 +32,9 @@ export class ChannelProxy {
   readonly #properties: Script.ChannelProperties;
 
   readonly #id = uuidv4();
+  readonly #logger?: LoggerFn;
 
-  constructor(channel: Script.ChannelProperties) {
+  constructor(channel: Script.ChannelProperties, logger?: LoggerFn) {
     if (
       ![0, null, undefined].includes(channel.serializationOptions?.maxDomDepth)
```
Lightning00Blade, line 39: @thiagowfx Side question is this still needed? I thought the deep serialization options landed.

src/bidiMapper/domains/script/channelProxy.ts
jrandolf: Done.

Lightning00Blade (approved):
LGTM + Is it possible to add a test for this scenario ?

jrandolf:
> Is it possible to add a test for this scenario

Not in a reasonable way since the error is transient.

PR #1146, opened 2023-08-04
chore: refactoring for network interception

thiagowfx:
Extract a few bits from #1050, to keep the other PR focused.


src/bidiMapper/CommandProcessor.ts
```diff
@@ -78,17 +82,18 @@ export class CommandProcessor extends EventEmitter<CommandProcessorEvents> {
       preloadScriptStorage,
       logger
     );
+    this.#cdpProcessor = new CdpProcessor(
+      browsingContextStorage,
+      cdpConnection
+    );
     this.#inputProcessor = InputProcessor.create(browsingContextStorage);
```
Lightning00Blade, line 89: Nit: Can we refactor away the .create we don't do anything async so no need for it, we should just put all things in the constructor

Lightning00Blade (approved):

Lightning00Blade (approved):

PR #1149, opened 2023-08-07
chore: implement injection framework

jrandolf:
This PR implements a decorator-based injection framework for use in Chromium-BiDi. It's expected to replace the parameter hell we've been facing inside the repo.

## Usage

The framework uses three decorators: `@pantry`, `@feed`, and `@eat`.

### Storing

`You` can store things in the `@pantry`:

```ts
class You {
  @pantry(Number)
  #value: number
}
```

`@pantry` takes either a class constructor or a symbol (more on this later).

### Injecting

`You` can `@feed` `Child`.

```ts
class You {
  @pantry(Number)
  #value: number

  @feed([Number])
  #child: Child = new Child()
}

class Child {}
```

### Consuming

`You` can `@eat` from `Parent` anything that `Parent` `@feed`s `You`.

```ts
class Parent {
  @pantry(Number)
  #value: number

  @feed([Number])
  #you: You = new You()
}

class You {
  @eat(Number)
  #value!: number;
}
```

### Deep injection

`You` can `@feed` `Child` anything that `Parent` `@feed`s `You`.

```ts
class Parent {
  @pantry(Number)
  #value: number

  @feed([Number])
  #you: You = new You()
}

class You {
  @feed([Number])
  #child: Child = new Child()
}

class Child {
  @eat(Number)
  #value!: number;
}
```

## Linting

Linting is implemented here: https://github.com/GoogleChromeLabs/chromium-bidi/pull/1150. It's quite comprehensive and any potential "bugs" you may face will be prevented by listening to the lint.

## Symbols

Sometimes it's not possible to use a class name, perhaps because you want an interface rather than a concrete class. For this, we allow symbols. Since the framework is strongly-typed, each symbol needs to be registered with `decorators.js` with its associated type:

```
export interface SomeAbstractInterface {
  // ...
}

export const SomeAbstractInterfaceSym = Symbol('SomeAbstractInterface');

declare module 'path/to/decorators.js' {
  interface InjectableRegistry {
    [SomeAbstractInterfaceSym]: SomeAbstractInterface
  }
}
```

### DIFF
--- a/src/utils/decorators.spec.ts
+++ b/src/utils/decorators.spec.ts
@@ -1,0 +1,544 @@
+/**
+ * Copyright 2023 Google LLC.
+ * Copyright (c) Microsoft Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as chai from 'chai';
+
+import {feed, eat, pantry} from './decorators.js';
+
+const expect = chai.expect;
+
+const SymbolSym = Symbol('symbol');
+const SymbolSym2 = Symbol('symbol2');
+
+declare module './decorators.js' {
+  interface InjectableRegistry {
+    [SymbolSym]: symbol;
+    [SymbolSym2]: symbol;
+  }
+}
+
+describe('Decorators', () => {
+  describe('inject: field, provides: field, consumes: field', () => {
+    it('should work', () => {
+      class A {
+        @pantry(SymbolSym)
+        readonly value = Symbol();
+
+        @feed
+        readonly b = new B();
+      }
+
+      class B {
+        @eat(SymbolSym)
+        readonly value!: symbol;
+      }
+
+      const a = new A();
+      expect(a.b.value).equals(a.value);
+    });
+
+    it('should work out of order', () => {
+      class A {
+        @feed
+        readonly b = new B();
+
+        @pantry(SymbolSym)
+        readonly value = Symbol();
+      }
+
+      class B {
+        @eat(SymbolSym)
+        readonly value!: symbol;
+      }
+
+      const a = new A();
+      expect(a.b.value).equals(a.value);
+    });
+
+    it('should not work with constructors', () => {
+      class A {
+        @pantry(SymbolSym)
+        readonly value: symbol;
+
+        @feed
+        readonly b: B;
+
+        constructor() {
+          this.value = Symbol();
+          this.b = new B();
+        }
+      }
+
+      class B {
+        @eat(SymbolSym)
+        readonly value!: symbol;
+      }
+
+      const a = new A();
+      expect(a.b.value).equals(undefined);
+    });
+
+    it('should not descend modifications', () => {
+      class A {
+        @pantry(SymbolSym)
+        readonly value = Symbol();
+
+        @feed
+        readonly b = new B();
+      }
+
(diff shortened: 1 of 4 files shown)


tsconfig.base.json
```diff
@@ -23,7 +23,7 @@
     "strictFunctionTypes": true,
     "strictNullChecks": true,
     "strictPropertyInitialization": true,
-    "target": "ESNext",
+    "target": "ES2022",
```
OrKoN, line 26: what is the status of decorators in ECMAScript/TypeScript? Does the compiled output run in the browser/node 16+?

src/utils/decorators.spec.ts
```diff
@@ -0,0 +1,509 @@
...
+    [SymbolSym]: symbol;
+    [SymbolSym2]: symbol;
+  }
+}
+
+describe('Decorators', () => {
+  describe('inject: field, provides: field, consumes: field', () => {
+    it('should work', () => {
+      class A {
+        @pantry(SymbolSym)
```
OrKoN, line 38: I think we should have a bigger discussion about this and wait for Thiago to decide on this. I personally not much in favour of using decorators for injecting members as I find it harder to read and more verbose than the regular parameter passing. I think we should discuss it more in the daily (is there a dog about pros/cons) and also consider if this will work in Node 16+.

tsconfig.base.json
jrandolf: Yes, it does!

src/utils/decorators.spec.ts
jrandolf: Sure, we can wait on Thiago.

My two cents: Sure, the approach can be a little verbose, but I can't agree that it's harder to read. A simple proof of this is by comparing the following two files:

Without injection:
https://github.com/GoogleChromeLabs/chromium-bidi/blob/41d16155105b706383ca6d9d6651443bd697b91c/src/bidiMapper/CommandProcessor.ts#L60

With injection:
https://github.com/GoogleChromeLabs/chromium-bidi/blob/9c382d37e89a4137459a72073f009f00edcbd5a9/src/bidiMapper/CommandProcessor.ts#L53

It's very clear from injection what is going into those objects without even looking at the definition of the constructor. If you consider line count, without injection is also about 20% more verbose and that's with only partial injection.

src/utils/decorators.spec.ts
```diff
@@ -0,0 +1,509 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as chai from 'chai';
+
+import {feed, eat, pantry} from './decorators.js';
```
Lightning00Blade, line 20: Is there a reference you can leave that describes how the names a chosen. For me they don't click without without looking at implementation/use.

jrandolf: I've updated the PR description with a usage guide.

OrKoN (commented):
So thinking about it more: I am not convinced that wiring is a big problem, with DI you need to wire things practically in the same way with less transparency with what goes where and using untrue types for the injected members. 

If the type issue can be solved (e.g., by marking injected members optional as they truly are), I would be in favour of using DI for utility classes such as loggers but not the "logic" classes. Instead, we should minimize the number of injections by reducing the number of classes that converge in a single class by improving the class structure.

I am curious about what @thiagowfx and @mathiasbynens think.

mathiasbynens:
This is a big enough change to warrant a proper proposal (doc?) and discussion. What is the motivation, what are some concrete examples of “the parameter hell” you’re referring to, what are the pros and cons, etc.?

jrandolf:
If the type issue can be solved (e.g., by marking injected members optional as they truly are), I would be in favor of using DI for utility classes such as loggers but not the "logic" classes.

+1 It should be remarked however that such a use is "transparent" so the class would work without injection.

jrandolf:
> This is a big enough change to warrant a proper proposal (doc?) and discussion. What is the motivation, what are some concrete examples of “the parameter hell” you’re referring to, what are the pros and cons, etc.?

Yes, let's discuss this in an internal doc.

PR #1167, opened 2023-08-10
chore: way to debug stuck Processing Queue

Lightning00Blade:
This will show the last event to be pick up by the processing queue if the Class gets stuck. 
Encountered a new issue where it's stuck in Puppeteer test:  `Locator should work with OOPIF` , need a way to debug.

### DIFF
--- a/src/utils/log.ts
+++ b/src/utils/log.ts
@@ -20,6 +20,7 @@ export enum LogType {
   bidi = 'bidi',
   cdp = 'cdp',
   debug = 'debug',
+  debugError = 'debug:error',
   // keep-sorted end
 }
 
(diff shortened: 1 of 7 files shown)


src/utils/log.ts
```diff
@@ -20,6 +20,7 @@ export enum LogType {
   bidi = 'bidi',
   cdp = 'cdp',
   debug = 'debug',
+  debugError = 'debug:error',
```
Lightning00Blade, line 23: We should do this for other places as well as currectly you can't listen on all debug events + sub debug at the same time via `DEBUG=<bidi>:debug:*` (this does not include `<bidi>:debug`)

OrKoN (approved):

PR #1184, opened 2023-08-21
fix: filter only sent cookies

### LINKED ISSUE #1011: Cookies should only contain cookies mentioned in the header
See https://github.com/GoogleChromeLabs/chromium-bidi/blob/main/src/bidiMapper/domains/network/networkRequest.ts#L381

It takes all `associatedCookies` that might include cookies that were not sent by the browser (e.g., blocked). The spec says though that only cookies includes in the request should be listed https://w3c.github.io/webdriver-bidi/#ref-for-get-a-cookie:~:text=For%20each%20cookie%20in%20the%20user%20agent%E2%80%99s%20cookie%20store%20that%20are%20included%20in%20request%3A

it sounds like we should at least remove blocked cookies and at most to parse the cookie header and make sure that we only include the cookies from the headers.

Lightning00Blade:
Closes #1011

### DIFF
--- a/src/bidiMapper/domains/network/NetworkRequest.ts
+++ b/src/bidiMapper/domains/network/NetworkRequest.ts
@@ -431,24 +431,26 @@ export class NetworkRequest {
   static #getCookies(
     associatedCookies: Protocol.Network.BlockedCookieWithReason[]
   ): Network.Cookie[] {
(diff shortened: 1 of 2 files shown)


thiagowfx:
It may be difficult, but if feasible, consider adding some test.

OrKoN (approved):

PR #1198, opened 2023-08-22
docs: add date of last update to site

Lightning00Blade:


tools/wpt-report-builder/formatter.mjs
```diff
@@ -133,6 +133,8 @@ function mergeSingleChildren(map) {
 }
 
 export function generateHtml(map) {
+  const date = new Intl.DateTimeFormat('en-GB', {}).format(new Date());
```
mathiasbynens, line 136: Or, if we want to apply https://xkcd.com/1179/:

```suggestion
  const date = new Date().toISOString().slice(0, 'yyyy-mm-dd'.length);
```

mathiasbynens (approved):

PR #1200, opened 2023-08-23
chore: remove some prefixes for website

Lightning00Blade:
Removes some `prefixes` in order to be more digestible and closer to the wpt.fyi dashboard that way it's easy to compare.
image

### DIFF
--- a/tools/wpt-report-builder/formatter.mjs
+++ b/tools/wpt-report-builder/formatter.mjs
@@ -55,6 +55,10 @@ export function flattenSingleTest(test) {
  *  is removed from the WPT test file, then the tests appear in the report
  *  as usual.
  */
+export function removeWebDriverBiDiPrefix(name) {
+  return name.replace('/webdriver/tests/bidi/', '');
+}
+
 export function excludeTentativeTests(test) {
   return !test.test.includes('_tentative.py');
 }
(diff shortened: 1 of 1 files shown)


thiagowfx:
The screenshot is not entirely correct:

- `logentry_added` -> `log/entry_added`
- `errorserrors.py` -> `errors/errors.py`

Removing prefixes is OK, but consider keeping at least the last directory name when the entry is terminal.

thiagowfx:
There's a second disadvantage: We lose the ability to easily do copy + paste; we would need to write the whole path by hand.

This is not a blocker from my side, I'm happy to accept the PR, but let's at least fix the issue mentioned in the previous comment.

thiagowfx (changes requested):
See comments in the PR

Lightning00Blade:
I missed that its a regression for sure -  will fix. Can you give an example for what you are suggesting.
Also thinking about it it may be easy to have a button with toggle - default is short - toggle to long, how does that sound?

thiagowfx:
>  Can you give an example for what you are suggesting.
> Also thinking about it it may be easy to have a button with toggle - default is short - toggle to long, how does that sound?

Hmm, for example:
image

I can easily copy and paste `/webdriver/tests/bidi/network/response_completed` to pass as an argument to `runWPTAll.sh` locally. With the short version, this is just `response_completed`, and one needs to write the full prefix manually.

That said, there's an alternative: We could augment the run WPT CLI tool to automatically prepend `/webdriver/tests/bidi` if it's not there. Then having the full prefix in the dashboard becomes unnecessary.

The toggle suggestion would work too.

I would say, let's try to go towards the simples route. :)

mathiasbynens:
Happy to defer to @thiagowfx here. Thanks!

Lightning00Blade:
image
image

PR #1202, opened 2023-08-24
fix: remove sandbox check

jrandolf:
It's possible for more than one sandbox to be created due to provisional frames. In this case, it's always the first one (i.e. the oldest one) that is more relevant since the user may have set that one up already through evaluation.

### DIFF
--- a/src/bidiMapper/domains/context/browsingContextImpl.ts
+++ b/src/bidiMapper/domains/context/browsingContextImpl.ts
@@ -32,6 +32,7 @@ import type {EventManager} from '../events/EventManager.js';
 import {Realm} from '../script/realm.js';
 import type {RealmStorage} from '../script/realmStorage.js';
 import type {Result} from '../../../utils/result.js';
+import {assert} from '../../../utils/assert.js';
 
 import type {BrowsingContextStorage} from './browsingContextStorage.js';
 import type {CdpTarget} from './cdpTarget.js';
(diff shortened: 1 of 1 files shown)


src/bidiMapper/domains/context/browsingContextImpl.ts
```diff
@@ -274,9 +274,10 @@ export class BrowsingContextImpl {
         sandbox,
       });
     }
-    if (maybeSandboxes.length !== 1) {
```
OrKoN, line 277: should not we still throw an error for the case `maybeSandboxes.length === 0`?

jrandolf: We create one if there isn't one in that case.

OrKoN: where does it happen? Let's change the type to return Realm|undefined then?

Lightning00Blade: I think we should still keep the check for `maybeSandboxes.length === 0` as we are relying on an Event that may or may not have been dispatched (see comment a few line above)

jrandolf: > I think we should still keep the check for `maybeSandboxes.length === 0` as we are relying on an Event that may or may not have been dispatched (see comment a few line above)

Ah, yes this is possible. I'll make it an assertion then.

PR #1208, opened 2023-08-24
chore: consistent naming of files

Lightning00Blade:


src/bidiMapper/domains/log/logHelper.spec.ts
```diff
@@ -19,7 +19,7 @@ import {expect} from 'chai';
 
 import type {Script} from '../../../protocol/protocol.js';
 
-import {logMessageFormatter} from './logHelper.js';
+import {getRemoteValuesText, logMessageFormatter} from './logHelper.js';
```
Lightning00Blade, line 22: Merged both Test files as they are not that big + its more readable to have one src <-> one spec. Happy to revert if that is not the case.

thiagowfx (approved):

src/bidiMapper/domains/log/logHelper.spec.ts
thiagowfx: LGTM

PR #1212, opened 2023-08-25
feat: implement clip for `BrowsingContext.captureScreenshot`

jrandolf:
This PR also temporarily removes support for non-top-level contexts as this is current unsupported.


src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -704,52 +707,105 @@ export class BrowsingContextImpl {
...
-              globalThis.document.documentElement.getBoundingClientRect();
-            return JSON.stringify({
-              x: docRect.x,
-              y: docRect.y,
-              width: docRect.width,
-              height: docRect.height,
-            });
-          }),
-          executionContextId: this.#defaultRealm.executionContextId,
+    const sandbox = await this.getOrCreateSandbox(undefined);
```
Lightning00Blade, line 723: Note: I was thinking about this and we should create a `BiDi Mapper` world and evaluate things there as user can remove things on the main one.
No need to action on this but please put a comment.

OrKoN (approved):
Any WPT tests to enable or e2e tests to add?

src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -704,52 +707,105 @@ export class BrowsingContextImpl {
...
 
     const result = await this.#cdpTarget.cdpClient.sendCommand(
       'Page.captureScreenshot',
-      {
-        clip: {
-          ...clip,
-          scale: 1.0,
-        },
-      }
+      {clip: {...clip, scale: 1.0}}
```
Lightning00Blade, line 808: Nit: older format looked better.

src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -853,6 +909,94 @@ export class BrowsingContextImpl {
...
+        }
+      : {
+          y: box.y,
+          height: box.height,
+        }),
+  };
+}
+
+function getIntersectionRect(
+  rect1: Readonly<Rect>,
```
Lightning00Blade, line 979: Nit: `first`, `second`

src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -853,6 +909,94 @@ export class BrowsingContextImpl {
...
+  }
+  return {
+    x: x.value,
+    y: y.value,
+    width: width.value,
+    height: height.value,
+  } as Rect;
+}
+
+function normalizeRect(box: Readonly<Rect>): Rect {
```
Lightning00Blade, line 955: Comment here would be helpful as its hard to visualize what we are doing.
Maybe extracting them into two variable can help reduce the confusion as its very nested.

src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -853,6 +909,94 @@ export class BrowsingContextImpl {
...
 }
 
+type Rect = {
+  x: number;
+  y: number;
+  width: number;
+  height: number;
+};
+
+function deserializeDOMRect(rect: Script.EvaluateResult) {
```
Lightning00Blade, line 919: Function does more then it should as it not only serializes but checks how the the command responded.
Also you can save a lot of `reac.` if use destruct on property. 
This can be a helper on the class and create sandbox and other parts.

src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -704,52 +707,105 @@ export class BrowsingContextImpl {
...
-            return JSON.stringify({
-              x: docRect.x,
-              y: docRect.y,
-              width: docRect.width,
-              height: docRect.height,
-            });
-          }),
-          executionContextId: this.#defaultRealm.executionContextId,
+    const sandbox = await this.getOrCreateSandbox(undefined);
+    let clip = await (async () => {
```
Lightning00Blade, line 724: Can we move this to a parser function?

src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -704,52 +707,105 @@ export class BrowsingContextImpl {
...
+            );
+          }
+          assert(result.result.type === 'boolean');
+          if (!result.result.value) {
+            throw new NoSuchElementException(
+              `Node ${clip.element.sharedId} is not an Element`
+            );
+          }
+          if (clip.scrollIntoView) {
+            throw new UnsupportedOperationException(
```
Lightning00Blade, line 755: Extract this before the switch statement.
Also add this to https://github.com/GoogleChromeLabs/chromium-bidi/issues/514 if you are not gonna work on it next. Same for Top-level Context

src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -853,6 +909,94 @@ export class BrowsingContextImpl {
...
+    width?.type !== 'number'
+  ) {
+    return;
+  }
+  return {
+    x: x.value,
+    y: y.value,
+    width: width.value,
+    height: height.value,
+  } as Rect;
```
Lightning00Blade, line 952: `satisfies` should be better.

src/bidiMapper/domains/context/BrowsingContextImpl.ts
jrandolf: Indeed, we should be doing this. I'll create an issue; there are other places where this occurs but don't have a comment. I think we can manage in one swoop later by looking at the issue (rather than some comment).

src/bidiMapper/domains/context/BrowsingContextImpl.ts
jrandolf: We need to `as` because those may be `SpecialNumber`s

src/bidiMapper/domains/context/BrowsingContextImpl.ts
jrandolf: Cannot because we type discriminate here.

src/bidiMapper/domains/context/BrowsingContextImpl.ts
jrandolf: I think it's still "correct" to check how the command responded. After all, it's converting `Script.EvaluateResult`. I've tried to use destructuring but it doesn't work because of type checking.

Lightning00Blade: I think the function should accept a Script.RemoveValue (or similar type) rather then a command reponse.

tests/browsing_context/capture_screenshot/test_capture_screenshot.py
```diff
@@ -68,6 +68,50 @@ async def test_screenshot(websocket, context_id, png_filename,
...
 
 
+@pytest.mark.asyncio
+async def test_screenshot_element(websocket, context_id, query_selector,
+                                  get_cdp_session_id):
+    await goto_url(websocket, context_id, 'data:text/html,<div>hello</div>')
+    session_id = await get_cdp_session_id(context_id)
+
+    # Set a fixed viewport to make the test deterministic.
+    await execute_command(
```
Lightning00Blade, line 78: There is BiDi command to set viewport that one is preferred if there are no issue with it.

tests/test_helpers.py
```diff
@@ -209,8 +209,15 @@ def assert_images_equal(img1: Image.Image | str, img2: Image.Image | str):
...
-    equal_content = not ImageChops.difference(img1.convert("RGB"),
-                                              img2.convert("RGB")).getbbox()
+    difference = ImageChops.difference(img1.convert("RGB"),
+                                       img2.convert("RGB")).getdata()
+    pixel_count = 0
+    for pixel in difference:
+        if pixel == (0, 0, 0):
+            pixel_count += 1
+
+    equal_content = pixel_count / len(difference) > 0.90
```
Lightning00Blade, line 219: Let this be variable for the function with default set to this.

tests/browsing_context/capture_screenshot/test_capture_screenshot.py
Lightning00Blade: Comment not addressed.

src/bidiMapper/domains/context/BrowsingContextImpl.ts
Lightning00Blade: Closed without a comment.

tests/browsing_context/capture_screenshot/test_capture_screenshot.py
```diff
@@ -68,6 +68,50 @@ async def test_screenshot(websocket, context_id, png_filename,
         assert_images_equal(resp["result"]["data"], png_base64)
 
 
+@pytest.mark.asyncio
+async def test_screenshot_element(websocket, context_id, query_selector,
+                                  get_cdp_session_id):
+    await goto_url(websocket, context_id, 'data:text/html,<div>hello</div>')
```
thiagowfx, line 74: use `html` fixture here

tests/browsing_context/capture_screenshot/test_capture_screenshot.py
thiagowfx: +1

This pattern was used before there was a native BiDi command, but now that there's one, it would be better to use it directly.

Bonus points if you refactor all existing usages as well (there shouldn't be many of them). This could be done in another PR.

thiagowfx (commented):
A few initial comments

src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -704,56 +707,37 @@ export class BrowsingContextImpl {
     await this.#cdpTarget.cdpClient.sendCommand('Page.bringToFront');
   }
 
-  async captureScreenshot(): Promise<BrowsingContext.CaptureScreenshotResult> {
+  async captureScreenshot(
+    params: BrowsingContext.CaptureScreenshotParameters
+  ): Promise<BrowsingContext.CaptureScreenshotResult> {
+    if (!this.isTopLevelContext()) {
+      throw new UnsupportedOperationException(
+        `Non-top-level 'context' is currently not supported`
```
thiagowfx, line 715: Did you mean:

`Non-top-level context ${contextId} is currently not supported`

Otherwise I don't get why `context` is in quotes.

Either way: Printing the ID here would be nicer, IMHO.

src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -848,11 +832,161 @@ export class BrowsingContextImpl {
     }
   }
 
+  async #parseRect(clip?: BrowsingContext.ClipRectangle) {
```
thiagowfx, line 835: add jsdoc

thiagowfx, line 835: add return type

src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -848,11 +832,161 @@ export class BrowsingContextImpl {
...
+    switch (clip.type) {
+      case 'viewport':
+        return {x: clip.x, y: clip.y, width: clip.width, height: clip.height};
+      case 'element': {
+        if (clip.scrollIntoView) {
+          throw new UnsupportedOperationException(
+            `'scrollIntoView' is currently not supported`
+          );
+        }
+        const sandbox = await this.getOrCreateSandbox(undefined);
```
thiagowfx, line 848: add a `// TODO:` here and link to https://github.com/GoogleChromeLabs/chromium-bidi/issues/1213

google3 style:

```
// TODO: #1213: Use custom sandbox specifically for Chromium BiDi
```

src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -848,11 +832,161 @@ export class BrowsingContextImpl {
...
+          }),
+          {type: 'undefined'},
+          [clip.element],
+          false,
+          Script.ResultOwnership.None,
+          {}
+        );
+        if (result.type === 'exception') {
+          throw new NoSuchElementException(
+            `Element ${clip.element.sharedId} was not found`
```
thiagowfx, line 861: add single quotes

src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -848,11 +832,161 @@ export class BrowsingContextImpl {
...
+          );
+        }
+        assert(result.result.type === 'boolean');
+        if (!result.result.value) {
+          throw new NoSuchElementException(
+            `Node ${clip.element.sharedId} is not an Element`
```
thiagowfx, line 867: add single quotes

src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -848,11 +832,161 @@ export class BrowsingContextImpl {
...
+            {type: 'undefined'},
+            [clip.element],
+            false,
+            Script.ResultOwnership.None,
+            {}
+          );
+          assert(result.type === 'success');
+          const rect = deserializeDOMRect(result.result);
+          if (!rect) {
+            throw new UnableToCaptureScreenException(
```
thiagowfx, line 890: add single quotes

src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -848,11 +832,161 @@ export class BrowsingContextImpl {
...
+      }
+    }
+  }
+
   async close(): Promise<void> {
     await this.#cdpTarget.cdpClient.sendCommand('Page.close');
   }
 }
 
+type Rect = {
```
thiagowfx, line 905: Why?

Use https://chromedevtools.github.io/devtools-protocol/tot/DOM/#type-Rect (`DOM.Rect`) instead, so that we do not reinvent the wheel.

src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -848,11 +832,161 @@ export class BrowsingContextImpl {
...
+                width: rect.width,
+              };
+            }),
+            {type: 'undefined'},
+            [clip.element],
+            false,
+            Script.ResultOwnership.None,
+            {}
+          );
+          assert(result.type === 'success');
```
thiagowfx, line 887: Would this crash the mapper if it's false?

Is it better to throw an Exception instead?

(I am not sure, just asking.)

src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -848,11 +832,161 @@ export class BrowsingContextImpl {
...
 }
 
+type Rect = {
+  x: number;
+  y: number;
+  width: number;
+  height: number;
+};
+
+function deserializeDOMRect(result: Script.RemoteValue) {
```
thiagowfx, line 912: add jsdoc and return type (`DOM.Rect`)

src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -848,11 +832,161 @@ export class BrowsingContextImpl {
...
+  }
+  return {
+    x: x.value,
+    y: y.value,
+    width: width.value,
+    height: height.value,
+  } as Rect;
+}
+
+// https://w3c.github.io/webdriver-bidi/#normalize-rect
```
thiagowfx, line 944: To satisfy jsdoc style, do:

```
/** @see https://w3c.github.io/webdriver-bidi/#normalize-rect */
```

src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -848,11 +832,161 @@ export class BrowsingContextImpl {
...
+          height: -box.height,
+        }
+      : {
+          y: box.y,
+          height: box.height,
+        }),
+  };
+}
+
+function getIntersectionRect(
```
thiagowfx, line 968: add jsdoc

src/bidiMapper/domains/context/BrowsingContextImpl.ts
thiagowfx, line 1: We should:

1) Move all these "rect" helpers to a separate file
2) Add unit tests to them

thiagowfx (changes requested):

src/bidiMapper/domains/context/BrowsingContextImpl.ts
Lightning00Blade: Given this is in a command it should be caught and turned into `Unknown Exeption`, if you looks at assert's implementation it checks the value and throws an error.

tests/test_helpers.py
```diff
@@ -209,8 +209,15 @@ def assert_images_equal(img1: Image.Image | str, img2: Image.Image | str):
...
-    equal_content = not ImageChops.difference(img1.convert("RGB"),
-                                              img2.convert("RGB")).getbbox()
+    difference = ImageChops.difference(img1.convert("RGB"),
+                                       img2.convert("RGB")).getdata()
+    pixel_count = 0
+    for pixel in difference:
+        if pixel == (0,0,0):
+            pixel_count += 1
+            
+    equal_content = pixel_count / len(difference) > 0.90
```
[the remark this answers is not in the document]
jrandolf: Indeed, that's what it means.

src/bidiMapper/domains/context/BrowsingContextImpl.ts
jrandolf: Because it's referencing the parameter.

src/bidiMapper/domains/context/BrowsingContextImpl.ts
jrandolf: We don't use return types for private functions. The reason we do it for public functions is for contracting with external modules.

tests/browsing_context/capture_screenshot/test_capture_screenshot.py
jrandolf: I think it makes sense to do this in another PR if we have multiple usage.

src/bidiMapper/domains/context/BrowsingContextImpl.ts
jrandolf: Given that the implementation WPT and the implementation tests test this, I'm not sure if this is beneficial/tests anything that isn't already tested.

thiagowfx: Personally I am exhaustively adding unit tests where applicable for the PRs I am sending. It also acts as a contract/documentation for the helpers. http://go/pyramid-of-testing comes to mind here.

That said, we can also discuss this during the stand-up to have consensus. But more testing never hurts (except for the opportunity cost of time spent elsewhere).

tests/test_helpers.py
```diff
@@ -193,7 +193,7 @@ def AnyExtending(expected: list | dict):
     return expected
 
 
-def assert_images_equal(img1: Image.Image | str, img2: Image.Image | str):
+def assert_images_similar(img1: Image.Image | str, img2: Image.Image | str, percent = 0.95):
     """Assert that the given images are equal."""
```
thiagowfx, line 199: this docstring is now outdated

src/bidiMapper/domains/context/BrowsingContextImpl.ts
thiagowfx: @Lightning00Blade did you mean this as an AI or just a statement?

src/bidiMapper/domains/context/BrowsingContextImpl.ts
thiagowfx: Also note: your comment pertains to 2), but 1) is still applicable here and would be cleaner, IMHO.

thiagowfx (changes requested):

src/bidiMapper/domains/context/BrowsingContextImpl.ts
jrandolf: > Move all these "rect" helpers to a separate file

This isn't useful since they are only used in one file. Usually exporting to a module requires the functions to be used _elsewhere_. Since these functions only have callsites within the class, I don't see a purpose in doing this except making refactoring harder and the build system try harder.

thiagowfx: I disagree, but for the sake of being pragmatic, won't block it.

thiagowfx (approved):

src/bidiMapper/domains/context/BrowsingContextImpl.ts
Lightning00Blade: From what I remember when we try to process a command we do it though a `try/catch` if anything fails and the error is not of Type `Exeption` (Internal class of all BiDi specific errors) we return a Unknown error.

PR #1215, opened 2023-08-25
feat: add network redirects

Lightning00Blade:
This should cover the base case of redirects, some edge cases may need to be fix similar to what to do with the `responseExtraInfo` that's emitted between events. Puppeteer sort handles them as Reponses.

### DIFF
--- a/src/bidiMapper/domains/context/BrowsingContextImpl.ts
+++ b/src/bidiMapper/domains/context/BrowsingContextImpl.ts
@@ -65,6 +65,8 @@ export class BrowsingContextImpl {
         DOMContentLoaded: new Deferred<Protocol.Page.LifecycleEventEvent>(),
         load: new Deferred<Protocol.Page.LifecycleEventEvent>(),
       },
+      frameStartedLoading:
+        new Deferred<Protocol.Page.FrameStartedLoadingEvent>(),
     },
   };
 
@@ -588,7 +590,7 @@ export class BrowsingContextImpl {
     wait: BrowsingContext.ReadinessState
   ): Promise<BrowsingContext.NavigateResult> {
     try {
-      void new URL(url);
+      new URL(url);
     } catch {
       throw new InvalidArgumentException(`Invalid URL: ${url}`);
     }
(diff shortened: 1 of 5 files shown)


src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -628,7 +628,8 @@ export class BrowsingContextImpl {
 
     return {
       navigation: cdpNavigateResult.loaderId ?? null,
-      url,
+      // Url can change due to redirect get the latest one.
+      url: wait === BrowsingContext.ReadinessState.None ? url : this.#url,
```
Lightning00Blade, line 632: We may need to await for something like `Frame.startedLoading` event and always resolve the url from `this` .
Will take a look in the spec of how that should behave.

src/bidiMapper/domains/network/NetworkStorage.ts
```diff
@@ -135,6 +130,14 @@ export class NetworkStorage {
     };
   }
 
+  get(id: Network.Request) {
+    return this.#requestMap.get(id);
+  }
+
+  delete(id: Network.Request) {
+    this.#requestMap.delete(id);
+  }
```
thiagowfx, line 139: Not a fan of these method names, because there are 3 maps in this class. It's not obvious which one it refers to.

src/bidiMapper/domains/network/NetworkStorage.spec.ts
```diff
@@ -34,12 +34,6 @@ describe('NetworkStorage', () => {
     networkStorage = new NetworkStorage(eventManager);
   });
 
-  it('disposeRequestMap', () => {
```
thiagowfx, line 37: replace the test?

Lightning00Blade: Best we can do is pass in a mock Request and see if its dispose gets called. Other then that it's a private methods so we can't poke.

src/bidiMapper/domains/context/BrowsingContextImpl.ts
Lightning00Blade: Waiting for the `Frame.startedLoading` would have minimal effect as the event does not contain enough information.

src/bidiMapper/domains/network/NetworkStorage.ts
Lightning00Blade: Added a Specific Request

PR #1217, opened 2023-08-28
fix: use dpr for scrolling

jrandolf:


src/bidiMapper/domains/input/ActionDispatcher.ts
```diff
@@ -550,8 +554,8 @@ export class ActionDispatcher {
           'Input.dispatchMouseEvent',
           {
             type: 'mouseWheel',
-            deltaX,
-            deltaY,
+            deltaX: deltaX * devicePixelRatio,
```
OrKoN, line 557: we need to device by devicePixelRatio here more likely to account for the `* devicePixelRatio` on the CDP side

OrKoN (approved):

PR #1235, opened 2023-09-01
chore: refactor to use assert and some Default map issues

Lightning00Blade:


src/bidiMapper/domains/network/NetworkStorage.ts
```diff
@@ -135,12 +133,10 @@ export class NetworkStorage {
   }
 
   deleteRequest(id: Network.Request) {
-    const request = this.#requestMap.get(id);
-    if (request) {
-      request.dispose();
+    if (this.#requestMap.has(id)) {
```
Lightning00Blade, line 136: Default maps are tricky as they will create the resources every time.
We should only dispose them when we actually have them.

thiagowfx (approved):

Issue #1183, opened 2023-08-21
Network Interception: Add and Remove Intercepts


thiagowfx:
This is a subtask of https://github.com/GoogleChromeLabs/chromium-bidi/issues/644. The first submitted PR was https://github.com/GoogleChromeLabs/chromium-bidi/pull/1050.

Template:

```
Bug: https://github.com/GoogleChromeLabs/chromium-bidi/issues/644
DevTools: https://source.chromium.org/chromium/chromium/src/+/main:content/browser/devtools/protocol/fetch_handler.cc
Doc: [go/webdriver-bidi:network-request-interception-proposal](http://go/webdriver-bidi:network-request-interception-proposal)
Spec: https://pr-preview.s3.amazonaws.com/w3c/webdriver-bidi/pull/429.html#network-intercepts
Puppeteer: https://github.com/puppeteer/puppeteer/blob/main/test/src/requestinterception.spec.ts
Upstream: https://github.com/w3c/webdriver-bidi/pull/429
```

Simplified template:

```
Bug: https://github.com/GoogleChromeLabs/chromium-bidi/issues/644
Doc: [go/webdriver-bidi:network-request-interception-proposal](http://go/webdriver-bidi:network-request-interception-proposal)
Spec: https://w3c.github.io/webdriver-bidi/#network-intercepts
```

Priority: High

References:

- https://wicg.github.io/urlpattern/#match

thiagowfx:
Closing this in favour of #644.

PR #1250, opened 2023-09-04
feat: add get network intercepts method

thiagowfx:
Uses dependency: https://www.npmjs.com/package/urlpattern-polyfill

Issue: #1183
Spec: https://w3c.github.io/webdriver-bidi/#get-the-network-intercepts
WPT: https://github.com/web-platform-tests/wpt/pull/41662/

### DIFF
--- a/package.json
+++ b/package.json
@@ -220,7 +220,7 @@
     "terser": "5.19.2",
     "tslib": "2.6.1",
     "typescript": "5.1.6",
-    "urlpattern-polyfill": "^9.0.0",
+    "urlpattern-polyfill": "9.0.0",
     "websocket": "1.0.34",
     "wireit": "0.10.0",
     "ws": "8.13.0",
--- a/package.json
+++ b/package.json
@@ -220,6 +220,7 @@
     "terser": "5.19.2",
     "tslib": "2.6.1",
     "typescript": "5.1.6",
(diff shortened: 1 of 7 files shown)


thiagowfx:
Missing:

- `// TODO: Step 4. URL serialization.`
- `static matchUrlPattern`

package.json
```diff
@@ -220,6 +220,7 @@
     "terser": "5.19.2",
     "tslib": "2.6.1",
     "typescript": "5.1.6",
+    "urlpattern-polyfill": "^9.0.0",
```
Lightning00Blade, line 223: Pin. Also It may be needed in Dependencies, rather then here.

jrandolf, line 223: Lock the version

src/bidiMapper/domains/network/NetworkProcessor.ts
```diff
@@ -14,6 +14,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import {URLPattern} from 'urlpattern-polyfill';
```
Lightning00Blade, line 17: I think we should try to load it only if its not there, like we did with `uuid`.

thiagowfx: Done

PR #1251, opened 2023-09-04
feat: add url getter to NetworkRequest

thiagowfx:
Issue: #1183

### DIFF
--- a/src/bidiMapper/domains/network/NetworkRequest.ts
+++ b/src/bidiMapper/domains/network/NetworkRequest.ts
@@ -62,6 +62,18 @@ export class NetworkRequest {
     this.#eventManager = eventManager;
   }
 
+  /** Returns the URL associated with this request. */
+  get url(): string | undefined {
+    if (this.#requestWillBeSentEvent) {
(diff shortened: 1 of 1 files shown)


src/bidiMapper/domains/network/NetworkRequest.ts
```diff
@@ -237,7 +249,7 @@ export class NetworkRequest {
     return {
       request:
         this.#requestWillBeSentEvent?.requestId ?? NetworkRequest.#unknown,
-      url: this.#requestWillBeSentEvent?.request.url ?? NetworkRequest.#unknown,
+      url: this.url ?? NetworkRequest.#unknown,
```
Lightning00Blade, line 252: Not sure if this is correct when we have redirects, but I will verify it with my change.

Lightning00Blade (approved):

src/bidiMapper/domains/network/NetworkRequest.ts
thiagowfx: Noted

PR #1255, opened 2023-09-05
feat: `browsingContext.reload`: return the navigation instead of empty

thiagowfx:
Blocked until WPT tests are updated: https://github.com/GoogleChromeLabs/chromium-bidi/pull/1255#issuecomment-1707285470

Issue: https://github.com/w3c/webdriver-bidi/issues/527
Issue: #650

### DIFF
--- a/src/bidiMapper/domains/context/BrowsingContextImpl.ts
+++ b/src/bidiMapper/domains/context/BrowsingContextImpl.ts
@@ -26,7 +26,10 @@ import {
   UnableToCaptureScreenException,
   UnknownErrorException,
   UnsupportedOperationException,
-  type EmptyResult,
+  InvalidArgumentException,
+  NoSuchElementException,
+  UnableToCaptureScreenException,
+  Script,
 } from '../../../protocol/protocol.js';
 import {assert} from '../../../utils/assert.js';
 import {Deferred} from '../../../utils/deferred.js';
@@ -35,6 +38,8 @@ import {inchesFromCm} from '../../../utils/unitConversions.js';
 import type {EventManager} from '../events/EventManager.js';
 import {Realm} from '../script/Realm.js';
 import type {RealmStorage} from '../script/RealmStorage.js';
+import type {Result} from '../../../utils/result.js';
+import {assert} from '../../../utils/assert.js';
 
 import type {BrowsingContextStorage} from './BrowsingContextStorage.js';
 import type {CdpTarget} from './CdpTarget.js';
@@ -650,7 +655,7 @@ export class BrowsingContextImpl {
   async reload(
     ignoreCache: boolean,
     wait: BrowsingContext.ReadinessState
-  ): Promise<EmptyResult> {
+  ): Promise<BrowsingContext.NavigateResult> {
     await this.targetUnblockedOrThrow();
 
     await this.#cdpTarget.cdpClient.sendCommand('Page.reload', {
@@ -670,7 +675,13 @@ export class BrowsingContextImpl {
         break;
     }
 
-    return {};
+    return {
+      navigation:
+        wait === BrowsingContext.ReadinessState.None
+          ? null
+          : this.navigableId ?? null,
+      url: this.url,
+    };
   }
 
   async setViewport(viewport: BrowsingContext.Viewport | null) {
--- a/src/bidiMapper/domains/context/BrowsingContextImpl.ts
+++ b/src/bidiMapper/domains/context/BrowsingContextImpl.ts
@@ -26,10 +26,6 @@ import {
   UnableToCaptureScreenException,
   UnknownErrorException,
   UnsupportedOperationException,
-  InvalidArgumentException,
-  NoSuchElementException,
-  UnableToCaptureScreenException,
-  Script,
 } from '../../../protocol/protocol.js';
 import {assert} from '../../../utils/assert.js';
 import {Deferred} from '../../../utils/deferred.js';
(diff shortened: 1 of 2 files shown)


src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -664,7 +664,10 @@ export class BrowsingContextImpl {
         break;
     }
 
-    return {};
+    return {
+      navigation: this.loaderId ?? null,
```
thiagowfx, line 668: As per https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-navigate:

> Loader identifier. This is omitted in case of same-document navigation, as the previously committed loaderId would not change.

My understanding is that `reload` does not alter this ID. Same below for the URL.

src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -143,9 +142,10 @@ export class BrowsingContextImpl {
   }
 
   /**
+   * The navigable ID.
    * @see https://html.spec.whatwg.org/multipage/document-sequences.html#navigable
    */
-  get navigableId(): string | undefined {
+  get loaderId(): string | undefined {
```
Lightning00Blade, line 148: Revert this. BrowsingContext should be thought as the BiDi BrowsingContext this is CDP specific naming. 
They both mean roughly the same thing.

src/bidiMapper/domains/context/BrowsingContextImpl.ts
OrKoN, line 668: I think it should get a new ID according to the spec. Is there a WPT change?

Note that reload is not a same-document navigation.

thiagowfx: Where would the ID come from? CDP `Page.reload` doesn't return any ID.

OrKoN: it might come via lifecycle events or we change Page.reload implementation

thiagowfx: If it comes via lifecycle events, then the getters should be correct. No?

OrKoN: idk, how about a test for that?

thiagowfx: Augmented E2E tests. This seems to confirm my hypothesis. The tests are all passing.

`None` is the only case wherein we do not have the correct (new) ID.

OrKoN: what if a reload is triggering a redirect?

OrKoN (approved):

src/bidiMapper/domains/context/BrowsingContextImpl.ts
Lightning00Blade: I still believe this is mistake.

src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -664,7 +663,10 @@ export class BrowsingContextImpl {
         break;
     }
 
-    return {};
+    return {
+      navigation: this.navigableId ?? null,
```
Lightning00Blade, line 667: I think this in correct. We should set this to null if `BrowsingContext.ReadinessState.None` as now we will return the older one.

Lightning00Blade (approved):

src/bidiMapper/domains/context/BrowsingContextImpl.ts
thiagowfx: Can we defer this question to the upstream issue and to WPT tests?

tests/browsing_context/test_reload.py
```diff
@@ -38,12 +38,15 @@ async def test_browsingContext_reload_waitNone(websocket, context_id, html):
         })
 
     # Assert command done.
-    response = await read_JSON_message(websocket)
-    assert response["result"] == {}
+    response1 = await read_JSON_message(websocket)
```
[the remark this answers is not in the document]
thiagowfx: Could you give a more concrete suggestion of what you mean?

src/bidiMapper/domains/context/BrowsingContextImpl.ts
thiagowfx: I will change it to `null` for now which is better than returning the old ID, but I am not sure whether that's the final desired state.

thiagowfx:
Updated `test_reload.py` based on `wait=None` returning `navigation=null`.

tests/browsing_context/test_reload.py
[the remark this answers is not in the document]
Lightning00Blade: Some of them seem to be events so `dom_content_loaded_event` for example would be the name.

Lightning00Blade (approved):

thiagowfx:
WPT is expectedly failing. The tests will need to be updated before we submit this.

@whimboo are you going to do so?

whimboo:
> @whimboo are you going to do so?

Yes, but we have to wait for the ongoing discussion to be finished. Once we know the expected return value [I will fix our WebDriver BiDi implementation in Firefox and update the tests](https://bugzilla.mozilla.org/show_bug.cgi?id=1851583).

src/bidiMapper/domains/context/BrowsingContextImpl.ts
thiagowfx: https://github.com/w3c/webdriver-bidi/issues/532

PR #1256, opened 2023-09-05
build: add prettier and yapf to package.json

thiagowfx:


thiagowfx:
I need this back because my local pre-commit setup is broken.

Lightning00Blade (approved):

PR #1257, opened 2023-09-05
chore: fix logging issues

Lightning00Blade:
Some variable renaming cleanup

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -6,7 +6,7 @@ name: E2E tests
 permissions: read-all
 
 env:
-  DEBUG: 'bidi:server:info,bidi:mapper:debug:*'
+  DEBUG: 'bidi:server:*,bidi:mapper:debug:*'
   FORCE_COLOR: 3
   PIP_DISABLE_PIP_VERSION_CHECK: 1
 
--- a/src/bidiServer/MapperServer.ts
+++ b/src/bidiServer/MapperServer.ts
@@ -17,21 +17,30 @@
 
 import type Protocol from 'devtools-protocol';
 import WebSocket from 'ws';
-import debug from 'debug';
+import debug, {type Debugger} from 'debug';
 
 import {CdpConnection} from '../cdp/cdpConnection.js';
 import type {CdpClient} from '../cdp/cdpClient.js';
-import {LogType} from '../utils/log.js';
+import type {LogType, LogPrefix} from '../utils/log.js';
 import {WebSocketTransport} from '../utils/websocketTransport.js';
 
 const debugInternal = debug('bidi:mapper:internal');
 const debugInfo = debug('bidi:mapper:info');
-const debugMapperDebugOthers = debug('bidi:mapper:debug:others');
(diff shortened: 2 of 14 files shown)


.github/workflows/e2e.yml
```diff
@@ -6,7 +6,7 @@ name: E2E tests
 permissions: read-all
 
 env:
-  DEBUG: 'bidi:server:info,bidi:mapper:debug:*'
+  DEBUG: 'bidi:server:*,bidi:mapper:debug:*'
```
Lightning00Blade, line 9: Decided to go with this rather then update `DEBUG` when `VERBOSE` so they don't depend on each. Else it becomes problematic with mix and matching.

src/bidiTab/mapperTabPage.ts
```diff
@@ -69,8 +69,11 @@ export function log(logType: LogPrefix, ...messages: unknown[]) {
     return;
   }
 
-  // If `sendDebugMessage` is defined, send the log message there.
-  global.window?.sendDebugMessage?.(JSON.stringify({logType, messages}));
+  // Skip sending BiDi logs as they are logged once by `bidi:server:*`
+  if (!logType.startsWith(LogType.bidi)) {
```
Lightning00Blade, line 73: Move to here as it makes more sense to not send the info back rather then sending and filtering on the other end.

src/utils/log.ts
```diff
@@ -21,6 +21,7 @@ export enum LogType {
   cdp = 'cdp',
   debug = 'debug',
   debugError = 'debug:error',
+  debugInfo = 'debug:info',
```
Lightning00Blade, line 24: If you set `bidi:debug:*` things with `bidi:debug` prefix will not be shown.

tools/run-bidi-server.mjs
```diff
@@ -60,7 +60,8 @@ if (CHANNEL === 'local') {
 }
 
 // DEBUG = (empty string) is allowed.
-const DEBUG = process.env.DEBUG ?? '*';
+let DEBUG = process.env.DEBUG ?? 'bidi:*';
```
Lightning00Blade, line 63: Don't show every package that uses `debug`

src/bidiServer/MapperServer.ts
```diff
@@ -17,21 +17,30 @@
...
 import {WebSocketTransport} from '../utils/websocketTransport.js';
 
 const debugInternal = debug('bidi:mapper:internal');
 const debugInfo = debug('bidi:mapper:info');
-const debugMapperDebugOthers = debug('bidi:mapper:debug:others');
-const debugMapperDebugPrefix = 'bidi:mapper:debug:';
+const debugOthers = debug('bidi:mapper:debug:others');
+// Memorizes a debug creation
+const loggers = new Map<string, Debugger>();
+const getLogger = (type: LogPrefix) => {
```
Lightning00Blade, line 32: This should help with speed imo, as internally it does some checks and setups for each call which I am not sure if calling `debug` memorizes itself.

thiagowfx (approved):

PR #1267, opened 2023-09-07
fix: reject errors instead of throwing in CDP target init

jrandolf:


src/bidiMapper/domains/context/CdpTarget.ts
```diff
@@ -162,7 +162,8 @@ export class CdpTarget {
     } catch (error: any) {
       // The target might have been closed before the initialization finished.
       if (!this.#cdpClient.isCloseError(error)) {
-        throw error;
+        this.#targetUnblocked.reject(error);
```
Lightning00Blade, line 165: Shouldn't it be better to resolve with Result<Error>?

OrKoN (approved):

src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -600,7 +600,10 @@ export class BrowsingContextImpl {
       throw new InvalidArgumentException(`Invalid URL: ${url}`);
     }
 
-    await this.targetUnblocked();
+    const result = await this.targetUnblocked();
```
thiagowfx, line 603: If we do this everywhere, is it more appropriate to either (i) add a boolean argument to `targetUnblocked` to optionally `throw` on error, OR (ii) always throw unconditionally?

PR #1270, opened 2023-09-08
fix: unhandled promises in CDP targets

jrandolf:
This PR fixes an issue where Node.js throws from unhandled promises in the array due to the `await` from sending 'Runtime.runIfWaitingForDebugger'.


OrKoN:
cc @thiagowfx @Lightning00Blade

OrKoN (approved):

PR #1273, opened 2023-09-08
fix: catch uncaught throws from #getHandleFromWindow

jrandolf:


src/bidiMapper/domains/script/ChannelProxy.ts
```diff
@@ -56,8 +56,12 @@ export class ChannelProxy {
 
   /** Gets a ChannelProxy from window and returns its handle. */
   async startListenerFromWindow(realm: Realm, eventManager: EventManager) {
-    const channelHandle = await this.#getHandleFromWindow(realm);
-    void this.#startListener(realm, channelHandle, eventManager);
+    try {
+      const channelHandle = await this.#getHandleFromWindow(realm);
+      void this.#startListener(realm, channelHandle, eventManager);
```
Lightning00Blade, line 61: If this fails it will throw unhandled as its never awaited.

Lightning00Blade (changes requested):

src/bidiMapper/domains/script/ChannelProxy.ts
jrandolf: #startListener never fails.

Lightning00Blade: Approving but it should not have a `async` as it never resolves as well.

Lightning00Blade (approved):

PR #1280, opened 2023-09-11
feat: throw InvalidArgument instead of UnsupportedOperation for print…

thiagowfx:
…ing an empty content area

Related, in sync with, and blocked on: https://github.com/web-platform-tests/wpt/pull/40872


thiagowfx:
This PR should not be submitted until the WPT test above is merged.

thiagowfx:
I'm mistaken, it turns out this isn't blocked on upstream. It's actually the other way around: this needs to be landed before the WPT test.

sadym-chromium (approved):

PR #1285, opened 2023-09-12
fix: replace empty string `namespaceURI` with null

sadym-chromium:
Temp fix before the [CDP fix](crrev.com/c/4856861) is landed


Lightning00Blade (approved):
WPT needs updated expectations.

PR #1292, opened 2023-09-12
fix: flatten event type

jrandolf:


package-lock.json
```diff
@@ -53,6 +52,7 @@
         "terser": "5.19.2",
         "tslib": "2.6.1",
         "typescript": "5.1.6",
+        "urlpattern-polyfill": "9.0.0",
```
thiagowfx, line 55: Is this change deliberate?

thiagowfx (approved):

package-lock.json
jrandolf: The `package-lock.json` wasn't updated before.

PR #1296, opened 2023-09-13
chore: improve CDP event types

jrandolf:


src/protocol/cdp.ts
```diff
@@ -20,6 +20,8 @@ import type Protocol from 'devtools-protocol';
 
 import type {BrowsingContext, JsUint} from './webdriver-bidi.js';
 
+export type EventNames = 'cdp' | Event['method'];
```
Lightning00Blade, line 23: It seem we never fix this type this should not include `cdp` as `cdp` is threaded as a module and should be added to the module enum.

src/protocol/cdp.ts
```diff
@@ -70,18 +72,14 @@ export type GetSessionResult = {
...
-> = {
-  event: EventName;
-  params: ProtocolMapping.Events[EventName][0];
-  session: Protocol.Target.SessionID;
+export type EventDataFor<EventName extends keyof ProtocolMapping.Events> = {
+  method: `cdp.${EventName}`;
+  params: EventParametersFor<EventName>;
 };
 
-// XXX: Support `cdp` (all events)
```
thiagowfx, line 86: To be clear: Is `cdp` (=all events) supported with this PR?

thiagowfx: It seems the answer is no, and that you just moved the TODO to the other file. Just checking.

thiagowfx (approved):

Lightning00Blade (approved):

PR #1301, opened 2023-09-14
fix: restore functionality to subscribe to all CDP events

Lightning00Blade:
Fixes an issue where we could not subscribe to `cdp` module directly but were required to subscribe to each event individually.


src/bidiMapper/domains/events/EventManager.ts
```diff
@@ -275,24 +285,13 @@ export class EventManager {
...
-]);
-
-function checkEventName(name: string) {
-  if (
-    !EVENT_NAMES.has(name as never) &&
-    !name.startsWith('cdp.') &&
-    name !== 'cdp'
-  ) {
-    throw new InvalidArgumentException(`Unknown event: ${name}`);
+  static isSupportedEvent(name: string) {
```
thiagowfx, line 289: The function name is a bit misleading: it sounds like it would return a boolean.

How about: `assertSupportedEvent` or `ensureSupportedEvent`? (I prefer the former)

src/bidiMapper/domains/events/EventManager.ts
```diff
@@ -275,24 +285,13 @@ export class EventManager {
...
+    if (
+      !EVENT_NAMES.has(name as never) &&
+      !name.startsWith(ChromiumBidi.BiDiModule.Cdp)
```
thiagowfx, line 292: Should we re-add the `.` suffix to `startsWith`?

Although highly unlikely, this would match "cdpee". If we re-add the dot, then `cdp.` ensures uniqueness.

thiagowfx: Alternatively: name + split on "." + get first element + check that it equals Cdp

The first approach is simpler though

src/bidiMapper/domains/events/EventManager.ts
```diff
@@ -29,6 +29,16 @@ import {OutgoingMessage} from '../../OutgoingMessage.js';
 
 import {SubscriptionManager} from './SubscriptionManager.js';
 
+const EVENT_NAMES = new Set([
```
thiagowfx, line 32: What if we moved `EVENT_NAMES` to https://github.com/GoogleChromeLabs/chromium-bidi/blob/793c0af2def70a2fce67d0f79ae60b197bcc1fb8/src/protocol/chromium-bidi.ts#L21, alongside the `EventNames` type?

I think it would be easier to maintain / keep in sync (especially when/as new modules are added).

tests/cdp/test_cdp.py
thiagowfx, line 1: The main reason this file was within the `session` directory was to mirror the WPT structure: https://googlechromelabs.github.io/chromium-bidi/ as there's a `session` directory there as well. Then it is easier to find E2E tests associated with the same domain of WPT tests.

Although "cdp" is reasonably valid here, but it breaks this structure. Could we discuss this in a separate PR? (Or maybe not do it at all?)

src/bidiMapper/domains/events/SubscriptionManager.ts
```diff
@@ -130,7 +130,26 @@ export class SubscriptionManager {
...
+        if (eventMethod.startsWith(ChromiumBidi.BiDiModule.Cdp)) {
+          const cdpPriority = contextToEventMap
+            .get(context)
+            ?.get(ChromiumBidi.BiDiModule.Cdp);
+          // If we subscribe to the event directly and `cdp` module as well
+          // priority will be different we take the one with higher priority
+          return priority && cdpPriority
+            ? Math.min(priority, cdpPriority)
+            : // At this point we know that we have subscribed
+              //to only one of the two
```
thiagowfx, line 148:
```suggestion
              // to only one of the two
```

src/bidiMapper/domains/events/SubscriptionManager.ts
```diff
@@ -130,7 +130,26 @@ export class SubscriptionManager {
...
+        const priority = contextToEventMap.get(context)?.get(eventMethod);
+        // For CDP we can't provide specific event name when subscribing
+        // to the module directly.
+        // Because of that we need to see event `cdp` exits in the map.
+        if (eventMethod.startsWith(ChromiumBidi.BiDiModule.Cdp)) {
+          const cdpPriority = contextToEventMap
+            .get(context)
+            ?.get(ChromiumBidi.BiDiModule.Cdp);
+          // If we subscribe to the event directly and `cdp` module as well
+          // priority will be different we take the one with higher priority
```
thiagowfx, line 144: The comment says "take the one with higher priority", but we're using `Math.min`. Is this intentional?

src/bidiMapper/domains/events/SubscriptionManager.ts
```diff
@@ -130,7 +130,26 @@ export class SubscriptionManager {
...
     // Get all the subscription priorities.
     const priorities: number[] = relevantContexts
-      .map((c) => contextToEventMap.get(c)?.get(eventMethod))
+      .map((context) => {
+        // Get the priority for exact event name
+        const priority = contextToEventMap.get(context)?.get(eventMethod);
+        // For CDP we can't provide specific event name when subscribing
+        // to the module directly.
+        // Because of that we need to see event `cdp` exits in the map.
+        if (eventMethod.startsWith(ChromiumBidi.BiDiModule.Cdp)) {
```
thiagowfx, line 139: We use `name.startsWith(ChromiumBidi.BiDiModule.Cdp)` in at least two different places (plus see the comment in the other place about the dot).

Should we extract this into an `isCdpEventName` helper?

src/bidiMapper/domains/events/EventManager.ts
thiagowfx, line 1: With this PR, the XXX on https://github.com/GoogleChromeLabs/chromium-bidi/blob/793c0af2def70a2fce67d0f79ae60b197bcc1fb8/src/protocol/chromium-bidi.ts#L35C23-L35C23 can be removed! :D

thiagowfx (changes requested):

src/bidiMapper/domains/events/EventManager.ts
Lightning00Blade: We need to match `cdp` as well as `cdp` what is why I did it as so, but you are correct we should match for both individually.

tests/cdp/test_cdp.py
Lightning00Blade: Ok will revert. Motivation was `cdp` module should have a separate folder as the other modules.

src/bidiMapper/domains/events/SubscriptionManager.ts
Lightning00Blade: It should be the same as the other comment `minimum` priority.

src/bidiMapper/domains/events/EventManager.ts
Lightning00Blade: Thanks for pointing that out!

src/bidiMapper/domains/events/SubscriptionManager.ts
Lightning00Blade: I think with your suggestion for using a better way to determine if its under the `cdp` module scope yes, but for a single like I would be opposed to it.

src/bidiMapper/domains/events/EventManager.ts
Lightning00Blade: I just moved it to the top in this PR, will think about if there is a way to make it seems less.

thiagowfx (approved):

Issue #575, opened 2023-03-27
Add local server


sadym-chromium:
To test some scenarios (OOPIF, redirects) and in the same time to keep the tests local, we need to have a possibility to run local server supporting different DNS aliases and SSL connections.

Requirements:
- [x] Custom response, including custom HTTP response codes
~~- [ ] (optional) DNS aliases~~ (can be done via Chrome CLI args)
- [x] SSL connections

WPT can be used as a reference.

## Tracking

- [x] Add local server support
- [x] Add a test for closing browsingContext after navigating to a faulty certificate page
- [x] Test for 200 page
- [x] Test for 301 page
- [x] Test for 401 page
- [x] Test for hanging response

sadym-chromium:
Example where is can be needed: #513

thiagowfx:
Some ideas:

- `http.server` from python: https://pythonbasics.org/webserver/ (via @OrKoN)
- `http-server` from npm: https://www.npmjs.com/package/http-server (via @mathiasbynens)

OrKoN:
DNS aliases is a feature of Chrome and we don't need it in the server.

thiagowfx:
I just stumbled upon https://pytest-httpserver.readthedocs.io/en/latest... @Maksim Sadym , I wonder if we could reuse most of that library for our purposes, instead of https://github.com/GoogleChromeLabs/chromium-bidi/pull/1359
> pytest-httpserver is a python package which allows you to start a real HTTP server for your tests. The server can be configured programmatically to how to respond to requests.> This project aims to provide an easy to use API to start the server, configure the request handlers and then shut it down gracefully. All of these without touching a configuration file or dealing with daemons.

It also supports TLS.

thiagowfx:
>  SSL connections

This one can be marked as done (c.f. `bad_ssl_url`).

mathiasbynens:
> > SSL connections
> 
> This one can be marked as done (c.f. `bad_ssl_url`).

Thanks!

sadym-chromium:
Done

PR #1304, opened 2023-09-15
feat: handle `Fetch.requestPaused` event

thiagowfx:
Bug: #644

### DIFF
--- a/src/bidiMapper/domains/network/NetworkRequest.ts
+++ b/src/bidiMapper/domains/network/NetworkRequest.ts
@@ -45,7 +45,11 @@ export class NetworkRequest {
    */
   readonly requestId: Network.Request;
 
+  /** Indicates whether the request is blocked by a network intercept. */
+  #isBlocked = false;
+
   #servedFromCache = false;
+
   #redirectCount: number;
 
   #eventManager: EventManager;
@@ -184,6 +188,11 @@ export class NetworkRequest {
     );
   }
 
+  /** Fired whenever a network request interception is hit. */
+  onRequestPaused(_event: Protocol.Fetch.RequestPausedEvent) {
+    this.#isBlocked = true;
+  }
+
   dispose() {
     const result = {
       kind: 'error' as const,
@@ -199,11 +208,9 @@ export class NetworkRequest {
 
   #getBaseEventParams(): Network.BaseParameters {
     return {
-      // TODO: implement.
-      isBlocked: false,
+      isBlocked: this.#isBlocked,
       context: this.#context,
       navigation: this.#getNavigationId(),
-      // TODO: implement.
       redirectCount: this.#redirectCount,
       request: this.#getRequestData(),
       // Timestamp should be in milliseconds, while CDP provides it in seconds.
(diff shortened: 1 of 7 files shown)


src/bidiMapper/domains/network/NetworkRequest.ts
```diff
@@ -196,7 +207,7 @@ export class NetworkRequest {
   #getBaseEventParams(): Network.BaseParameters {
     return {
       // TODO: implement.
```
OrKoN, line 209: remove todo?

tests/network/test_add_intercept.py
```diff
@@ -255,3 +257,87 @@ async def test_add_intercept_type_pattern_port_empty_invalid(websocket):
...
+        #     },
+        # ],
+    ],
+    ids=[
+        "string",
+        #  "pattern",
+        #  "string and pattern",
+    ])
+async def test_add_intercept_blocks(websocket, context_id, url_patterns):
+    url = "https://www.example.com/"
```
OrKoN, line 299: do we want to test with example.com or perhaps set up a local server?

tests/network/test_add_intercept.py
```diff
@@ -255,3 +257,87 @@ async def test_add_intercept_type_pattern_port_empty_invalid(websocket):
...
+    await send_JSON_command(
+        websocket, {
+            "method": "browsingContext.navigate",
+            "params": {
+                "url": url,
+                "context": context_id,
+            }
+        })
+
+    event_response = await wait_for_event(websocket, "cdp.Fetch.requestPaused")
```
OrKoN, line 325: do we want to test using a bidi event? or alternatively add a todo.

tests/network/test_add_intercept.py
thiagowfx: We still do not have local server support: https://github.com/GoogleChromeLabs/chromium-bidi/issues/575

tests/network/test_add_intercept.py
thiagowfx: There's no BiDi event equivalent to `Fetch.requestPaused`.

And the other events (response started, response completed, etc) require other intercept commands/methods which aren't yet available.

I am considering the TODO to write more tests, but IMHO this test is OK as is - having at least one test that tests the CDP event should be fine. That said, we can add a TODO to stop using example.com

thiagowfx:
New question: https://github.com/w3c/webdriver-bidi/issues/550

thiagowfx:
There's still a lot to do (intended to follow up in other PRs), but I'd say this PR is ready for submission as is.

tests/network/test_add_intercept.py
OrKoN: should we address it next? we will need to have many more tests were the example.com might not be sufficient or reliable.

thiagowfx: I agree, the more the time passes the more important this feature becomes. Let's try to prioritize it. There are some pre-existing tests that use example.com that would also benefit from local servers.

(Out of scope of this PR though.)

PR #1305, opened 2023-09-15
chore: extract IIFE into method on LogManager

Lightning00Blade:


thiagowfx (approved):
LGTM but it would be nice if you could add one sentence to the commit message to explain the rationale

src/bidiMapper/domains/log/LogManager.ts
```diff
@@ -179,4 +168,20 @@ export class LogManager {
       }
     );
   }
+
+  /**
+   * Try the best to get the exception text.
+   */
+  static async #getExceptionText(
```
jrandolf, line 175: This doesn't use `this`

Lightning00Blade: It's `static`.

PR #1308, opened 2023-09-15
chore: extract event utils, add tests and docs

thiagowfx:


thiagowfx:
Sadly, it's not possible to make `isCdpEvent` be more precise and check for actual event names, unless we maintain the list of events manually.

src/bidiMapper/domains/events/SubscriptionManager.ts
```diff
@@ -22,6 +22,8 @@ import {
 } from '../../../protocol/protocol.js';
 import type {BrowsingContextStorage} from '../context/BrowsingContextStorage.js';
 
+import * as events from './events.js';
```
Lightning00Blade, line 25: Lets import them individually.

PR #1313, opened 2023-09-18
feat: support redirect response

Lightning00Blade:


thiagowfx (commented):
Should this make some test pass, and/or should we add one?

(It's OK to do it in another PR, but we should have at least one test for redirection if feasible. Feasible = we may need a local server, which we don't have support for yet...)

thiagowfx (approved):

Lightning00Blade:
Tests are failing as `Headful` auto updates `http` to `https` I will look into WPT if there is a test that can confirm this behavior and if not add it.

OrKoN (approved):

thiagowfx (approved):

PR #1314, opened 2023-09-18
ci: update devtools-protocol automatically

Lightning00Blade:


.pre-commit-config.yaml
```diff
@@ -27,6 +27,7 @@ repos:
       - id: check-executables-have-shebangs
       - id: check-yaml
       - id: end-of-file-fixer
+        exclude: '.browser'
```
Lightning00Blade, line 30: This way we don't need to trim this file when getting content.

.github/workflows/update-browser-version.yml
```diff
@@ -24,15 +24,10 @@ jobs:
...
+          node-version: lts/*
           cache: npm
-      - uses: google/wireit@f3a3c79c553122e2fe5829eeac7d815326502903 # setup-github-actions-caching/v1
-      - name: Install and build npm dependencies
-        run: npm ci
-      - name: Download the latest version
-        run: |
-          browser_spec="$(npx @puppeteer/browsers install chrome@canary --path /tmp/browsers | cut -f 1 -d' ')"
-          echo "$browser_spec" > .browser
+      - name: Update browser pin and devtools-protocol
```
thiagowfx, line 29:
```suggestion
      - name: Update .browser pin and devtools-protocol
```

.github/workflows/update-browser-version.yml
```diff
@@ -24,15 +24,10 @@ jobs:
       - name: Set up Node.js
         uses: actions/setup-node@5e21ff4d9bc1a8cf6de233a3057d20ec6b3fb69d # v3.8.1
         with:
-          node-version: 18
+          node-version: lts/*
```
thiagowfx, line 27: Could we leave the version unchanged in this PR?

Ideally we should use the same version in _every_ workflow. It's currently inconsistent: https://github.com/search?q=repo%3AGoogleChromeLabs%2Fchromium-bidi%20node-version&type=code

It's better to address that in a separate PR

tools/update_chrome_revision.mjs
```diff
@@ -0,0 +1,69 @@
...
+}
+
+const {version, revision} = await getVersionAndRevisionForCanary();
+await updateDevToolsProtocolVersion(revision);
+
+const browserVersion = `chrome@${version}`;
+await writeFile('./.browser', browserVersion, 'utf-8');
+
+// Create new `package-lock.json` as we update devtools-protocol
+execSync('npm install --ignore-scripts');
```
thiagowfx, line 67: If we're still doing `npm install`, then `wireit` should be kept in the github workflow

.github/workflows/update-browser-version.yml
thiagowfx: https://github.com/GoogleChromeLabs/chromium-bidi/pull/1315

thiagowfx (changes requested):

OrKoN (approved):

tools/update_chrome_revision.mjs
Lightning00Blade: Nope that is why `--ignore-scripts` is there/

thiagowfx: Ah OK

.github/workflows/update-browser-version.yml
Lightning00Blade: I disagree as browser pin is `.browser` file. Comment is only valid if you say we should write it as `Update .browser file`.

thiagowfx (approved):
The only other potential comment would be to add an unit test to the node file. But given that (i) we currently do not have precedent for this (in tools/) and (ii) it would be mostly mocked anyway (because of the network requests) I'm OK with skipping this. Just make sure to test the script and the workflow.

.github/workflows/update-browser-version.yml
thiagowfx: Not a strong opinion here as it was just a nit, feel free to leave as is then

PR #1318, opened 2023-09-19
feat: implement network interception fail request

thiagowfx:
Bug: #644
Spec: https://w3c.github.io/webdriver-bidi/#command-network-failRequest

### DIFF
--- a/src/bidiMapper/domains/network/NetworkManager.ts
+++ b/src/bidiMapper/domains/network/NetworkManager.ts
@@ -25,7 +25,7 @@ import type Protocol from 'devtools-protocol';
 import type {Network} from '../../../protocol/protocol.js';
 import type {CdpTarget} from '../context/CdpTarget.js';
 
-import type {NetworkRequest} from './NetworkRequest.js';
+import {NetworkRequest} from './NetworkRequest.js';
 import type {NetworkStorage} from './NetworkStorage.js';
 
 /** Maps 1:1 to CdpTarget. */
@@ -47,12 +47,25 @@ export class NetworkManager {
    * Gets the network request with the given ID, if any.
    * Otherwise, creates a new network request with the given ID and cdp target.
    */
-  #getOrCreateNetworkRequest(id: Network.Request): NetworkRequest {
-    const request = this.#networkStorage.getRequest(id);
+  #getOrCreateNetworkRequest(
+    id: Network.Request,
+    redirectCount?: number
+  ): NetworkRequest {
+    let request = this.#networkStorage.getRequest(id);
     if (request) {
       return request;
     }
-    return this.#networkStorage.createRequest(id, {cdpTarget: this.#cdpTarget});
+
+    request = new NetworkRequest(
+      id,
+      this.#networkStorage.eventManager,
+      this.#cdpTarget,
+      redirectCount
+    );
+
+    this.#networkStorage.addRequest(request);
+
+    return request;
   }
 
   static create(
@@ -81,17 +94,17 @@ export class NetworkManager {
         if (request && request.isRedirecting()) {
           request.handleRedirect(params);
           networkManager.#networkStorage.deleteRequest(params.requestId);
-          networkManager.#networkStorage
-            .createRequest(params.requestId, {
-              cdpTarget,
-              redirectCount: request.redirectCount + 1,
-            })
+          networkManager
+            .#getOrCreateNetworkRequest(
+              params.requestId,
+              request.redirectCount + 1
+            )
             .onRequestWillBeSentEvent(params);
         } else if (request) {
           request.onRequestWillBeSentEvent(params);
         } else {
-          networkManager.#networkStorage
-            .createRequest(params.requestId, {cdpTarget})
+          networkManager
+            .#getOrCreateNetworkRequest(params.requestId)
             .onRequestWillBeSentEvent(params);
         }
       }
--- a/src/bidiMapper/domains/network/NetworkManager.ts
+++ b/src/bidiMapper/domains/network/NetworkManager.ts
@@ -47,15 +47,12 @@ export class NetworkManager {
    * Gets the network request with the given ID, if any.
    * Otherwise, creates a new network request with the given ID and cdp target.
    */
-  #getOrCreateNetworkRequest(
-    id: Network.Request,
-    cdpTarget: CdpTarget
-  ): NetworkRequest {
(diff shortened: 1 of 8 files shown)


src/bidiMapper/domains/network/NetworkManager.ts
```diff
@@ -43,12 +43,19 @@ export class NetworkManager {
...
   }
 
-  #getOrCreateNetworkRequest(id: Network.Request): NetworkRequest {
+  /**
+   * Gets the network request with the given ID, if any.
+   * Otherwise, creates a new network request with the given ID and cdp target.
+   */
+  #getOrCreateNetworkRequest(
+    id: Network.Request,
+    cdpTarget: CdpTarget
```
OrKoN, line 52: why do we need to pass the target around everywhere? is it just an id or the entire target?

thiagowfx: Alternatively we could pass the session ID, and retrieve the target from it.

Here's the issue: **We need to have the target available in the NetworkProcessor, so that it can issue commands to CDP. **

I was discussing it with @sadym-chromium yesterday, and the only effective way we found to do so is to get the target from the NetworkManager. And even then, we store the target in the NetworkRequest, not in the NetworkManager. Whenever there are some intercept commands, it then uses the request to retrieve the target.

If you can think of some other (better) way to inject the CdpTarget in the NetworkProcessor, I'd be happy to make changes to reflect that. It's just that, the way things are currently set up:

- NetworkManager is created from CdpTarget
- NetworkProcessor is created from CommandProcessor (which is created from BidiServer)

There's currently no straightforward way to inject the target into the processor. (Unless we use the NetworkStorage for that, but it would violate its contract.)

OrKoN, line 52: should this always be `this.#cdpTarget`?

OrKoN: I am not sure I have a good solution but the current versions looks somewhat confusing to me.

src/bidiMapper/domains/network/NetworkManager.ts
```diff
@@ -78,13 +85,16 @@ export class NetworkManager {
           request.handleRedirect(params);
           networkManager.#networkStorage.deleteRequest(params.requestId);
           networkManager.#networkStorage
-            .createRequest(params.requestId, request.redirectCount + 1)
+            .createRequest(params.requestId, {
```
OrKoN, line 88: why do we sometimes create requests directly through storage and sometimes though the network manager?

src/bidiMapper/domains/network/NetworkManager.ts
thiagowfx: Good point. I think that should work as well and would simplify a few things. (Although, note the comment above still remains open in case you have a suggestion to inject CdpTarget into NetworkProcessor directly.)

src/bidiMapper/domains/network/NetworkProcessor.ts
```diff
@@ -75,8 +75,30 @@ export class NetworkProcessor {
...
+    const blockedRequest = this.#networkStorage.getBlockedRequest(networkId);
+    const {request: fetchId, phase} = blockedRequest;
+
+    if (phase === Network.InterceptPhase.AuthRequired) {
+      throw new InvalidArgumentException(
+        `Blocked request for network id '${networkId}' is in 'AuthRequired' phase`
+      );
+    }
+
+    const cdpTarget = this.#networkStorage.getRequest(networkId)?.cdpTarget;
```
OrKoN, line 91: I suggest moving this into a method on NetwrokRequest

src/bidiMapper/domains/network/NetworkManager.ts
thiagowfx: Done

src/bidiMapper/domains/network/NetworkStorage.ts
```diff
@@ -49,7 +51,8 @@ export class NetworkStorage {
   readonly #blockedRequestMap = new Map<
     Network.Request,
```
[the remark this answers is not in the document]
thiagowfx: Do you mean that there's a risk of collision in network request IDs?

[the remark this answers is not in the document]
OrKoN: if it is the same as Protocol.Fetch.RequestId, then I believe yes.

src/bidiMapper/domains/network/NetworkProcessor.ts
thiagowfx: Done

src/bidiMapper/domains/network/NetworkManager.ts
thiagowfx: Fixed

src/bidiMapper/domains/network/NetworkStorage.ts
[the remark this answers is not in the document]
OrKoN: looks like it would be a UUID here? if that is the case, it should be alright.

src/bidiMapper/domains/network/NetworkStorage.ts
```diff
@@ -255,6 +259,23 @@ export class NetworkStorage {
...
+   * Returns the blocked request associated with the given network ID, if any.
+   */
+  getBlockedRequest(networkId: Network.Request): {
+    request: Protocol.Fetch.RequestId;
+    phase?: Network.InterceptPhase; // TODO: make non-optional.
+    response?: Network.ResponseData; // TODO: make non-optional.
+  } {
+    const blockedRequest = this.#blockedRequestMap.get(networkId);
+    if (!blockedRequest) {
+      throw new NoSuchRequestException(
```
OrKoN, line 272: this looks like it belongs to the NetworkProcessor. I suggest moving it there.

src/bidiMapper/domains/network/NetworkManager.ts
OrKoN: the current version looks much better to me.

OrKoN (approved):

src/bidiMapper/domains/network/NetworkStorage.ts
thiagowfx: Initially that was the case, but I moved it to storage.

Just moved it back.

PR #1321, opened 2023-09-20
fix: separate click count by button

jrandolf:
Previously, the click count was measured globally among all buttons. This was incorrect as browsers separate click count by button.


Lightning00Blade (approved):
LGTM. 
Do you think this will help with falkyness we have observed?

PR #1324, opened 2023-09-20
ci: fix issues found by pytype

thiagowfx:
Docs: https://github.com/google/pytype

Not enabling pytype in pre-commit as it does not yet support python 3.11: https://github.com/google/pytype/issues/1308


thiagowfx:
```
Python versions > 3.10 are not yet supported.
```

Of course, I didn't notice this as I am using python 3.10 locally.

Then I will revert the pytype addition, and just leave the bug fixes it has found.

Lightning00Blade (approved):

PR #1331, opened 2023-09-21
feat(network interception): implement continue request

thiagowfx:
Bug: #644
Spec: https://w3c.github.io/webdriver-bidi/#command-network-continueRequest

### DIFF
--- a/src/bidiMapper/domains/network/NetworkManager.ts
+++ b/src/bidiMapper/domains/network/NetworkManager.ts
@@ -193,7 +193,7 @@ export class NetworkManager {
           networkManager.#networkStorage.addBlockedRequest(params.networkId, {
             request: params.requestId, // intercept request id
             phase,
-            // TODO: Finish populate response / ResponseData.
+            // TODO: Finish populating response / ResponseData.
             response: {
               url: params.request.url,
               // TODO: populate.
--- a/src/bidiMapper/domains/network/NetworkManager.ts
+++ b/src/bidiMapper/domains/network/NetworkManager.ts
@@ -175,6 +175,10 @@ export class NetworkManager {
             params.responseStatusCode === undefined
           ) {
             phase = Network.InterceptPhase.BeforeRequestSent;
+          } else if (params.responseStatusCode === 401) {
+            // TODO: Write a test for this phase.
+            // https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401
+            phase = Network.InterceptPhase.AuthRequired;
           } else {
             phase = Network.InterceptPhase.ResponseStarted;
           }
--- a/src/bidiMapper/domains/network/NetworkManager.ts
+++ b/src/bidiMapper/domains/network/NetworkManager.ts
@@ -22,15 +22,11 @@
  */
 import type Protocol from 'devtools-protocol';
 
-import {Network} from '../../../protocol/protocol.js';
+import type {Network} from '../../../protocol/protocol.js';
 import type {CdpTarget} from '../context/CdpTarget.js';
 
 import {NetworkRequest} from './NetworkRequest.js';
 import type {NetworkStorage} from './NetworkStorage.js';
-import {
-  computeResponseHeadersSize,
-  bidiNetworkHeadersFromCdpFetchHeaderEntryArray,
-} from './NetworkUtils.js';
 
 /** Maps 1:1 to CdpTarget. */
 export class NetworkManager {
@@ -164,62 +160,9 @@ export class NetworkManager {
       'Fetch.requestPaused',
       (params: Protocol.Fetch.RequestPausedEvent) => {
         if (params.networkId) {
-          // The stage of the request can be determined by presence of
-          // responseErrorReason and responseStatusCode -- the request is at
-          // the response stage if either of these fields is present and in the
-          // request stage otherwise.
-          let phase: Network.InterceptPhase;
-          if (
-            params.responseErrorReason === undefined &&
-            params.responseStatusCode === undefined
-          ) {
-            phase = Network.InterceptPhase.BeforeRequestSent;
-          } else if (
-            params.responseStatusCode === 401 &&
-            params.responseStatusText === 'Unauthorized'
-          ) {
-            // https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401
-            phase = Network.InterceptPhase.AuthRequired;
-          } else {
-            phase = Network.InterceptPhase.ResponseStarted;
-          }
-
-          const headers = bidiNetworkHeadersFromCdpFetchHeaderEntryArray(
-            // TODO: Use params.request.headers if request?
-            params.responseHeaders
-          );
-
-          networkManager.#networkStorage.addBlockedRequest(params.networkId, {
-            request: params.requestId, // intercept request id
-            phase,
-            // TODO: Finish populating response / ResponseData.
-            response: {
-              url: params.request.url,
-              // TODO: populate.
-              protocol: '',
-              status: params.responseStatusCode ?? 0,
-              statusText: params.responseStatusText ?? '',
-              // TODO: populate.
-              fromCache: false,
-              headers,
-              // TODO: populate.
-              mimeType: '',
-              // TODO: populate.
-              bytesReceived: 0,
-              headersSize: computeResponseHeadersSize(headers),
-              // TODO: consider removing from spec.
-              bodySize: 0,
-              // TODO: consider removing from spec.
-              content: {
-                size: 0,
-              },
-              // TODO: populate.
-              authChallenge: undefined,
-            },
-          });
           networkManager
             .#getOrCreateNetworkRequest(params.networkId)
-            .onRequestPaused(params);
+            .onRequestPaused(params, networkManager.#networkStorage);
         }
       }
     );
--- a/src/bidiMapper/domains/network/NetworkManager.ts
+++ b/src/bidiMapper/domains/network/NetworkManager.ts
@@ -168,7 +168,6 @@ export class NetworkManager {
           // responseErrorReason and responseStatusCode -- the request is at
           // the response stage if either of these fields is present and in the
           // request stage otherwise.
-          // TODO: Populate the "authRequired" phase.
           let phase: Network.InterceptPhase;
           if (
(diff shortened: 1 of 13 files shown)


thiagowfx:
@sadym-chromium Just to let you know:

>  The stage of the request can be determined by presence of responseErrorReason and responseStatusCode -- the request is at the response stage if either of these fields is present and in the request stage otherwise.

from [`Fetch.requestPaused`](https://chromedevtools.github.io/devtools-protocol/tot/Fetch/#event-requestPaused) has indeed worked as intended.

The only thing missing is to find a way to distinguish the "authRequired" phase.

thiagowfx:
Marking as 'ready for review', but note the "TODOs". They are currently WIP, some of them may be still addressed in this PR, others will be deferred.

thiagowfx:
Friendly ping.

sadym-chromium:
> The only thing missing is to find a way to distinguish the "authRequired" phase.

Isn't it a response with code [`401`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401)?

src/bidiMapper/domains/network/NetworkProcessor.ts
```diff
@@ -66,8 +66,38 @@ export class NetworkProcessor {
     };
   }
 
-  continueRequest(_params: Network.ContinueRequestParameters): EmptyResult {
-    throw new UnknownCommandException('Not implemented yet.');
+  async continueRequest(
+    params: Network.ContinueRequestParameters
+  ): Promise<EmptyResult> {
+    const networkId = params.request;
+    const blockedRequest = this.#getBlockedRequest(networkId);
```
sadym-chromium, line 73:
```suggestion
    const blockedRequest = this.#getBlockedRequest(networkId);
    if(blockedRequest === undefined)
        throw new InvalidArgumentException(...)
```

src/bidiMapper/domains/network/NetworkProcessor.ts
```diff
@@ -66,8 +66,38 @@ export class NetworkProcessor {
...
+    // ; cookies
+    // ; body
+
+    await this.#networkStorage
+      .getRequest(networkId)
+      ?.continueRequest(fetchId, url, method);
+
+    this.#networkStorage.removeBlockedRequest(networkId);
+
+    // TODO: Remove from network request map?
```
sadym-chromium, line 98: I don't think so, as the request still exists.

src/bidiMapper/domains/network/NetworkRequest.ts
```diff
@@ -210,6 +211,29 @@ export class NetworkRequest {
       requestId: fetchRequestId,
       errorReason,
     });
+
+    // TODO: mark is as non-blocked?
+    // this.#isBlocked = false;
```
sadym-chromium, line 216: Is the request still blocked.

src/bidiMapper/domains/network/NetworkRequest.ts
```diff
@@ -210,6 +211,29 @@ export class NetworkRequest {
...
+  }
+
+  /** @see https://chromedevtools.github.io/devtools-protocol/tot/Fetch/#method-continueRequest */
+  async continueRequest(
+    fetchRequestId: Protocol.Fetch.RequestId,
+    url?: string,
+    method?: string
+  ) {
+    // TODO: Expand.
+    await this.#cdpTarget.cdpClient.sendCommand('Fetch.continueRequest', {
```
sadym-chromium, line 226: What is the request is not blocked? Should it be handled in this method?

tests/network/test_continue_request.py
```diff
@@ -0,0 +1,241 @@
...
+@pytest.mark.asyncio
+@pytest.mark.parametrize("phase", [
+    "responseStarted",
+    pytest.param("authRequired",
+                 marks=pytest.mark.xfail(
+                     reason='TODO: #644: implement "authRequired" phase'))
+])
+async def test_continue_request_invalid_phase(websocket, context_id, phase):
+    # TODO: make offline.
+    url = "https://www.example.com/"
```
sadym-chromium, line 48: NIT: if we move `https://www.example.com/` into a fixture, switching to local server will be way easier.

tests/network/test_continue_request.py
```diff
@@ -0,0 +1,241 @@
...
+    await send_JSON_command(
+        websocket, {
+            "method": "browsingContext.navigate",
+            "params": {
+                "url": url,
+                "context": context_id,
+            }
+        })
+    event_response = await wait_for_event(websocket, "cdp.Fetch.requestPaused")
+    network_id = event_response["params"]["params"]["networkId"]
```
sadym-chromium, line 121: This looks odd. Do we really need waiting for the CDP request first?

sadym-chromium, line 121: This looks odd. Do we really need waiting for the CDP request first?

sadym-chromium (commented):
In general LGTM, but the e2e tests are odd. Can we make them pure BiDi?

src/bidiMapper/domains/network/NetworkProcessor.ts
thiagowfx: Done (will push later)

src/bidiMapper/domains/network/NetworkRequest.ts
thiagowfx: It shouldn't be, as at that point we have "continued" it. I just uncommented that line to reflect so.

tests/network/test_continue_request.py
thiagowfx: This is a good point, but since this PR is getting quite big already, let's address it in a separate one.

thiagowfx:
> Isn't it a response with code [401](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401)?

That didn't work. Tested with http://httpstat.us/401

tests/network/test_continue_request.py
thiagowfx: The responseStarted event is never emitted in this case, that's why.

I am not yet sure whether we should manually emit it (then we could use it in the test).

Either way: For now, this is out of scope of this PR (let's keep using CDP events for consistency, then we can revisit all tests later and change them to BiDi events if warranted).

But let me know if you think we should emit the events manually.

src/bidiMapper/domains/network/NetworkRequest.ts
thiagowfx: I don't understand the comment. Could you elaborate?

sadym-chromium: What if user tries to continue the request which is not blocked?

sadym-chromium (approved):

tests/network/test_continue_request.py
sadym-chromium: Agree to move it out of scope, but eventually we should get rid of this CDP hack.

sadym-chromium: Actually, I don't quite follow the e2e test. If there is no BiDi event pointing the request is blocked, how user is expected to continue it?

sadym-chromium: Right, there is `responseStarted ` event, as irequest is intercepted on the request state. Butthere should be an event `network.beforeRequestSent` event with `isBlocked: true` should be emitted. Why cannot we use it?

src/bidiMapper/domains/network/NetworkManager.ts
```diff
@@ -155,14 +159,64 @@ export class NetworkManager {
       }
     );
 
+    // https://chromedevtools.github.io/devtools-protocol/tot/Fetch/#event-requestPaused
     cdpTarget.cdpClient.on(
       'Fetch.requestPaused',
       (params: Protocol.Fetch.RequestPausedEvent) => {
         if (params.networkId) {
+          // The stage of the request can be determined by presence of
```
OrKoN, line 167: nit: it's more readable if the event handler is extracted into a method and there is a clear place where the event handlers are removed.

src/bidiMapper/domains/network/NetworkRequest.ts
thiagowfx: Added `test_continue_request_non_blocked_request`

tests/network/test_fail_request.py
```diff
@@ -108,6 +155,90 @@ async def test_fail_request_twice(websocket, context_id):
...
+    (["beforeRequestSent", "authRequired"], False),
+])
+@pytest.mark.skip(reason="TODO: Use our own test server.")
+async def test_fail_request_with_auth_required_phase(
+        websocket, context_id, phases, exception_and_response_expected):
+    # TODO: make offline.
+    # All of these URLs work.
+    # url = "https://authenticationtest.com/HTTPAuth/"
+    # url = "http://the-internet.herokuapp.com/basic_auth"
+    url = "http://httpstat.us/401"
```
thiagowfx, line 173: TODO: Remove this before submitting

tests/network/test_continue_request.py
```diff
@@ -0,0 +1,293 @@
...
+    } == exception_info.value.args[0]
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    "phase, url",
+    [
+        # TODO: make offline.
+        # ("responseStarted", "https://www.example.com/"),
+        ("authRequired", "http://httpstat.us/401"
```
thiagowfx, line 45: TODO: remove this before submitting

tests/network/test_continue_request.py
thiagowfx: > But there should be an event network.beforeRequestSent event with isBlocked: true should be emitted. Why cannot we use it?

So basically you're answering 'yes' to the question I made above. OK. Let's emit those events in a separate PR then. We will need to revisit all methods to do so.

thiagowfx: Resolving but made a note of it.

sadym-chromium: https://github.com/GoogleChromeLabs/chromium-bidi/pull/1400

src/bidiMapper/domains/network/NetworkManager.ts
thiagowfx: Done

PR #1338, opened 2023-09-22
ci: cache support is only top level

Lightning00Blade:
For node set-up cache to work you need to have a `package-lock.json`.
As we will run this infrequently we can opt to not use cache.


OrKoN (approved):

.github/workflows/puppeteer.yml
```diff
@@ -24,7 +24,6 @@ jobs:
       - name: Set up Node.js
         uses: actions/setup-node@5e21ff4d9bc1a8cf6de233a3057d20ec6b3fb69d # v3.8.1
         with:
-          cache: npm
```
thiagowfx, line 27: It's very easy for this to be reintroduced in a future refactoring that tries to make all actions consistent (with cache). A tombstone comment here would have been helpful.

PR #1342, opened 2023-09-25
feat: implement angle inputs

jrandolf:


src/bidiMapper/domains/input/ActionDispatcher.ts
```diff
@@ -822,3 +820,70 @@ function getCdpButton(button: number) {
...
+    let tiltXRadians = 0;
+    let tiltYRadians = 0;
+
+    if (altitudeAngle === 0) {
+      // the pen is in the X-Y plane
+      if (azimuthAngle === 0 || azimuthAngle === 2 * Math.PI) {
+        // pen is on positive X axis
+        tiltXRadians = Math.PI / 2;
+      }
+      if (azimuthAngle === Math.PI / 2) {
```
OrKoN, line 848: does this comparison on floats work? I think we need to compare with some precision

OrKoN (changes requested):
is it tested by WPT? let's enable them here or add e2e tests.

src/bidiMapper/domains/input/ActionDispatcher.ts
jrandolf:
```
6.283185307179586 === Math.PI * 2 // true
```

OrKoN: what about `6.2831853071795867` or `6.2831853071795864`?

OrKoN: Interesting though, I guess this works as expected 👍

OrKoN (approved):

Issue #1344, opened 2023-09-25
Using `omitBackground` in `captureScreenshot`


jrandolf:
Opened this issue to discuss whether we should by default omit backgrounds on `captureScreenshot`. Firefox already has this behavior of screenshots and IMO if the format is `png`, we should be omitting the background as that's probably what's expected from the user.

OrKoN:
does the spec say anything about that?

jrandolf:
The spec doesn't say anything about this.

Lightning00Blade:
Closing this we should either use `goog:` values or open an issue on spec.

PR #1345, opened 2023-09-25
chore: remove default background in screenshots

jrandolf:
See https://github.com/GoogleChromeLabs/chromium-bidi/issues/1344


src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -757,13 +757,26 @@ export class BrowsingContextImpl {
...
+        }
+      );
+      return await this.#cdpTarget.cdpClient.sendCommand(
+        'Page.captureScreenshot',
+        {clip: {...rect, scale: 1.0}}
+      );
+    } finally {
+      await this.#cdpTarget.cdpClient
+        .sendCommand('Emulation.setDefaultBackgroundColorOverride', {
+          color: undefined,
```
thiagowfx, line 774: What if the background was previously set to some custom color? Would this command reset it to the previous value?

src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -757,13 +757,26 @@ export class BrowsingContextImpl {
...
+        })
+        .catch(() => {
+          void 0;
```
thiagowfx, line 777: Log some error within the catch block. It helps with debugging should this situation ever happen.

src/bidiMapper/domains/context/BrowsingContextImpl.ts
thiagowfx, line 1: Could you also add a test?

Considering the helper functions we already have in pytest, this should be relatively straightforward.

What I have in mind is a square ⬛ with fixed known dimensions in a page with a custom background (e.g. Red). Then you can easily compare the final image base64 value, it should match the square only.

thiagowfx (changes requested):
Wait for upstream approval before submission

src/bidiMapper/domains/context/BrowsingContextImpl.ts
OrKoN: Yes

thiagowfx: Let me rephrase to double-check:

- Assume background color is r=0 g=255 b=0 (green)
- If I am a BiDi+ user who called `Emulation.setDefaultBackgroundColorOverride(r: 255, g: 0, b: 0, a: 0)` (override to red),
- Then called `capture screenshot`

Will the background color be reset back to red afterwards?

It seems to me it would be reset, but to green.

This is unexpected, as it's not in the contract of the capture screenshot method. Right?

OrKoN: So the I would expect the color to be back to green assuming we have the following sequence of calls and CDP is working as expected:

1) setDefaultBackgroundColorOverride(green, sessionA) # overrides = (green)
2) setDefaultBackgroundColorOverride(red, sessionB) # overrides = (green, red)
3) setDefaultBackgroundColorOverride(white, sessionA) # overrides = (green, white)
4) setDefaultBackgroundColorOverride(undefined, sessionA) # overrides = (green)

That is how overrides in CDP should work, if it turns out that this override works differently we could see if we want to get that fixed in CDP. Also, this method is usually only called for screenshot taking with the white bg color.

OrKoN: Looks like we would need to fix CDP to make this CDP API work properly across sessions https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/inspector/inspector_emulation_agent.cc;l=582;drc=58b8aff108cb11d3755f4cc18efa10f971be9682;bpv=1;bpt=1

OrKoN:
@jrandolf was it mean to be closed?

PR #1349, opened 2023-09-26
ci: set default debug depth

Lightning00Blade:


OrKoN (approved):

sadym-chromium (approved):
Let's either add a line into readme, or set the limit to something like `999`.

sadym-chromium (approved):
Let's either add a line into readme, or set the limit to something like `999`.

PR #1354, opened 2023-09-28
test: improve preload script tests

sadym-chromium:
1. Split `test_preloadScript_add_runImmediately` into 2 separate tests: 
  * `test_preloadScript_add_sandbox_new_context`
  * `test_preloadScript_add_sandbox_existing_context`
2. Switched some tests to accumulative value to verify the scripts order is correct and no unexpected side-effects are present.


tests/script/test_add_preload_script.py
```diff
@@ -500,42 +482,47 @@ async def test_preloadScript_add_loadedInNewContexts(websocket, context_id,
...
+async def test_preloadScript_add_sandbox_new_context(websocket, html):
+    await execute_command(
         websocket, {
             "method": "script.addPreloadScript",
             "params": {
                 "functionDeclaration": "() => { window.foo='bar'; }",
                 "sandbox": "MY_SANDBOX",
             }
         })
-    assert result == {'script': ANY_UUID}
```
thiagowfx, line 512: `assert result == {'script': ANY_UUID}`

==> keeping this should be harmless

tests/script/test_add_preload_script.py
```diff
@@ -546,58 +533,65 @@ async def test_preloadScript_add_sandbox(websocket, context_id, html):
...
+    # Add preload script, which checks if the page script was already executed.
     await execute_command(
         websocket, {
             "method": "script.addPreloadScript",
             "params": {
-                "functionDeclaration": "() => { window.foo='bar'; }",
+                "functionDeclaration": "() => { "
+                                       "window.SOME_VAR=["
+                                       "...(window.SOME_VAR ?? []), "
+                                       "'MY_SANDBOX_PRELOAD_SCRIPT']; }",
```
thiagowfx, line 546: It could be more readable to use `"""` here.

(If you agree, please change all occurrences)

tests/script/test_add_preload_script.py
```diff
@@ -95,77 +112,42 @@ async def test_preloadScript_add_multipleScripts(websocket, context_id, html):
...
-        })
-    assert result == {'script': ANY_UUID}
-    id1 = result['script']
-
-    result = await execute_command(
-        websocket, {
-            "method": "script.addPreloadScript",
-            "params": {
-                "functionDeclaration": EXPRESSION,
+                "functionDeclaration": "() => { "
```
thiagowfx, line 125: Please re-add this assertion:

```
assert id1 != id2
```

The preload scripts should have different IDs.

tests/script/test_add_preload_script.py
```diff
@@ -95,77 +112,42 @@ async def test_preloadScript_add_multipleScripts(websocket, context_id, html):
...
-                },
-                "awaitPromise": True,
-                "resultOwnership": "root"
-            }
-        })
-    assert result["result"] == {"type": "number", "value": 3}
-
-
-@pytest.mark.asyncio
-async def test_preloadScript_add_sameScript_multipleTimes(
```
thiagowfx, line 127: Any reason to remove this? `test_preloadScript_add_sameScript_multipleTimes`

The idea was to ensure the implementation is not "taking any shortcut" or caching the scripts in any way, and that they are effectively added twice, and with different IDs.

I think it would be a good idea not to remove this test.
You could refactor it to the way it is here, then we'd have an array with two elements.

thiagowfx (changes requested):

OrKoN:
Probably `test:` instead of `feat:` would be better.

tests/script/test_add_preload_script.py
sadym-chromium: done

tests/script/test_add_preload_script.py
sadym-chromium: Ok, added it back

tests/script/test_add_preload_script.py
sadym-chromium: It is already tested in a few places, and I removed it to reduce potential noise and improve test readability. Anyways, I rolled it back, if you think it's better to have it.

tests/script/test_add_preload_script.py
sadym-chromium: No objections. Done.

thiagowfx (approved):

PR #1386, opened 2023-10-05
Implement image formats in `BrowsingContext.CaptureScreenshot`

jrandolf:


src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -930,6 +932,31 @@ export class BrowsingContextImpl {
   }
 }
 
+function getImageFormatParameters(
+  params: Readonly<BrowsingContext.CaptureScreenshotParameters>
```
thiagowfx, line 936: Because this includes `context` which is not relevant to this function, I would make the type narrower: just pass two arguments:

`getImageFormatParameters(type: string, quality?: number)`

thiagowfx (approved):

src/bidiMapper/domains/context/BrowsingContextImpl.ts
jrandolf: This would require more checks on the parameters because type and quality are gated within the same object. It would just be more code.

src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -930,6 +932,31 @@ export class BrowsingContextImpl {
...
+        ...(quality === undefined ? {} : {quality: Math.round(quality * 100)}),
+      } as const;
+    }
+    case 'image/webp': {
+      return {format: 'webp'} as const;
+    }
+  }
+  throw new UnsupportedOperationException(
+    `Image format '${type}' is not a supported format`
+  );
```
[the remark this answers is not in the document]
jrandolf: Since this can be linted, I would rather have this done at once then bikeshed it here.

PR #1397, opened 2023-10-06
test: pytest `wait_for_event` helper: match prefixes

thiagowfx:
This added syntactic sugar is more flexible when we want to wait for any event from a given domain, e.g. "cdp" or "network".


sadym-chromium (approved):
nice!

PR #1403, opened 2023-10-08
test: use `match` argument in `pytest.raises` to simplify assertions

thiagowfx:
Docs: https://stackoverflow.com/a/56569533/1745064

### DIFF
--- a/tests/network/test_remove_intercept.py
+++ b/tests/network/test_remove_intercept.py
@@ -21,7 +21,12 @@ from test_helpers import (ANY_TIMESTAMP, ANY_UUID, AnyExtending,
 
 @pytest.mark.asyncio
 async def test_remove_intercept_no_such_intercept(websocket):
-    with pytest.raises(Exception) as exception_info:
+    with pytest.raises(
+            Exception,
+            match=str({
+                "error": "no such intercept",
+                "message": "Intercept '00000000-0000-0000-0000-000000000000' does not exist."
+            })):
         await execute_command(
             websocket, {
                 "method": "network.removeIntercept",
@@ -29,11 +34,6 @@ async def test_remove_intercept_no_such_intercept(websocket):
                     "intercept": "00000000-0000-0000-0000-000000000000",
                 },
             })
-
-    assert {
-        "error": "no such intercept",
-        "message": "Intercept '00000000-0000-0000-0000-000000000000' does not exist."
-    } == exception_info.value.args[0]
 
 
 @pytest.mark.asyncio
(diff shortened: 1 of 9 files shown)


tests/network/test_remove_intercept.py
```diff
@@ -21,7 +21,12 @@
 
 @pytest.mark.asyncio
 async def test_remove_intercept_no_such_intercept(websocket):
-    with pytest.raises(Exception) as exception_info:
+    with pytest.raises(
```
sadym-chromium, line 24: Not related to the PR: This should be `invalid argument`, as `params` are not provided.

thiagowfx: They are provided:

```
        await execute_command(
            websocket, {
                "method": "network.removeIntercept",
                "params": {
                    "intercept": "00000000-0000-0000-0000-000000000000",
                },
            })
```

thiagowfx: Just expand the diff

tests/script/test_shared_id.py
```diff
@@ -166,7 +167,13 @@ async def test_sharedId_in_different_navigable(websocket, context_id, html):
...
     await goto_url(websocket, context_id, html("some other page"))
 
-    with pytest.raises(Exception) as exception_info:
+    with pytest.raises(
+            Exception,
+            match=re.compile(
+                str({
+                    "error": "no such node",
+                    "message": 'SharedId ".*" belongs to different document. Current document is .*'
+                }))):
```
sadym-chromium, line 176: I'd prefer to have a more readable verification here. What about the following?

```suggestion
            check=lambda e: {
                "error": "no such node",
                "message": AnyContains("different document")
            } == exception_info.value.args[0]
        )
```

tests/script/test_shared_id.py
```diff
@@ -209,7 +211,12 @@ async def test_sharedId_not_found(websocket, context_id, html):
 
     shared_id = result["result"]["sharedId"] + "9999"
 
-    with pytest.raises(Exception) as exception_info:
+    with pytest.raises(Exception,
```
sadym-chromium, line 214: same here

sadym-chromium (approved):
LGTM, but consider changing the regex to `check` lambda

tests/script/test_shared_id.py
thiagowfx: Where did you get the `check` argument from?

https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-raises

It doesn't exist

thiagowfx: The only way to do what you're suggesting is to revert to how it was before whenever a regex is involved.

sadym-chromium: Indeed it's not in the API, my bad

sadym-chromium (approved):

PR #1400, opened 2023-10-06
fix: unblock event queue network request is blocked

sadym-chromium:
1. Unblock event queue when the network event is blocked.
2. Update e2e so that it does not rely on CDP event.

Issue #1332, opened 2023-09-21
`navigation` value should always be present, even for wait=none


sadym-chromium:
`BrowsingContext.navigate` and `BrowsingContext.reload` should have `navigation` value even in case of `wait: none`.

In BiDi, the [`await a navigation](https://w3c.github.io/webdriver-bidi/#await-a-navigation) step generates a new `navigation` UUID for each navigation disregard of the wait condition. Even in case of same document [fragment navigation](https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigate-fragid).

In CDP:
* [Page.navigate](https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-navigate)'s `loaderId ` is omitted in case of same-document navigation, as the previously committed loaderId would not change.
* [Page.reload](https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-reload) does not provide any `loaderId`. The `loaderId` can be retrieved only from the `Page.lifecycleEvent` of from the network events.

Currently, the `navigation` is not provided in case of reloading with wait `none` or same-document navigation. 

WPT tests: https://wpt.fyi/results/webdriver/tests/bidi/browsing_context/navigate/hash.py?label=master&label=experimental&aligned

thiagowfx:
Upstream: https://github.com/w3c/webdriver-bidi/issues/532

PR #1409, opened 2023-10-09
feat: unblock event queue when network events are blocked

thiagowfx:
Furthermore, update some E2E tests so that they do not rely on CDP events.

Bug #644, #1080
Co-authored-by: Maksim Sadym <sadym@chromium.org>

Re-created from #1400.

### DIFF
--- a/src/bidiMapper/domains/context/CdpTarget.ts
+++ b/src/bidiMapper/domains/context/CdpTarget.ts
@@ -117,19 +117,6 @@ export class CdpTarget {
   }
 
   /**
-   * Calls `Fetch.disable` followed by `Fetch.enable`.
-   * The order is important. Do not use `Promise.all`.
-   *
-   * This is necessary because `Fetch.disable` removes all intercepts.
-   * In a situation where there are two or more intercepts and one of them is
-   * removed, the `Fetch.enable` call will restore the remaining intercepts.
-   */
-  async fetchApply() {
-    await this.fetchDisable();
-    await this.fetchEnable();
-  }
-
-  /**
    * Enables all the required CDP domains and unblocks the target.
    */
   async #unblock() {
@@ -143,7 +130,7 @@ export class CdpTarget {
         // XXX: #1080: Do not always enable the network domain globally.
         // TODO: enable Network domain for OOPiF targets.
         this.#cdpClient.sendCommand('Network.enable'),
-        this.fetchApply(),
+        this.fetchEnable(),
         this.#cdpClient.sendCommand('Target.setAutoAttach', {
           autoAttach: true,
           waitForDebuggerOnStart: true,
--- a/src/bidiMapper/domains/context/CdpTarget.ts
+++ b/src/bidiMapper/domains/context/CdpTarget.ts
@@ -130,6 +130,7 @@ export class CdpTarget {
         // XXX: #1080: Do not always enable the network domain globally.
         // TODO: enable Network domain for OOPiF targets.
         this.#cdpClient.sendCommand('Network.enable'),
+        // XXX: #1080: Do not always enable the fetch domain globally.
         this.fetchEnable(),
         this.#cdpClient.sendCommand('Target.setAutoAttach', {
           autoAttach: true,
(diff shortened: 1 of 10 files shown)


thiagowfx:
It's just missing "continue request", but sending out for review already.

src/bidiMapper/domains/context/CdpTarget.ts
```diff
@@ -143,7 +130,7 @@ export class CdpTarget {
         // XXX: #1080: Do not always enable the network domain globally.
         // TODO: enable Network domain for OOPiF targets.
         this.#cdpClient.sendCommand('Network.enable'),
-        this.fetchApply(),
+        this.fetchEnable(),
```
sadym-chromium, line 133: This would block all the network requests. I'm quite sure we don't need it.

src/bidiMapper/domains/network/NetworkRequest.ts
```diff
@@ -407,7 +414,7 @@ export class NetworkRequest {
     return {
       method: ChromiumBidi.Network.EventNames.BeforeRequestSent,
       params: {
-        ...this.#getBaseEventParams(),
+        ...this.#getBaseEventParams(Network.InterceptPhase.BeforeRequestSent),
```
sadym-chromium, line 417: The same should be done for `network.responseStarted` event, but it looks like we don't have it implemented yet. Correct me if I'm wrong please.

tests/network/test_add_intercept.py
```diff
@@ -334,3 +335,91 @@ async def test_add_intercept_blocks(websocket, context_id, url_patterns,
...
         "type": "event",
     }
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("url_patterns", [
+    [
+        {
+            "type": "string",
+            "pattern": "https://www.example.com/",
```
sadym-chromium, line 345: Can you use fixture instead of `https://www.example.com/`?

thiagowfx: It's not possible to use fixtures in parameters: https://github.com/pytest-dev/pytest/issues/349

tests/network/test_fail_request.py
```diff
@@ -201,8 +206,40 @@ async def test_fail_request_with_auth_required_phase(
...
+                "cookies": [],
+                "headersSize": -1,
+                "bodySize": 0,
+                "timings": ANY_DICT,
+            },
+            "timestamp": ANY_TIMESTAMP,
+        },
+        "type": "event",
+    }
+    network_id_from_bidi_event = network_event_response["params"]["request"][
```
sadym-chromium, line 235: Same here. You can get network id from navigation result

tests/network/test_add_intercept.py
thiagowfx: We could expose a global string instead of a fixture. Or we could break the test down into multiple functions so that it is not parameterized anymore. Otherwise it's not really possible at the moment.

tests/network/test_fail_request.py
```diff
@@ -112,26 +110,33 @@ async def test_fail_request_twice(websocket, context_id, example_url):
...
+                "headersSize": -1,
+                "bodySize": 0,
+                "timings": ANY_DICT,
+            },
+            "timestamp": ANY_TIMESTAMP,
         },
         "type": "event",
     }
-    network_id = event_response["params"]["params"]["networkId"]
+    network_id = event_response["params"]["request"]["request"]
```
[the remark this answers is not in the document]
thiagowfx: We cannot really do so until #1332 is fixed.

tests/network/test_fail_request.py
thiagowfx: Ditto: Blocked until #1332 is fixed.

tests/network/test_add_intercept.py
thiagowfx: Actually, I forgot to mention a third possibility: We could use a pytest plug-in that allows so. Do you have a preference? See the upstream issue above.

PR #1415, opened 2023-10-11
[CI TEST]: enable response_completed WPT test

Lightning00Blade:


wpt-metadata/chromedriver/headless/webdriver/tests/bidi/network/response_completed/response_completed.py.ini
```diff
@@ -0,0 +1,8 @@
+[response_completed.py]
+  [test_response_status[101-Switching Protocols\]]
+    disabled: Implementation fails
```
thiagowfx, line 3: `implementation-status` _should_ be enough

thiagowfx: ...but it **is not**

PR #1423, opened 2023-10-12
test: manual local http server runner

sadym-chromium:
Add script to run local http server, so that it can be run manually.

Bug #575


README.md
```diff
@@ -270,7 +270,25 @@ Use the `PORT` environment variable to connect to another port:
...
+### Local http server
+
+E2e tests use local http
+server [`pytest-httpserver`](https://pytest-httpserver.readthedocs.io/), which is run
+automatically with the tests. However,
+sometimes it is useful to run the http server outside the test
+case, for example for manual debugging. This can be done by running:
+
+```sh
+python tools/run_local_http_server.py
```
thiagowfx, line 282:
```suggestion
python3 tools/run_local_http_server.py
```

tests/conftest.py
```diff
@@ -25,6 +25,16 @@
 from tools.local_http_server import LocalHttpServer
 
 
+@pytest_asyncio.fixture
```
thiagowfx, line 28: defining the same fixture twice

README.md
```diff
@@ -270,7 +270,25 @@ Use the `PORT` environment variable to connect to another port:
 PORT=8081 npm run e2e
 ```
 
-#### Examples
+### Local http server
+
+E2e tests use local http
```
thiagowfx, line 275:
```suggestion
E2E tests use local http
```

tools/run_local_http_server.py
```diff
@@ -0,0 +1,33 @@
...
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+import os
+import sys
+
+from pytest_httpserver import HTTPServer
+
+# Current directory is noit a module, so to import `LocalHttpServer`, it's path
```
thiagowfx, line 21:
```suggestion
# Current directory is not a module, so to import `LocalHttpServer`, its path
```

tools/run_local_http_server.py
```diff
@@ -0,0 +1,33 @@
...
+# has to be added to `sys.path`. It is done relative to this file's directory.
+# The `flake8` is disabled for this reason.
+sys.path.append(os.path.dirname(__file__) + '/../tests/tools/')
```
thiagowfx, line 24: Please use the pathlib library here (`from pathlib import Path`) to append to paths

tools/run_local_http_server.py
```diff
@@ -0,0 +1,33 @@
...
+
+import local_http_server  # noqa: E402
+
+httpserver = HTTPServer()
+httpserver.start()
+local_http_server = local_http_server.LocalHttpServer(httpserver)
+print("qweqwe Local http server started...")
```
thiagowfx, line 31: delete qweqwe

thiagowfx (changes requested):
.

tests/conftest.py
sadym-chromium: Indeed. I screw up with rebasing... Fixed.

tools/run_local_http_server.py
sadym-chromium: done

tools/run_local_http_server.py
sadym-chromium: oops )

tools/run_local_http_server.py
```diff
@@ -0,0 +1,33 @@
...
+
+import sys
+from pathlib import Path
+
+from pytest_httpserver import HTTPServer
+
+# Current directory is not a module, so to import `LocalHttpServer`, its path
+# has to be added to `sys.path`. It is done relative to this file's directory.
+# The `flake8` is disabled for this reason.
+sys.path.append(str(Path(__file__).resolve().parent / '../tests/tools/'))
```
thiagowfx, line 24: Are you sure you need `..` if you added `parent`? Or are you looking for `.parent.parent`?

tools/run_local_http_server.py
```diff
@@ -0,0 +1,33 @@
...
+
+import local_http_server  # noqa: E402
+
+httpserver = HTTPServer()
+httpserver.start()
+local_http_server = local_http_server.LocalHttpServer(httpserver)
+print("Local http server started...")
```
thiagowfx, line 31: Just a slightly improved suggestion:

```
print(f"""
Local http server started...
  - 200: {...}
  - 301 / permanent redirect: {...}
"""

And maybe also print in the first line where the server started? E.g. localhost:8080 or similar.
```

thiagowfx (approved):
LGTM. Giving approval as the requested changes are straightforward but happy to take another look if you want

tools/run_local_http_server.py
sadym-chromium: `parent` refers to the file's directory. `tools` can be referenced either by `.parents[1]`, or by `../` prefix. Do you have a strong preference of one over another way?

thiagowfx (approved):

tools/run_local_http_server.py
thiagowfx: The appeal of the pathlib library is to be OS agnostic, so `.parent` is always better than `../` - IIRC the latter does not work on Windows. LGTM as it is now.

tools/run_local_http_server.py
sadym-chromium: Write where the server runs requires adding it in the API, and I don't think it is needed for tests. The developer can easily see the address part in the lines below.

thiagowfx (approved):

PR #1433, opened 2023-10-16
build: update deps 

Lightning00Blade:
This also fixes formatting command as it was broken, also it seems that the `Pre-commit` Action was broken as well, so fixed that as well.


thiagowfx (approved):
Makes sense, because we changed it to `pre-push`. Good catch, thanks!

thiagowfx:
Note: There are some merge conflicts

thiagowfx (approved):

PR #1434, opened 2023-10-16
misc: specify Browser CDP Session to be used

sadym-chromium:
reparation for multiple session support. http://go/webdriver:bidi-multiple-sessions-proposal

**_Step 2._** Specify Browser CDP Session. 

To support multiple BiDi sessions in a single Mapper Tab, the Mapper instance should be aware of the Browser CDP session it has to use. This PR:
* Adds a method `createBrowserSession` on the CdpConnection.
* Removes `browserClient` from the `CdpConnection`, as it can be different for different instances.
* For the Mapper instance and it's components, specify which exactly `browserClient` should be used.


src/cdp/CdpConnection.ts
```diff
@@ -45,14 +44,14 @@ export class CdpConnection implements ICdpConnection {
...
+  readonly #mainBrowserCdpClient: ICdpClient;
   readonly #transport: ITransport;
 
-  /** The CdpClient object attached to the root browser session. */
-  readonly #browserCdpClient: CdpClient;
-
   /** Map from session ID to CdpClient. */
-  readonly #sessionCdpClients = new Map<Protocol.Target.SessionID, CdpClient>();
-
+  readonly #sessionCdpClients = new Map<
```
thiagowfx, line 51: Can we document here what `undefined` means?

src/cdp/CdpConnection.ts
```diff
@@ -162,4 +165,12 @@ export class CdpConnection implements ICdpConnection {
       }
     }
   };
+
+  #createCdpClient(
+    sessionId: Protocol.Target.SessionID | undefined
+  ): CdpClient {
+    const cdpClient = new CdpClient(this, sessionId);
+    this.#sessionCdpClients.set(sessionId, cdpClient);
```
thiagowfx, line 173: Is it possible (however unlikely) this could end up replacing an existing session?

What about `undefined`?

If this is possible, consider checking if it exists before adding it.
But my impression / understanding is that "this should never happen".

thiagowfx (approved):

src/cdp/CdpConnection.ts
sadym-chromium: Added comment

src/cdp/CdpConnection.ts
sadym-chromium: Undefined session ID points to the main main browser session. Added comment.

sadym-chromium (commented):
Added comments

PR #1436, opened 2023-10-16
test: Add a `/hang_forever` and `/401` to local http server

sadym-chromium:
* `/hang_forever` simply hangs forever.
* `/401`, if there is no `Authorization` header, returns an authentication request. Otherwise, returns parsed credentials with code `200`.

Bug #575

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -155,16 +155,22 @@ def another_example_url():
 
 
 @pytest.fixture
-def auth_required_url(local_server):
+def auth_required_url():
     """Return a URL that requires authentication (status code 401)."""
-    return local_server.url_basic_auth()
+    # TODO: Switch to a local server so that it works off-line.
+    # All of these URLs work, just pick one.
+    # url = "https://authenticationtest.com/HTTPAuth/"
+    # url = "http://the-internet.herokuapp.com/basic_auth"
+    pytest.skip(reason='TODO: Use our own test server.')
+    return "http://httpstat.us/401"
 
 
 @pytest.fixture
-def hang_url(local_server):
+def hang_url():
     # TODO: Start a local server alongside tests that use this fixture.
     """Return a URL that hangs forever."""
-    return local_server.url_hang_forever()
+    pytest.skip(reason='TODO: Use our own test server.')
+    return "http://127.0.0.1:5000/hang"
 
 
 @pytest.fixture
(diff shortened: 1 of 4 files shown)


thiagowfx (approved):

thiagowfx:
Why is the conventional commit linter failing? I tried to close and re-open the issue but it keeps failing.

sadym-chromium:
Added another commit with `/401` handler.

thiagowfx (approved):

thiagowfx:
As you can see, E2E tests fail, so there must be something else that is missing.

If you prefer, feel free to uncommit my changes and go ahead with your original PR, but we'll need to investigate the failures at some point.

sadym-chromium:
> As you can see, E2E tests fail, so there must be something else that is missing.
> 
> If you prefer, feel free to uncommit my changes and go ahead with your original PR, but we'll need to investigate the failures at some point.

I rolled the tests back and will follow up in another PR.

thiagowfx (approved):

thiagowfx:
To reapply later:

```
@pytest.fixture
def auth_required_url(local_server):
    """Return a URL that requires authentication (status code 401)."""
    return local_server.url_basic_auth()


@pytest.fixture
def hang_url(local_server):
    """Return a URL that hangs forever."""
    return local_server.url_hang_forever()
```

```
                "responseStatusText": "UNAUTHORIZED",
```

```
 def url_basic_auth(self) -> str:
        """
        Returns the url for the page with a basic auth (401).
        To test with non-local servers, use one of the following URLs:
            - "https://authenticationtest.com/HTTPAuth/"
            - "http://the-internet.herokuapp.com/basic_auth"
            - "http://httpstat.us/401"
        """
        return self.__http_server.url_for(self.__path_basic_auth)
```

sadym-chromium:
> As you can see, E2E tests fail, so there must be something else that is missing.
> 
> If you prefer, feel free to uncommit my changes and go ahead with your original PR, but we'll need to investigate the failures at some point.

The test failed,  because 

> As you can see, E2E tests fail, so there must be something else that is missing.
> 
> If you prefer, feel free to uncommit my changes and go ahead with your original PR, but we'll need to investigate the failures at some point.

The failing tests were never passing with the remote resource. They were skipped because of the [`pytest.skip(reason='TODO: Use our own test server.')`](https://github.com/GoogleChromeLabs/chromium-bidi/blob/7c8b3edb3db5f0a6c3a95fa460fa6f869d315889/tests/conftest.py#L164). The failing reason is in the problems with tearing the test down and closing the browser.

Issue #1428, opened 2023-10-13
`test_fail_request_completes_new_request_still_blocks` e2e test is flaky


sadym-chromium:
The test is flaky on headful. Example: https://github.com/GoogleChromeLabs/chromium-bidi/actions/runs/6507248949/job/17674295005
This causes false-negative CI runs.

thiagowfx:
Ack

thiagowfx:
The request fails because it is fetching some image: ` 'url': 'data:image/png;ba `. Huh.

PR #1445, opened 2023-10-17
test: add wait=complete to navigation to try to mitigate flakiness

thiagowfx:
Fixed: #1428

### DIFF
--- a/tests/network/test_fail_request.py
+++ b/tests/network/test_fail_request.py
@@ -476,6 +476,7 @@ async def test_fail_request_completes_new_request_still_blocks(
             "params": {
                 "url": example_url,
                 "context": context_id,
+                "wait": "complete",
             }
         })
 
@@ -542,6 +543,7 @@ async def test_fail_request_completes_new_request_still_blocks(
             "params": {
                 "url": example_url,
                 "context": context_id,
+                "wait": "complete",
             }
         })
 


thiagowfx:
```
===================== 50 passed, 1050 deselected in 14.27s =====================
✅ Ran 2 scripts and skipped 2 in 15.5s.
^Cnpm run e2e --watch -- tests/network/test_{continue,fail}_request.py -vv -s -  3.12s user 2.48s system 27% cpu 20.017 total
tperrotta@tperrotta-macbookpro ~/Projects/chromium-bidi (git)-[1428-flaky-fail-request]
% npm run e2e --watch -- tests/network/test_{continue,fail}_request.py -vv -s -k test_fail_request_completes_new_request_still_blocks --flake-finder
```

tests/network/test_fail_request.py
```diff
@@ -476,6 +476,7 @@ async def test_fail_request_completes_new_request_still_blocks(
             "params": {
                 "url": example_url,
                 "context": context_id,
+                "wait": "complete",
```
Lightning00Blade, line 479: Can we use the helper `goto_url` that way we do the same default everywhere.

Lightning00Blade: Oops, I think we can't do that as that waits for the command to finish, should we allow the helper to not wait for the command? and use it in all instances of the the `browsingContext.navigate`?

thiagowfx: `goto_url` does not work here because it skips too many events

thiagowfx: I don't oppose your suggestion but it will require a lot of refactoring, and it's out of scope of the intent of this PR. Maybe in a separate one

sadym-chromium (approved):

PR #1449, opened 2023-10-17
test: use local http server when possible

sadym-chromium:
Switch from `example.com` to local http server when possible.

Out of scope: 
* hardcoded addresses in url patterns.

Bug: #575

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -135,42 +135,38 @@ def url_cross_origin(request):
     'data:text/html,<h2>child page</h2>',  # Data URL: Cross-origin
 ])
 def url_all_origins(request):
-    """Return an URL exhaustively, including same-origin and cross-origin."""
+    """Return a URL exhaustively, including same-origin and cross-origin."""
     return request.param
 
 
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -135,42 +131,38 @@ def url_cross_origin(request):
...
 @pytest.fixture
-def auth_required_url():
+def auth_required_url(local_server: LocalHttpServer):
     """Return a URL that requires authentication (status code 401)."""
-    # TODO: Switch to a local server so that it works off-line.
     # All of these URLs work, just pick one.
     # url = "https://authenticationtest.com/HTTPAuth/"
     # url = "http://the-internet.herokuapp.com/basic_auth"
     pytest.skip(reason='TODO: Use our own test server.')
-    return "http://httpstat.us/401"
```
thiagowfx, line 165: Please add this URL to the list above

thiagowfx (approved):

tests/conftest.py
thiagowfx: Done

Issue #1407, opened 2023-10-09
`test_input_performActions* ` e2e test is flaky


sadym-chromium:
e2e tests of input are flaky. Example:

`test_input_performActionsEmitsDragging`:
https://github.com/GoogleChromeLabs/chromium-bidi/actions/runs/6448243919/job/17509797173?pr=1403

Although the output is really hard to read, the difference is expected and missing the following part at the beginning: 
```
      {
        type: 'object',
        value: [
          [
            'event',
            {
              type: 'string',
              value: 'mousemove',
            },
          ],
          [
            'button',
            {
              type: 'number',
              value: 0,
            },
          ],
          [
            'buttons',
            {
              type: 'number',
              value: 0,
            },
          ],
          [
            'clientX',
            {
              type: 'number',
              value: 630,
            },
          ],
          [
            'clientY',
            {
              type: 'number',
              value: 336,
            },
          ],
        ],
      },
```

Another example:

`test_input_performActionsEmitsPointerEvents`:
Example: https://github.com/GoogleChromeLabs/chromium-bidi/actions/runs/6548857552/job/17784481994?pr=1443
Related: https://github.com/GoogleChromeLabs/chromium-bidi/issues/1407

jrandolf:
This is a result of headful mode. The solution is to disable the test in headful.

sadym-chromium:
> This is a result of headful mode. The solution is to disable the test in headful.

Cannot we make the test more flexible, so that it tolerates the fluctuations?

sadym-chromium:
friendly ping

sadym-chromium:
another example: https://github.com/GoogleChromeLabs/chromium-bidi/actions/runs/6509030633/job/17679618738

sadym-chromium:
The issue addressed, expectations do not tolerate flakiness anymore

Issue #1452, opened 2023-10-17
`test_input_performActionsEmitsPointerEvents` e2e test is flaky


thiagowfx:
Example: https://github.com/GoogleChromeLabs/chromium-bidi/actions/runs/6548857552/job/17784481994?pr=1443

Related: https://github.com/GoogleChromeLabs/chromium-bidi/issues/1407

thiagowfx:
Closing in favour of https://github.com/GoogleChromeLabs/chromium-bidi/issues/1407 (merging).

PR #1453, opened 2023-10-17
ci: ignore input tests in headful mode

thiagowfx:
They are extremely flaky in CI.

Tested:
  - Before: 268 passed, 11 skipped, 3 xfailed
  - After: 262 passed, 11 skipped, 2 xfailed

Fixed: #1407, #1452

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
       ]
     },
     "e2e-headful": {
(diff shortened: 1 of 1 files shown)


Lightning00Blade (approved):
But let's keep the issues as the test are still flaky we didn't fix them we just don't run them.

PR #1457, opened 2023-10-18
feat: implement network interception provide response

thiagowfx:
Bug: #644

### DIFF
--- a/src/bidiMapper/domains/network/NetworkProcessor.ts
+++ b/src/bidiMapper/domains/network/NetworkProcessor.ts
@@ -185,7 +185,7 @@ export class NetworkProcessor {
 
     await request.provideResponse(
       fetchId,
-      statusCode!, // TODO: What if the status code is not present?
+      statusCode ?? request.statusCode,
       reasonPhrase,
       responseHeaders,
       body?.value // TODO: Differ base64 / string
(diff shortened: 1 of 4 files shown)


thiagowfx:
This is now ready for review

src/bidiMapper/domains/network/NetworkProcessor.ts
```diff
@@ -163,8 +163,37 @@ export class NetworkProcessor {
...
+
+    // TODO: Set / expand.
+    // ; Step 10. cookies
+    // ; Step 11. credentials
+    const request = this.#networkStorage.getRequest(networkId);
+    assert(request, `Network request with ID ${networkId} doesn't exist`);
+
+    await request.provideResponse(
+      fetchId,
+      statusCode!, // TODO: What if the status code is not present?
```
sadym-chromium, line 188: Provide one from the request, it should be there.

thiagowfx: We still need to provide a fallback in case it is undefined. I am using 200. Is that OK?

sadym-chromium (approved):
One comment regarding the test.

tests/network/test_provide_response.py
```diff
@@ -0,0 +1,569 @@
...
+                    },
+                }, ],
+                "url": example_url,
+                "status": 200,  # TODO: Update status code to 501?
+                "statusText": "Remember to drink water",
+            }),
+            "timestamp": ANY_TIMESTAMP,
+        },
+        "type": "event",
+    }
```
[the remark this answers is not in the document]
thiagowfx: Response contains `'bodySize': 0,`. This seems incorrect.

PR #1464, opened 2023-10-20
chore: use local EsLint/Prettier for hook

Lightning00Blade:
EsLint never fixed the issue that it could auto resolve. Also we had situation where its possible to have 2 different version of EsLint and Prettier to the installed via npm. 
Note: wait for all test to pass + Puppeteer run, before merging.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -85,6 +85,7 @@ module.exports = {
     '@typescript-eslint/prefer-return-this-type': 'warn',
     '@typescript-eslint/require-await': 'warn',
     '@typescript-eslint/restrict-template-expressions': 'off',
+    '@typescript-eslint/return-await': ['error', 'always'], // This is more performant; see https://v8.dev/blog/fast-async.
     '@typescript-eslint/switch-exhaustiveness-check': 'error',
     'func-names': 'error',
     'import/first': 'error',
@@ -99,7 +100,6 @@ module.exports = {
     'no-empty': ['warn', {allowEmptyCatch: true}],
     'no-implicit-coercion': 'error',
     'no-negated-condition': 'error',
-    'no-return-await': 'warn',
     'no-undef': 'error',
     'no-underscore-dangle': 'error',
     'object-shorthand': 'error',
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -88,9 +88,6 @@ repos:
     rev: v0.1.1
     hooks:
       - id: keep-sorted
-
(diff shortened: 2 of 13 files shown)


.eslintrc.js
```diff
@@ -85,6 +85,8 @@ module.exports = {
     '@typescript-eslint/prefer-return-this-type': 'warn',
     '@typescript-eslint/require-await': 'warn',
     '@typescript-eslint/restrict-template-expressions': 'off',
+    // This is more performant; see https://v8.dev/blog/fast-async.
```
thiagowfx, line 88: please add this comment to the right side of the rule, so that keep-sorted is happy

.pre-commit-config.yaml
thiagowfx, line 1: The pre-commit change seems unrelated.

We don't use `local` so that pre-commit is reproducible (always installing the same fixed version). Different people may have different versions of prettier installed locally due to different OSes and package managers (e.g. npm, homebrew, apt, etc).

Can you revert this part?

Lightning00Blade: Why is this not reproducible if we have a `package-lock.json`, `npx` run the first `node_module/bin` ?
This way we have a single dependency and not 2 that can become out of sync. 
Also I am not sure currently it works correctly as in the mirror you are told you need to define `additional_dependencies` https://github.com/pre-commit/mirrors-eslint. 
And this pre-hook never auto-formats.

.pre-commit-config.yaml
```diff
@@ -122,9 +118,20 @@ repos:
...
+  - repo: local
+    hooks:
+      - id: eslint
+        name: EsLint format
+        entry: 'npx eslint --fix'
+        language: system
+        verbose: true
+        files: \.[jt]s$
+        types: [file]
+  - repo: local
```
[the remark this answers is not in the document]
thiagowfx: In fact, even `id: copyright` could be consolidated here.

thiagowfx (approved):

.pre-commit-config.yaml
```diff
@@ -111,20 +114,24 @@ repos:
...
-    hooks:
       - id: eslint
+        name: EsLint format
+        entry: 'npx eslint --fix'
+        language: system
+        verbose: true
+        files: \.[jt]s$
+        types: [file]
+
+      - id: eslint
```
thiagowfx, line 132: eslint -> prettier

thiagowfx (changes requested):
See open comment

thiagowfx (approved):

PR #1465, opened 2023-10-23
ci: make --watch work with npm run wpt

thiagowfx:
Watch WPT test files

### DIFF
--- a/package.json
+++ b/package.json
@@ -149,7 +149,8 @@
     "wpt": {
       "command": "tools/run-wpt.mjs",
       "files": [
-        "tools/run-wpt.mjs"
+        "tools/run-wpt.mjs",
+        "wpt/**/*.py"
       ],
       "dependencies": [
         "rollup"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -158,7 +159,9 @@
     "wpt-all": {
       "command": "tools/runWPTAll.sh",
       "files": [
-        "tools/runWPTAll.sh"
+        "tools/run-wpt.mjs",
+        "tools/runWPTAll.sh",
+        "wpt/**/*.py"
```
Lightning00Blade, line 164: Scope it lower to `wpt/webdriver/tests/**` that way it doesn't take thousand of files to watch.

thiagowfx: It's not comprehensive enough.

For example: We also care about `tools/webdriver/webdriver/**/*.py`.

I could change to match these two entries at least, but there could be others I'm unaware of.

thiagowfx: Done.

Lightning00Blade (approved):

PR #1466, opened 2023-10-23
feat: network request: prioritize response status code over extraInfo

thiagowfx:
Bug: #644

### DIFF
--- a/src/bidiMapper/domains/network/NetworkRequest.ts
+++ b/src/bidiMapper/domains/network/NetworkRequest.ts
@@ -259,7 +259,7 @@ export class NetworkRequest {
         mimeType: '',
         // TODO: populate.
         bytesReceived: 0,
-        headersSize: computeResponseHeadersSize(headers),
+        headersSize: computeResponseHeadersSize(headers), // TODO: Should this be computeRequestHeadersSize?
         // TODO: consider removing from spec.
         bodySize: 0,
         // TODO: consider removing from spec.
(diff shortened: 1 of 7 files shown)


thiagowfx:
I set the `response_completed` tests to `[PASS, FAIL]` since they are intermittently flaky.

src/bidiMapper/domains/network/NetworkRequest.ts
```diff
@@ -361,8 +361,8 @@ export class NetworkRequest {
   /** Returns the HTTP status code associated with this request if any. */
   get statusCode(): number {
     return (
-      this.#response.extraInfo?.statusCode ??
       this.#response.info?.response.status ??
+      this.#response.extraInfo?.statusCode ??
```
Lightning00Blade, line 365: This is wrong... The extra info has the correct one if we return for example `304` not modified response.
Now it return 200 and brakes Puppeteer test. 
We need to find a way to compute which is correct.

Lightning00Blade: https://github.com/puppeteer/puppeteer/actions/runs/6706355305/job/18222699456?pr=11284

thiagowfx: For network interception, this is correct.

We'll need to add some extra conditions in that method then. If request isBlocked, keep it like this. Otherwise do it the other way around.

Issue #1353, opened 2023-09-28
Extend `AddPreloadScript` with `ContextId` parameter


sadym-chromium:
The parameter was added in https://github.com/w3c/webdriver-bidi/pull/517 needs to be implemented, so it can be used in Puppeteer.

WPT tests will be added in https://github.com/web-platform-tests/wpt/pull/41932.

- [X] Invalid inputs
- [X] PreloadScripts are ran supplied Contexts
- [X] PreloadScripts are not ran for newly created BrowsingContexts creation (via `browsingContext.create`)
- ~[ ] PreloadScripts are ran for Workers, fenced frames (Connected Contexts but separate CDP targets)~

thiagowfx:
cc @Lightning00Blade

PR #1471, opened 2023-10-24
chore: parse addPreloadScript params

Lightning00Blade:
WPT should cover this case https://github.com/web-platform-tests/wpt/pull/41932

Tracked at #1353

### DIFF
--- a/src/bidiMapper/domains/script/ScriptProcessor.ts
+++ b/src/bidiMapper/domains/script/ScriptProcessor.ts
@@ -52,15 +52,16 @@ export class ScriptProcessor {
     params: Script.AddPreloadScriptParameters
   ): Promise<Script.AddPreloadScriptResult> {
     if (params.contexts) {
+      // Remove once https://github.com/google/cddlconv/issues/16 is implemented
       if (params.contexts.length === 0) {
-        throw new InvalidArgumentException('Invalid empty contexts list.');
+        throw new InvalidArgumentException('Contexts list is empty.');
       }
 
       for (const contextId of params.contexts) {
         const context = this.#browsingContextStorage.getContext(contextId);
         if (!context.isTopLevelContext()) {
           throw new InvalidArgumentException(
-            `addPreloadScript only supported on the top-level contexts, ${contextId} given.`
+            `Non top-level contexts given '${contextId}'.`
           );
         }
       }
--- a/src/bidiMapper/domains/script/ScriptProcessor.ts
+++ b/src/bidiMapper/domains/script/ScriptProcessor.ts
@@ -52,7 +52,7 @@ export class ScriptProcessor {
     params: Script.AddPreloadScriptParameters
   ): Promise<Script.AddPreloadScriptResult> {
     if (params.contexts) {
-      // Remove once https://github.com/google/cddlconv/issues/16 is implemented
+      // XXX: Remove once https://github.com/google/cddlconv/issues/16 is implemented
       if (params.contexts.length === 0) {
         throw new InvalidArgumentException('Contexts list is empty.');
       }
@@ -61,7 +61,7 @@ export class ScriptProcessor {
         const context = this.#browsingContextStorage.getContext(contextId);
         if (!context.isTopLevelContext()) {
           throw new InvalidArgumentException(
-            `Non top-level contexts given '${contextId}'.`
+            `Non top-level context '${contextId}' given.`
           );
         }
       }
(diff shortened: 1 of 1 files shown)


src/bidiMapper/domains/script/ScriptProcessor.ts
```diff
@@ -50,6 +51,21 @@ export class ScriptProcessor {
   async addPreloadScript(
     params: Script.AddPreloadScriptParameters
   ): Promise<Script.AddPreloadScriptResult> {
+    if (params.contexts) {
+      if (params.contexts.length === 0) {
```
OrKoN, line 55: if zod validates this, we could remove this check

OrKoN (approved):

src/bidiMapper/domains/script/ScriptProcessor.ts
Lightning00Blade: Sadly the auto-generated `zod` validator don't validate the one or more in `[+Object]` from CDDL, I will take a peak if its possible to implement. 

CC @jrandolf

thiagowfx: Can we file a FR here? https://github.com/google/cddlconv

I swear I had filed one already, but I cannot find it.

src/bidiMapper/domains/script/ScriptProcessor.ts
```diff
@@ -50,6 +51,21 @@ export class ScriptProcessor {
...
+        throw new InvalidArgumentException('Invalid empty contexts list.');
```
thiagowfx, line 56: How about: "Contexts list is empty".

Invalid is implicit in the error type, we don't need to repeat it in the error message

src/bidiMapper/domains/script/ScriptProcessor.ts
```diff
@@ -50,6 +51,21 @@ export class ScriptProcessor {
...
+      }
+
+      for (const contextId of params.contexts) {
+        const context = this.#browsingContextStorage.getContext(contextId);
+        if (!context.isTopLevelContext()) {
+          throw new InvalidArgumentException(
+            `addPreloadScript only supported on the top-level contexts, ${contextId} given.`
```
thiagowfx, line 63:
```suggestion
            `addPreloadScript is only supported on top-level contexts, but was '${contextId}' given.`
```

thiagowfx (commented):
Add at least one test? It could be E2E. Unless it's already covered in WPT.

src/bidiMapper/domains/script/ScriptProcessor.ts
Lightning00Blade: Filed - https://github.com/google/cddlconv/issues/16

thiagowfx: Thanks

Lightning00Blade:
WPT test will cover all the use cases for invalid, I would prefer not to repeat them.

src/bidiMapper/domains/script/ScriptProcessor.ts
```diff
@@ -50,6 +51,22 @@ export class ScriptProcessor {
...
+      if (params.contexts.length === 0) {
+        throw new InvalidArgumentException('Contexts list is empty.');
+      }
+
+      for (const contextId of params.contexts) {
+        const context = this.#browsingContextStorage.getContext(contextId);
+        if (!context.isTopLevelContext()) {
+          throw new InvalidArgumentException(
+            `Non top-level contexts given '${contextId}'.`
+          );
```
thiagowfx, line 65:
```suggestion
            `Non top-level context '${contextId}' given.`
          );
```

thiagowfx (approved):

src/bidiMapper/domains/script/ScriptProcessor.ts
```diff
@@ -50,6 +51,22 @@ export class ScriptProcessor {
   async addPreloadScript(
     params: Script.AddPreloadScriptParameters
   ): Promise<Script.AddPreloadScriptResult> {
+    if (params.contexts) {
+      // Remove once https://github.com/google/cddlconv/issues/16 is implemented
```
thiagowfx, line 55:
```suggestion
      // XXX: Remove once https://github.com/google/cddlconv/issues/16 is implemented
```

thiagowfx:
Added Bug: #1353

PR #1475, opened 2023-10-25
feat: addPreloadScripts respects contexts param for old contexts

Lightning00Blade:
This only makes the already created Contexts respect the PreloadScript, the newly create one will still run the script atm.

### DIFF
--- a/src/bidiMapper/domains/script/PreloadScript.ts
+++ b/src/bidiMapper/domains/script/PreloadScript.ts
@@ -79,7 +79,7 @@ export class PreloadScript {
     return this.#channels;
   }
 
-  /** Channels of the preload script. */
+  /** Contexts of the preload script, if any */
   get contexts(): BrowsingContext.BrowsingContext[] | undefined {
     return this.#contexts;
   }
--- a/src/bidiMapper/domains/script/PreloadScript.ts
+++ b/src/bidiMapper/domains/script/PreloadScript.ts
@@ -18,7 +18,7 @@
 
 import type Protocol from 'devtools-protocol';
 
-import type {Script} from '../../../protocol/protocol.js';
+import type {BrowsingContext, Script} from '../../../protocol/protocol.js';
 import {uuidv4} from '../../../utils/uuid.js';
 import type {CdpTarget} from '../context/CdpTarget.js';
 import type {LoggerFn} from '../../../utils/log.js';
@@ -55,6 +55,8 @@ export class PreloadScript {
   readonly #channels: ChannelProxy[];
   /** The script sandbox / world name. */
   readonly #sandbox?: string;
+  /** The script contexts to execute in */
+  readonly #contexts?: BrowsingContext.BrowsingContext[];
 
   get id(): string {
     return this.#id;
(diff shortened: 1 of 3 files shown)


tests/script/test_add_preload_script.py
```diff
@@ -834,3 +834,64 @@ async def test_preloadScript_channel_newContext(websocket,
...
+            "params": {
+                "functionDeclaration": """
+                    () => {
+                        window.FOO = "BAR"
+                    }""",
+                "contexts": [context_id]
+            }
+        })
+
+    # NAvigate both contexts to trigger PreloadScripts
```
thiagowfx, line 865:
```suggestion
    # Navigate both contexts to trigger PreloadScripts
```

src/bidiMapper/domains/script/ScriptProcessor.ts
```diff
@@ -51,6 +52,7 @@ export class ScriptProcessor {
   async addPreloadScript(
     params: Script.AddPreloadScriptParameters
   ): Promise<Script.AddPreloadScriptResult> {
+    const contexts = new Set<BrowsingContextImpl>();
```
thiagowfx, line 55: Be careful here: The intention of the `Set` is likely to deduplicate contexts, but "BrowsingContextImpl" is an opaque object. The deduplication would only happen if it was e.g. via context id (a simple string).

Side effect: Let's add one test that intentionally passes the same context twice, and verify that the script is only performed once. Something like this:

```
.addPreloadScript(script="window.count = (window.count ?? 0) + 1", contexts=[context_id, context_id]);

Expected: I _think_ it should be equal to 1, as it should be deduplicated. But the way the current implementation is done I believe it would equal 2.
```

src/bidiMapper/domains/script/ScriptProcessor.ts
```diff
@@ -70,11 +74,16 @@ export class ScriptProcessor {
...
-        .map((context) => context.cdpTarget)
-    );
+    const cdpTargets =
+      contexts.size === 0
+        ? new Set<CdpTarget>(
+            this.#browsingContextStorage
+              .getTopLevelContexts()
+              .map((context) => context.cdpTarget)
+          )
+        : new Set<CdpTarget>(
```
thiagowfx, line 84: The same issue applies here. If "contexts" contains the same context twice, then the same cdp target will end up being included twice, which is not desired.

If we do the fix in the first set (`contexts`), then no AI is needed here. It's also possible to do the deduplication here instead. But I think it's cleaner to do it above.

src/bidiMapper/domains/script/PreloadScript.ts
```diff
@@ -69,13 +71,19 @@ export class PreloadScript {
...
     this.#sandbox = params.sandbox;
+    this.#contexts = params.contexts;
   }
 
   /** Channels of the preload script. */
   get channels(): ChannelProxy[] {
     return this.#channels;
   }
 
+  /** Channels of the preload script. */
```
thiagowfx, line 82: Copy & paste, please update the comment

src/bidiMapper/domains/script/PreloadScript.ts
```diff
@@ -55,6 +55,8 @@ export class PreloadScript {
   readonly #channels: ChannelProxy[];
   /** The script sandbox / world name. */
   readonly #sandbox?: string;
+  /** The script contexts to execute in */
```
thiagowfx, line 58:
```suggestion
  /** The browsing contexts to execute the preload scripts in, if any. */
```

thiagowfx (approved):

OrKoN:
@Lightning00Blade please don't wait on my review

PR #1477, opened 2023-10-25
ci: speed up prettier/eslint pre-commit hooks and use types instead of files

thiagowfx:
c.f. https://github.com/pre-commit/mirrors-prettier/blob/main/.pre-commit-hooks.yaml and https://github.com/pre-commit/identify/blob/main/identify/extensions.py

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -127,7 +127,7 @@ repos:
       # ESLint may update code style (formatting) so keep prettier last.
       - id: prettier
         name: prettier
-        entry: 'npx prettier --check .'
+        entry: 'npx prettier --write --ignore-unknown --check .'
         language: system
         files: \.[jt]s$
-        types: [file]
+        types: [text]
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -108,26 +108,24 @@ repos:
...
         language: system
         verbose: true
-        files: \.[jt]s$
-        types: [file]
+        types: [javascript, ts]
       # ESLint may update code style (formatting) so keep prettier last.
       - id: prettier
         name: prettier
-        entry: 'npx prettier --check .'
+        entry: 'npx prettier --write --ignore-unknown --check .'
```
Lightning00Blade, line 129: Do we need the `.` at the end? I think that will run all files Prettier can format in the repo? We may if we want to format `.yml`, `.json` etc.

Lightning00Blade (commented):
Please verify that EsLint is running only once as I had issue when it ran one time per file changed

.pre-commit-config.yaml
thiagowfx: No. Removed.

thiagowfx: Yes, that's OK. Markdown, YAML, etc. Whatever prettier supports.

thiagowfx:
I had made a mistake. It's `types_or`, not `types`. Fixed.

Lightning00Blade (approved):

Issue #1182, opened 2023-08-21
Python / `pip`: Code scanning alerts and environment management


thiagowfx:
https://github.com/GoogleChromeLabs/chromium-bidi/security/code-scanning has some suggestions to add hashes to `requirements.txt` (`pip`) for python, similarly to how we do for github action checkouts.

In order to effectively manage this, we should consider adopting some python environment manager. A few popular choices include `pipenv`, `poetry`, `tox`. At a quick glance, `poetry` seems to be a sensible choice.

Note: `pyenv` is just a version manager, and `pipx` is just a runner.

Originally reported by @Lightning00Blade via Chat.

thiagowfx:
This is blocked on https://github.com/jwodder/anys/pull/3.

thiagowfx:
Trying out with `pipenv` here: https://github.com/GoogleChromeLabs/chromium-bidi/pull/1480.

thiagowfx:
A list of resources:

- https://github.com/actions/setup-python/blob/65d7f2d534ac1bc67fcd62888c5f4f3d2cb2b236/docs/advanced-usage.md#caching-packages
- https://github.com/speced/bikeshed/blob/main/Pipfile
- https://docs.pytest.org/en/latest/reference/customize.html#configuration-file-formats

thiagowfx:
This is technically closed but I'll keep it open for two reasons:

- [ ] find a way to install / bootstrap `pipenv` deterministically (with pinning / hash)
- [ ] evaluate the other frameworks (especially `poetry` and `tox`)

OrKoN:
Closing for now

PR #1480, opened 2023-10-25
build: migrate requirements.txt to pipenv / Pipfile

thiagowfx:
Fixed: #1182


Lightning00Blade (approved):

OrKoN (approved):

thiagowfx:
Note: If this ends up breaking something, feel free to revert it. The use of `pipenv` is an experiment. We could have equally chosen `poetry` or `tox`, but let's see how this goes first for a few days / weeks at least.

PR #1484, opened 2023-10-26
feat(network intercept): implement continue with auth (cont.)

thiagowfx:
Bug: #644


src/bidiServer/WebSocketServer.ts
thiagowfx, line 1: Please ignore this file. It will be submitted as part of https://github.com/GoogleChromeLabs/chromium-bidi/pull/1490 instead

OrKoN (approved):

PR #1485, opened 2023-10-26
test(network intercept): remove tests that use cdp events, prefer bidi

thiagowfx:
Bug: #644


thiagowfx:
...as per @sadym-chromium's request.

thiagowfx:
Note that all removed tests have corresponding tests that use BiDi events instead. We are **not** losing coverage with this PR.

OrKoN (approved):

PR #1486, opened 2023-10-26
ci(python): install pipenv with apt instead of pip

thiagowfx:
So that the `pipenv` installation itself is pinned / reproducible.

Bug: #1182


thiagowfx:
This will not work due to some python mismatch config in Ubuntu: https://stackoverflow.com/a/74475168/1745064 :(

PR #1490, opened 2023-10-27
feat(logging): pretty print received and sent bidi server messages

thiagowfx:
Example: Screenshot 2023-10-27 at 11 05 56

Instead of having the output message string in a single line.

### DIFF
--- a/src/bidiServer/WebSocketServer.ts
+++ b/src/bidiServer/WebSocketServer.ts
@@ -156,10 +156,13 @@ export class WebSocketServer {
         }
 
         const plainCommandData = message.utf8Data;
-        try {
-          debugRecv(JSON.parse(plainCommandData));
-        } catch {
-          debugRecv(plainCommandData);
+
+        if (debugRecv.enabled) {
+          try {
+            debugRecv(JSON.parse(plainCommandData));
+          } catch {
+            debugRecv(plainCommandData);
+          }
         }
 
         // Try to parse the message to handle some of BiDi commands.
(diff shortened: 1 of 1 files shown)


src/bidiServer/WebSocketServer.ts
```diff
@@ -157,7 +156,16 @@ export class WebSocketServer {
         }
 
         const plainCommandData = message.utf8Data;
-        debugRecv(plainCommandData);
+        try {
+          const plainCommandDataPrettyPrint = JSON.stringify(
```
Lightning00Blade, line 160: No need to `strigify` it back I think, let the debug handle it.

Lightning00Blade (approved):
LGTM, note that this will have have some marginal impact on performance as you always parse the string, I am not sure where this is used, but I think we only use it for test, Puppeteer and ChromeDriver don't need it, right? If we use it in ChromeDriver we should think of mitigating it.

thiagowfx:
> performance

Hmmm, maybe we could do something like "if (logging_on) {..." to skip the parsing when there is no logging.

src/bidiServer/WebSocketServer.ts
thiagowfx: Very good, it even adds colors to the JSON output!

thiagowfx:
https://www.npmjs.com/package/debug#checking-whether-a-debug-target-is-enabled

thiagowfx:
re: performance: updated, PTAL

PR #1494, opened 2023-10-27
feat(network intercept): specify BeforeRequestSent whenever AuthRequi…

thiagowfx:
…red is specified

The BiDi spec supports "authRequired" by itself, however CDP requires both events.

Bug: #644


thiagowfx:
The only part missing is to call "continueWithRequest". We cannot do this reliably until "network.responseStarted" is implemented though. Let's defer this to another PR.

OrKoN (approved):

thiagowfx:
Or rather: We could also react to the "authRequired" event. Anyway, some event needs to be emitted before we proceed further.

PR #1502, opened 2023-10-31
chore: fix EsLint and update rules

Lightning00Blade:
`import/no-cycle` currently a `no-op` as we don't have this issue but its good to keep track of as it may brake in different envs.
`import/order` update the be required and more precise, this rule has an auto-fixer so no manual action is required.
```

This setting allows import to properly work with typescript else not all feature were used.
settings: {
    'import/resolver': {
      typescript: true,
    },
  },
  ```


thiagowfx (approved):

thiagowfx:
PS. pre-commit is failing

PR #1518, opened 2023-11-02
feat(add intercept): address global intercepts

thiagowfx:
Whenever no patterns are specified in BiDi, omit the "patterns" parameter in "Fetch.enable", in order to match all requests:

> If specified, only requests matching any of these patterns will
produce fetchRequested event and will be paused until clients response. If not set, all requests will be affected.

Bug: #644
Docs: https://chromedevtools.github.io/devtools-protocol/tot/Fetch/#method-enable

### DIFF
--- a/src/bidiMapper/domains/network/NetworkStorage.spec.ts
+++ b/src/bidiMapper/domains/network/NetworkStorage.spec.ts
@@ -158,7 +158,7 @@ describe('NetworkStorage', () => {
 
     expect(networkStorage.getFetchEnableParams()).to.deep.equal({
       handleAuthRequests: false,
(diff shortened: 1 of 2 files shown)


thiagowfx:
This will fix some WPT tests, however I'll need to first make a change in the tests because some of its set-up code is introducing side-effects, blocking requests too early.

Lightning00Blade (approved):

thiagowfx:
This approach does not work.

thiagowfx:
Closing in favour of https://github.com/GoogleChromeLabs/chromium-bidi/pull/1530

PR #1521, opened 2023-11-02
test: add `bad_ssl_url` with local server

thiagowfx:
Certificates were generated with this command:

    faketime 'last friday 5 pm' /bin/bash -c 'openssl req -x509 -newkey rsa:4096 -keyout key.pem -out cert.pem -days 6 -nodes'

Verification can be done with this command:

    openssl x509 -noout -text -in cert.pem

Bug: #575


thiagowfx:
Certificates were generated this way:

```
faketime 'last friday 5 pm' /bin/bash -c 'openssl req -x509 -newkey rsa:4096 -keyout key.pem -out cert.pem -days 6 -nodes'
openssl x509 -noout -text -in cert.pem
```

tests/conftest.py
```diff
@@ -168,18 +171,44 @@ def hang_url(local_server: LocalHttpServer):
...
+    cert_file_name = Path(__file__).parent / "tools" / "cert.pem"
+    key_file_name = Path(__file__).parent / "tools" / "key.pem"
+
+    context.load_cert_chain(cert_file_name, key_file_name)
+
+    return context
+
 
+@pytest.fixture(scope="session")
+def ssl_httpserver_err_cert_authority_invalid(
```
Lightning00Blade, line 188: How about we call it `https_server_invalid_cert` ?

tests/conftest.py
```diff
@@ -168,18 +171,44 @@ def hang_url(local_server: LocalHttpServer):
...
+        ssl_context_err_cert_authority_invalid):
+    """Return a HTTPServer instance with SSL enabled.
```
Lightning00Blade, line 190: We should probably change the name of the `HTTPServer` to `HttpServer` to be more inline with the code conventions we have.
```suggestion
    """Return a HttpsServer instance with invalid SSL enabled.
```

Lightning00Blade: I think this is non actionable as its lib import.

Lightning00Blade (approved):

tests/conftest.py
thiagowfx: I used a slightly different name taking your suggestion into account

PR #1528, opened 2023-11-07
fix: automatically continue ignored events

Lightning00Blade:
Currently we ignore `favicon.ico` request. We should not block them for interception.

### DIFF
--- a/src/bidiMapper/domains/network/NetworkRequest.ts
+++ b/src/bidiMapper/domains/network/NetworkRequest.ts
@@ -230,7 +230,14 @@ export class NetworkRequest {
   onRequestPaused(
     params: Protocol.Fetch.RequestPausedEvent,
     networkStorage: NetworkStorage
-  ) {
+  ): void {
+    if (this.#isIgnoredEvent()) {
+      void this.continueRequest(params.requestId).catch(() => {
+        // TODO: Add some logging
+      });
+      return;
+    }
+
     // The stage of the request can be determined by presence of
     // responseErrorReason and responseStatusCode -- the request is at
     // the response stage if either of these fields is present and in the


src/bidiMapper/domains/network/NetworkRequest.ts
```diff
@@ -230,7 +230,14 @@ export class NetworkRequest {
   onRequestPaused(
     params: Protocol.Fetch.RequestPausedEvent,
     networkStorage: NetworkStorage
-  ) {
+  ): void {
+    if (this.#isIgnoredEvent()) {
```
thiagowfx, line 234: It's `continueRequest`, so maybe we should do this only for `phase == Network.InterceptPhase.BeforeRequestSent`?

src/bidiMapper/domains/network/NetworkRequest.ts
```diff
@@ -230,7 +230,14 @@ export class NetworkRequest {
...
+      void this.continueRequest(params.requestId).catch(() => {
```
thiagowfx, line 235: No `await`?

Lightning00Blade: Here no difference will be observed so explicitly opting in for `void` to make the code easier to understand.

src/bidiMapper/domains/network/NetworkRequest.ts
Lightning00Blade: From CDP docs `Fetch.requestPaused` - `Issued when the domain is enabled and the request URL matches the specified filter. The request is paused until the client responds with one of continueRequest, failRequest or fulfillRequest. ` so I don't know when `continueRequest` is needed.

thiagowfx: OK, let's try. This is moving towards the right direction either way.

sadym-chromium (approved):

OrKoN:
Q: why do we ignore favicon?

Lightning00Blade:
@OrKoN If I recall correctly WPT test don't expect whose request/response and will fail assertions, and due to them firing at different time test will become very flaky.

PR #1539, opened 2023-11-13
feat: abort navigation if network request failed

sadym-chromium:
If browsing context is closed, cancel all ongoing navigations.

Required for [Puppeteer tests](https://github.com/puppeteer/puppeteer/blob/9f161ec76b19c419bb1dd36357c09c044d5f82f4/test/src/launcher.spec.ts#L40).

### DIFF
--- a/src/bidiMapper/domains/context/BrowsingContextImpl.ts
+++ b/src/bidiMapper/domains/context/BrowsingContextImpl.ts
@@ -69,6 +69,8 @@ export class BrowsingContextImpl {
   };
 
   #url = 'about:blank';
+  // Required to provide navigation info in case of navigation was canceled.
+  #ongoingNavigationUrl?: string;
   readonly #eventManager: EventManager;
   readonly #realmStorage: RealmStorage;
   #loaderId?: Protocol.Network.LoaderId;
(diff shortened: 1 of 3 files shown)


src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -69,6 +69,8 @@ export class BrowsingContextImpl {
   };
 
   #url = 'about:blank';
+  // Required to provide navigation info in case of navigation was canceled.
+  #ongoingNavigationUrl?: string;
```
thiagowfx, line 73: This URL doesn't seem to be used for anything.

AI:

- either make this a boolean, OR
- print it alongside "navigation canceled", for more useful logging

thiagowfx (changes requested):
See comment. LGTM for the test.

src/bidiMapper/domains/context/BrowsingContextImpl.ts
sadym-chromium: You are right. At first I was going to implement `browsingContext.navigationAborted` event. But then decided to leave it aside, as just failing navigation would be enough for Puppeteer.

thiagowfx (approved):

PR #1543, opened 2023-11-13
chore: reduce WS handshake latency

sadym-chromium:
Attempt to address WPT flakiness:
* **Speed up WS connection in pure Mapper mode.** In pure Mapper mode, don't wait for the browser to start, and accept WS connection immediately. Close connection if browser failed to be launched.
* **Disable tests with timeout expected.**

### DIFF
--- a/wpt-metadata/chromedriver/headless/webdriver/tests/bidi/browsing_context/set_viewport/set_viewport.py.ini
+++ b/wpt-metadata/chromedriver/headless/webdriver/tests/bidi/browsing_context/set_viewport/set_viewport.py.ini
@@ -1,3 +1,0 @@
-[set_viewport.py]
-  [test_set_viewport_reset]
-    disabled: flaky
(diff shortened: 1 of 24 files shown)


wpt-metadata/chromedriver/headless/webdriver/tests/bidi/browsing_context/set_viewport/set_viewport.py.ini
```diff
@@ -0,0 +1,3 @@
+[set_viewport.py]
+  [test_set_viewport_reset]
+    disabled: flaky
```
thiagowfx, line 3: To mark as flaky: `expected: [PASS, ERROR]` is more explicit. Or `FAIL` instead of `ERROR` depending on the circumstance.

sadym-chromium: Do we really want to run test if we don't care about it's result? I find disabling with a mark "flaky" more informative, as it does not give a false confidence after the test passed. Plus, disabling flaky tests reduces the test runtime. 

Of cause I'm open for discussion.

thiagowfx: I'm open to either approach.

thiagowfx (approved):

PR #1544, opened 2023-11-14
ci(wpt): add --run-by-dir

thiagowfx:
Docs: https://web-platform-tests.org/running-tests/command-line-arguments.html


thiagowfx:
@sadym-chromium This run has passed on the first try. Maybe just a coincidence? Should we submit it as an experiment?

thiagowfx:
I'll rebase to trigger another run.

sadym-chromium (approved):
I'd assume the problem can be in running too many tests in parallel. But feel free to merge if the second WPT run passes as well

PR #1546, opened 2023-11-14
ci(e2e): re-introduce pytest-rerunfailures to mitigate CI flakiness

thiagowfx:
Docs: https://pypi.org/project/pytest-rerunfailures/


Lightning00Blade (commented):
Question: Can we do this only for CI? Else it would be annoying running test locally that fail and having to run them 2 times.
Also what logs get produced, we will still get the error for the flaky run or only if both fail we get message?

thiagowfx:
Logs: You should see both.

CI only: That's a good idea. I'm not sure if it's possible. Let me see.

thiagowfx:
Done.

Lightning00Blade (approved):

PR #1553, opened 2023-11-14
feat!: `acceptInsecureCerts`

sadym-chromium:
BREAKING CHANGE: `BidiServer.createAndStart` signature changed. New optional parameter `options` is added. Breaking change for Puppeteer, while ChromeDriver is not affected, as it uses the Mapper Tab.

When creating Mapper, add possibility to add `acceptInsecureCerts` parameter. Required to provide capability `acceptInsecureCerts`.

Alternative would be handling `session.new` command on the Mapper side, and forward capabilities there. But that would add an extra step to the ChromeDriver - Mapper link.

In order to respect capabilities from both Classic and Bidi, the Mapper launcher has to be refactored, so that it keeps sessions.

### DIFF
--- a/src/bidiServer/WebSocketServer.ts
+++ b/src/bidiServer/WebSocketServer.ts
@@ -221,7 +221,7 @@ export class WebSocketServer {
       }
 
       connection.on('message', async (message) => {
-        // If |type| is not text, return error.
+        // If type is not text, return error.
         if (message.type !== 'utf8') {
           this.#respondWithError(
             connection,
--- a/src/bidiServer/WebSocketServer.ts
+++ b/src/bidiServer/WebSocketServer.ts
@@ -179,26 +179,26 @@ export class WebSocketServer {
       // Session is set either by Classic or BiDi commands.
       let session: Session | undefined;
 
-      const sessionId = (request.resource ?? '').split('/').pop();
+      const requestSessionId = (request.resource ?? '').split('/').pop();
       debugInternal(
         `new WS request received. Path: ${JSON.stringify(
           request.resourceURL.path
-        )}, sessionId: ${JSON.stringify(sessionId)}`
+        )}, sessionId: ${JSON.stringify(requestSessionId)}`
       );
 
       if (
-        sessionId !== '' &&
-        sessionId !== undefined &&
-        !this.#sessions.has(sessionId)
+        requestSessionId !== '' &&
+        requestSessionId !== undefined &&
+        !this.#sessions.has(requestSessionId)
       ) {
-        debugInternal('Unknown session id:', sessionId);
+        debugInternal('Unknown session id:', requestSessionId);
         request.reject();
         return;
       }
 
       const connection = request.accept();
 
-      session = this.#sessions.get(sessionId ?? '');
+      session = this.#sessions.get(requestSessionId ?? '');
       if (session !== undefined) {
         // BrowserInstance is created for each new WS connection, even for the
         // same SessionId. This is because WPT uses a single session for all the
@@ -313,7 +313,7 @@ export class WebSocketServer {
               id: parsedCommandData.id,
               type: 'success',
               result: {
-                sessionId: '1',
+                sessionId: session.sessionId,
                 capabilities: {},
               },
             },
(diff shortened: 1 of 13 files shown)


thiagowfx (approved):

thiagowfx:
LGTM but what is this intended for?

sadym-chromium:
> LGTM but what is this intended for?

Oh, sorry, didn't mean to add you as a reviewer. It's still a draft.

sadym-chromium:
> LGTM but what is this intended for?

extended explanation

tests/conftest.py
```diff
@@ -38,14 +38,26 @@ def local_server(httpserver) -> LocalHttpServer:
...
 
 
+@pytest_asyncio.fixture
+async def websocket(websocket_connection):
+    """Return a websocket with an active BiDi session."""
+    await execute_command(websocket_connection, {
+        "method": "session.new",
+        "params": {}
+    })
+    yield websocket_connection
```
thiagowfx, line 58: return

`yield` is only needed if you have more statements afterwards

tests/browsing_context/test_navigate.py
```diff
@@ -418,3 +419,59 @@ async def test_browsingContext_navigationStarted_browsingContextClosedBeforeNavi
         'id': close_command_id,
         'type': 'success',
     })
+
+
+@pytest.mark.asyncio
+async def test_browsingContext_navigateBadSsl_notNavigated(
+        websocket, context_id):
+    # TODO: make offline.
```
thiagowfx, line 427: FYI we already have an off-line fixture

tests/conftest.py
sadym-chromium: fixed

tests/browsing_context/test_navigate.py
sadym-chromium: ack, thanks!

thiagowfx: It's still using the online URL

thiagowfx: bad_ssl_url

thiagowfx (changes requested):
Will re-review in a bit

tests/browsing_context/test_navigate.py
sadym-chromium: my bad ~ssl~. Done now

tests/conftest.py
```diff
@@ -38,14 +38,32 @@ def local_server(httpserver) -> LocalHttpServer:
...
     port = os.getenv("PORT", 8080)
     url = f"ws://localhost:{port}"
     async with websockets.connect(url) as connection:
         yield connection
 
 
+@pytest_asyncio.fixture(params=[{"capabilities": {}}])
+async def websocket(request, _websocket_connection):
+    capabilities = request.param['capabilities']
+    """Return a websocket with an active BiDi session."""
```
thiagowfx, line 54: nit: docstring should come first, right after the function definition

src/bidiServer/WebSocketServer.ts
```diff
@@ -30,7 +31,32 @@ const debugInternal = debug('bidi:server:internal');
 const debugSend = debug('bidi:server:SEND ▸');
 const debugRecv = debug('bidi:server:RECV ◂');
 
+type Session = {
+  sessionId: string;
+  // Promised is used to decrease WebSocket handshake latency. If session is set via
```
thiagowfx, line 36:
```suggestion
  // Promise is used to decrease WebSocket handshake latency. If session is set via
```

src/bidiServer/WebSocketServer.ts
```diff
@@ -126,44 +176,52 @@ export class WebSocketServer {
...
-        }
-      })();
+      }
 
       connection.on('message', async (message) => {
-        // Wait for browser instance to be created.
-        const browserInstance = await browserInstanceDeferred;
-
-        // If |type| is not text, return a error.
+        // If |type| is not text, return error.
```
thiagowfx, line 224:
```suggestion
        // If type is not text, return error.
```

`||` is spec formatting, not relevant here

tests/conftest.py
sadym-chromium: done

iKn0wH0w (approved):

PR #1565, opened 2023-11-17
feat: implement dedicated workers

jrandolf:


src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -447,32 +454,45 @@ export class BrowsingContextImpl {
...
-            ? params.context.name
-            : undefined,
+          sandbox,
           this.#cdpTarget.cdpClient,
           this.#eventManager,
           this.#logger
         );
 
-        if (params.context.auxData.isDefault) {
+        if (auxData && auxData.isDefault) {
```
OrKoN, line 495:
```suggestion
        if (auxData?.isDefault) {
```

src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -447,32 +454,45 @@ export class BrowsingContextImpl {
...
+            case 'default':
+              break;
+            default:
+              return;
+          }
         }
-        // Only this execution contexts are supported for now.
-        if (!['default', 'isolated'].includes(params.context.auxData.type)) {
-          return;
+        if (['://', ''].includes(origin)) {
```
OrKoN, line 476: let's extract it in a named helper or keep the comment. Current version does not have an indication why it is so,

src/bidiMapper/domains/context/BrowsingContextProcessor.ts
```diff
@@ -332,59 +332,99 @@ export class BrowsingContextProcessor {
...
-        this.#logger
-      );
+        const maybeContext = this.#browsingContextStorage.findContext(
+          targetInfo.targetId
+        );
+        if (maybeContext) {
+          // OOPiF.
+          maybeContext.updateCdpTarget(cdpTarget);
+        } else {
+          // New context.
```
OrKoN, line 404: I don't think this is applicable for workers

OrKoN: I meant OOPIF handling

src/bidiMapper/domains/context/BrowsingContextProcessor.ts
```diff
@@ -332,59 +332,99 @@ export class BrowsingContextProcessor {
...
+          BrowsingContextImpl.create(
```
OrKoN, line 405: this actually looks wrong too since the worker should not have a browsing context

tests/script/test_realm.py
```diff
@@ -149,3 +201,50 @@ async def test_realm_realmDestroyed_sandbox(websocket, context_id):
...
+        websocket, {
+            "method": "browsingContext.navigate",
+            "params": {
+                "context": context_id,
+                "url": url,
+                "wait": "complete",
+            }
+        })
+
+    assert {
```
OrKoN, line 223: why is the realm destroyed event expected here?

OrKoN: is it the previous realm in the page? can we in some way assert that the right realm is destroyed?

src/bidiMapper/domains/context/BrowsingContextProcessor.ts
```diff
@@ -332,88 +333,133 @@ export class BrowsingContextProcessor {
...
-    // XXX: params.targetId is deprecated. Update this class to track using
-    // params.sessionId instead.
-    // https://github.com/GoogleChromeLabs/chromium-bidi/issues/60
-    const contextId = params.targetId!;
-    this.#browsingContextStorage.findContext(contextId)?.dispose();
-
-    this.#preloadScriptStorage
-      .find({targetId: contextId})
-      .map((preloadScript) => preloadScript.dispose(contextId));
+    const context = this.#browsingContextStorage
```
OrKoN, line 437: dispose is not called anymore on the context?

src/bidiMapper/domains/context/BrowsingContextProcessor.ts
```diff
@@ -332,88 +333,133 @@ export class BrowsingContextProcessor {
...
-        targetInfo.targetId,
-        null,
-        this.#eventManager,
         this.#browsingContextStorage,
+        uniqueId,
+        cdpTarget.targetId,
+        id,
+        serializeOrigin(origin),
+        'dedicated-worker',
+        undefined,
```
OrKoN, line 425: leave a todo that sandboxes for workers have to be handled here?

tests/script/test_realm.py
jrandolf: It's the previous realm. It's guaranteed that this will always be the realm of the previous page.

src/bidiMapper/domains/context/BrowsingContextProcessor.ts
jrandolf: It is called? Next line.

src/bidiMapper/domains/context/BrowsingContextProcessor.ts
jrandolf: We don't support sandboxes for workers.

OrKoN (approved):
LGTM thanks

PR #1569, opened 2023-11-17
test: re-baseline WPT expectations

sadym-chromium:
Step 1: remove all expectations and run CI.
Step 2: compare generated expectations with the actual one.


thiagowfx (approved):

thiagowfx:
This is indeed useful to do sometimes, because the `wpt` CLI does not have the ability to remove old (e.g. renamed) expectations.

thiagowfx:
Ideally we could have a monthly (or even weekly?) job that does what you're doing now. What do you think?

sadym-chromium:
> Ideally we could have a monthly (or even weekly?) job that does what you're doing now. What do you think?

Agree! It would be nice to automate and run monthly

sadym-chromium:
The PR can bring some flakiness. Postponing to the next week.

thiagowfx (approved):

thiagowfx:
Merge?

sadym-chromium:
Run CI again to verify it's stability

PR #1588, opened 2023-11-23
test: split WPT CI runs into chunks

sadym-chromium:
In order to speed up WPT CI runtime, split tests into chunks to run in parallel.


OrKoN:
I can suggest to create required checks as follows https://github.com/puppeteer/puppeteer/blob/main/.github/workflows/ci.yml#L213 so that not every individual matrix job needs to be listed

.github/workflows/wpt.yml
```diff
@@ -101,13 +101,15 @@ jobs:
           commit_options: -n --signoff
           file_pattern: 'wpt-metadata/**/*.ini'
   wpt:
-    name: ${{ matrix.kind }}-${{ matrix.head }}
+    name: wpt ${{ matrix.this_chunk }}/${{ matrix.total_chunks }} ${{ matrix.kind }}-${{ matrix.head }}
```
OrKoN, line 104:
```suggestion
    name: wpt ${{ matrix.kind }}-${{ matrix.head }} ${{ matrix.this_chunk }}/${{ matrix.total_chunks }}
```

sadym-chromium:
Splitting WPT tests in chunks is possible, but the logic of updating expectations and generating HTML report becoming way more complicated.

As an intermediate solution, we can run both chunked tests and use them for verification and full run for updating WPT expectations and generating HTML report.

I'm postponing this task for some time.

PR #1590, opened 2023-11-23
test: shard WPT tests

sadym-chromium:


sadym-chromium:
done in https://github.com/GoogleChromeLabs/chromium-bidi/pull/1682

PR #1593, opened 2023-11-24
feat: implement `storage.getCookies` and `storage.setCookie`

sadym-chromium:
Implement storage commands:
* `storage.getCookies`
* `storage.setCookie`

### DIFF
--- a/src/bidiMapper/domains/storage/StorageProcessor.ts
+++ b/src/bidiMapper/domains/storage/StorageProcessor.ts
@@ -19,8 +19,11 @@ import type {Protocol} from 'devtools-protocol';
 import type {CdpClient} from '../../../cdp/CdpClient.js';
 import type {Storage} from '../../../protocol/protocol.js';
 import {
+  InvalidArgumentException,
   Network,
   UnderspecifiedStoragePartitionException,
+  UnableToSetCookieException,
+  UnsupportedOperationException,
 } from '../../../protocol/protocol.js';
 import type {LoggerFn} from '../../../utils/log.js';
 import {LogType} from '../../../utils/log.js';
@@ -88,6 +91,55 @@ export class StorageProcessor {
     };
   }
 
+  async setCookie(
+    params: Storage.SetCookieParameters
+  ): Promise<Storage.SetCookieResult> {
+    const partitionKey = this.#extendStoragePartitionSpec(params);
+
+    if (params.cookie.value.type !== 'string') {
+      // TODO: add support for other types.
+      throw new UnsupportedOperationException(
+        'Only string cookie values are supported'
+      );
+    }
+    const deserializedValue = params.cookie.value.value;
+
+    try {
+      await this.#browserCdpClient.sendCommand('Storage.setCookies', {
+        cookies: [
+          {
+            name: params.cookie.name,
+            value: deserializedValue,
+            domain: params.cookie.domain,
+            path: params.cookie.path ?? '/',
+            secure: params.cookie.secure ?? false,
+            httpOnly: params.cookie.httpOnly ?? false,
+            // CDP's partition key is the source origin.
+            // partitionKey: partitionKey.get('sourceOrigin'),
+            ...(params.cookie.expiry !== undefined && {
+              expires: params.cookie.expiry,
+            }),
+            ...(params.cookie.sameSite !== undefined && {
+              sameSite: StorageProcessor.#sameSiteBiDiToCdp(
+                params.cookie.sameSite
+              ),
+            }),
+            // Note: the following fields supported by CDP are not set:
+            // url
+            // priority
+            // sameParty
+            // sourceScheme
+            // sourcePort
+          },
+        ],
+      });
+    } catch (e: any) {
+      this.#logger?.(LogType.debugError, e);
+      throw new UnableToSetCookieException(e.toString());
+    }
+    return {partitionKey: Object.fromEntries(partitionKey)};
+  }
+
   #extendStoragePartitionSpec(
     params: Storage.GetCookiesParameters
   ): Map<string, string> {
@@ -121,6 +173,22 @@ export class StorageProcessor {
         }
       }
     }
+
+    if (!partitionKey.has('sourceOrigin')) {
+      throw new UnderspecifiedStoragePartitionException(
+        'sourceOrigin or cookie.domain should be set'
+      );
+    }
+    const unsupportedPartitionKeys = Array.from(partitionKey.keys()).filter(
+      (k) => k !== 'sourceOrigin'
+    );
+    if (unsupportedPartitionKeys.length > 0) {
+      this.#logger?.(
+        LogType.debugInfo,
+        `Unsupported partition keys: ${unsupportedPartitionKeys}`
+      );
+    }
+
     return partitionKey;
   }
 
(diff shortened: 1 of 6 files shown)


src/protocol-parser/protocol-parser.ts
```diff
@@ -290,3 +290,14 @@ export namespace Cdp {
     return parseObject(params, GetSessionRequestSchema);
   }
 }
+
+export namespace Proto {
```
OrKoN, line 294: What does Proto mean? Let's just roll the types updates using the cddl generator based on the latest PR? it will also validate if the cddl definitions are correct. WDYT?

sadym-chromium: It's a bit non-trivial to do in CI, as the PR is from the fork. I can try to run update locally.

tests/storage/test_get_cookies.py
```diff
@@ -0,0 +1,101 @@
...
+            }),
+        ],
+        'partitionKey': {},
+    }
+
+
+@pytest.mark.asyncio
+async def test_get_cookies_filter(websocket, context_id, another_context_id,
+                                  example_url, another_example_url):
+    example_domain = example_url.split('://')[1].split('/')[0].split(':')[0]
```
OrKoN, line 78: is there a built-in URL parse in Python? https://docs.python.org/3/library/urllib.parse.html ?

tests/test_helpers.py
```diff
@@ -122,21 +122,32 @@ async def goto_url(
...
+                "awaitPromise": await_promise,
+                "resultOwnership": result_ownership
             }
         })
 
 
+async def set_html_content(websocket, context_id: str, html_content: str):
+    """Sets the current page content without navigation."""
+    await eval_expression(websocket, context_id,
+                          f"document.body.innerHTML = '{html_content}'")
```
OrKoN, line 148: for the future: this does not seem to properly escape so perhaps we need to switch to callFunction and pass html content as an arg

tests/storage/test_get_cookies.py
OrKoN: at least a helper for url to domain parsing would seem useful

.github/workflows/e2e.yml
```diff
@@ -11,7 +11,7 @@ env:
   FORCE_COLOR: 3
   PIP_DISABLE_PIP_VERSION_CHECK: 1
   # https://pypi.org/project/pytest-rerunfailures/
-  PYTEST_ADDOPTS: '--reruns 1'
+  # PYTEST_ADDOPTS: '--reruns 1'
```
OrKoN, line 14: what does this env var do?

src/protocol-parser/protocol-parser.ts
sadym-chromium: Oh, the [spec PR](https://github.com/w3c/webdriver-bidi/pull/501) is not CDDL-ready yet:
```
error: parser errors
    ┌─ input:854:3
    │
854 │   storage.DeleteCookiesResult
    │   ^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing definition for rule storage.DeleteCookiesResult
    ·
857 │ storage.PartitionKey {
    │ ^^^^^^^^^^^^^^^^^^^^ expected assignment token '=', '/=' or '//=' after rule identifier
    ·
865 │   partitionKey: storage.partitionKey,
    │                 ^^^^^^^^^^^^^^^^^^^^ missing definition for rule storage.partitionKey
    ·
869 │   partitionKey: storage.partitionKey
    │                 ^^^^^^^^^^^^^^^^^^^^ missing definition for rule storage.partitionKey

Error: "incremental parsing error"
```

.github/workflows/e2e.yml
sadym-chromium: It reruns failing tests. I removed it to unmask tst flakiness and to ease debugging tests

tests/storage/test_get_cookies.py
sadym-chromium: Good point. Done.

.github/workflows/e2e.yml
sadym-chromium: Added TODO to roll it back or to remove it

src/protocol-parser/protocol-parser.ts
sadym-chromium: https://github.com/GoogleChromeLabs/chromium-bidi/actions/runs/6982747779/job/19002554875

tests/test_helpers.py
sadym-chromium: ack

src/protocol-parser/protocol-parser.ts
sadym-chromium: `Proto` will me removed once proper types are generated from the spec.

OrKoN: Note that technically the spec only allows implementation extension modules with a colon character in the module name :)

OrKoN: Btw I don't think we can to land it before it is spec'ed so I don't see a point in having proto prefixes.

sadym-chromium: > Btw I don't think we can (or rather need to) to land it before it is spec'ed so I don't see a point in having proto prefixes.

I understand the reasons for both delaying and approving this PR until the related specification changes are finalized. The decision to delay is because it's implemented in the unspecified `proto` domain, and the specification and specific API might change. However, merging this PR would enable a proof of concept and prototype integration with Puppeteer. In my opinion, it's still a good idea to merge this PR. 

WDYT?

OrKoN: > However, merging this PR would enable a proof of concept and prototype integration with Puppeteer. In my opinion, it's still a good idea to merge this PR.

We can do a proof of concept in Puppeteer based on this branch (and in a branch as well). I think we don't need to publish the unspecified change for that.

tests/test_helpers.py
sadym-chromium: fixed

src/bidiMapper/domains/storage/StorageProcessor.ts
```diff
@@ -0,0 +1,283 @@
...
+      throw new UnsupportedOperationException(
+        'Only string cookie values are supported'
+      );
+    }
+    const deserializedValue = params.cookie.value.value;
+
+    try {
+      await this.#browserCdpClient.sendCommand('Storage.setCookies', {
+        cookies: [
+          {
```
jrandolf, line 94: Maybe create a helper to convert bidiCookiesToCdpCookies?

jrandolf:
lgtm!

jrandolf (approved):

src/bidiMapper/domains/storage/StorageProcessor.ts
sadym-chromium: Good idea! Done

src/bidiMapper/domains/storage/StorageProcessor.ts
```diff
@@ -0,0 +1,286 @@
...
+    };
+  }
+
+  #expandStoragePartitionSpecByBrowsingContext(
+    descriptor: Storage.BrowsingContextPartitionDescriptor
+  ): Storage.PartitionKey {
+    const browsingContextId: string = descriptor.context;
+    const browsingContext =
+      this.#browsingContextStorage.getContext(browsingContextId);
+    const url = NetworkProcessor.parseUrlString(browsingContext?.url ?? '');
```
Lightning00Blade, line 103: Nit: This should not live on the `NetworkProcessor` but extracted into helper.

PR #1594, opened 2023-11-24
ci: extand `Update WebdriverBidi types` with a custom reference

sadym-chromium:
Sometimes it is required to have types from not-yet merged branch of spec (https://github.com/GoogleChromeLabs/chromium-bidi/pull/1593). This PR provides a way to specify a branch of a specific commit to update types from.

### DIFF
--- a/.github/workflows/update-bidi-types.yml
+++ b/.github/workflows/update-bidi-types.yml
@@ -10,6 +10,12 @@ on:
     # Run daily at 10AM.
     - cron: '0 10 * * *'
   workflow_dispatch:
+    inputs:
+      source_ref:
+        description: Ref in spec repo to update from
+        default: main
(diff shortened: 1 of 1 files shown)


.github/workflows/update-bidi-types.yml
```diff
@@ -20,6 +26,7 @@ jobs:
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v3.5.3
         with:
           repository: w3c/webdriver-bidi
```
OrKoN, line 28: we can also add a parameter for the repo to support forks

sadym-chromium: good point. Done

OrKoN (approved):

PR #1531, opened 2023-11-08
test: Selenium CI

sadym-chromium:
Run Selenium with the current Mapper version.

The following functionality is covered:
* Create BrowsingContext
* Navigation
* Evaluate script
* Capture screenshot

Open questions:
1. Other scenarios?
2. Other platforms?
3. Headless?

PR #1615, opened 2023-12-12
test: WebdriverIO CI

christian-bromann:
Similar to https://github.com/GoogleChromeLabs/chromium-bidi/pull/1531 this patch adds WebdriverIO integration tests.


.github/workflows/webdriverio.yml
```diff
@@ -0,0 +1,54 @@
...
+  pull_request:
+  push:
+    branches: 'main'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  selenium:
```
sadym-chromium, line 19:
```suggestion
  webdriver-io:
```

.github/workflows/webdriverio.yml
```diff
@@ -0,0 +1,54 @@
...
+        id: browser
+        run: node tools/install-browser.mjs --github
+      - name: Install chromedriver for the pinned browser
+        id: driver
+        run: node tools/install-browser.mjs --chromedriver --github
+      - name: Run WebdriverIO test
+        timeout-minutes: 20
+        run: >
+          xvfb-run --auto-servernum
+          node tests_external/selenium.mjs
```
sadym-chromium, line 51:
```suggestion
          node tests_external/webdriverio.mjs
```

tests_external/helpers/index.mjs
```diff
@@ -0,0 +1,26 @@
+import {execSync} from 'child_process';
```
sadym-chromium, line 1: It perfectly make sense to move this logic to a common place, but I think it's better to move it to `tools`: https://github.com/GoogleChromeLabs/chromium-bidi/pull/1616

sadym-chromium (commented):
Nice job, thanks! Waiting for the https://github.com/GoogleChromeLabs/chromium-bidi/pull/1616 to be merged.

package.json
```diff
@@ -229,6 +229,7 @@
     "terser": "5.25.0",
     "tslib": "2.6.2",
     "typescript": "5.3.2",
+    "webdriverio": "^8.26.1",
```
Lightning00Blade, line 232: Please pin the version. Dependabot will handle updating it.

sadym-chromium, line 232:
```suggestion
    "webdriverio": "8.26.1",
```

Lightning00Blade: We can't use suggestion to update this as it does not update the `package-lock.json`.

sadym-chromium (approved):

OrKoN:
FYI we plan to remove this test in https://github.com/GoogleChromeLabs/chromium-bidi/pull/4156 to reduce the number of dependencies. I trust that you have sufficient test coverage downstream and I do not expect this test to catch things that WPT do not catch.

PR #1570, opened 2023-11-17
Revert "feat: implement dedicated workers"

sadym-chromium:
Reverts GoogleChromeLabs/chromium-bidi#1565

e2e tests started constantly timing out.

PR #1619, opened 2023-12-13
Revert "Revert "feat: implement dedicated workers""

jrandolf:
Reverts GoogleChromeLabs/chromium-bidi#1570


sadym-chromium:
@jrandolf could you please extend the description?

src/bidiMapper/domains/context/BrowsingContextProcessor.ts
```diff
@@ -335,89 +336,136 @@ export class BrowsingContextProcessor {
...
+    // debugger and ignore them.
+    targetCdpClient
+      .sendCommand('Runtime.runIfWaitingForDebugger')
+      .then(() =>
+        parentSessionCdpClient.sendCommand('Target.detachFromTarget', params)
+      )
+      .catch((error) => this.#logger?.(LogType.debugError, error));
+  }
+
+  #workers = new Map<string, Realm>();
```
sadym-chromium, line 418: Move to the beginning of the class.

src/bidiMapper/domains/context/BrowsingContextProcessor.ts
```diff
@@ -335,89 +336,136 @@ export class BrowsingContextProcessor {
...
+  #createWorkerRealm(cdpTarget: CdpTarget) {
```
sadym-chromium, line 419: This method does not create a worker realm, but rather adds listeners to create it after the corresponding event is received. Should it be called like `#createWorkerTarget` or `#processWorkerTarget` then?

src/bidiMapper/domains/context/BrowsingContextProcessor.ts
```diff
@@ -335,89 +336,136 @@ export class BrowsingContextProcessor {
...
   }
 
   #handleTargetInfoChangedEvent(
     params: Protocol.Target.TargetInfoChangedEvent
   ) {
-    const contextId = params.targetInfo.targetId;
-    this.#browsingContextStorage
-      .findContext(contextId)
-      ?.onTargetInfoChanged(params);
-  }
```
sadym-chromium, line 415: nit: any particular reason for changing the method?

tests/script/test_realm.py
```diff
@@ -82,6 +82,39 @@ async def test_realm_realmCreated_sandbox(websocket, context_id):
...
+            "method": "browsingContext.navigate",
+            "params": {
+                "context": context_id,
+                "url": url,
+                "wait": "complete",
+            }
+        })
+
+    # Wait for worker to be created
+    while True:
```
sadym-chromium, line 103: You can call twice the `wait_for_event("script.realmCreated")` and verify the second event is from realm.

tests/script/test_realm.py
```diff
@@ -149,3 +182,62 @@ async def test_realm_realmDestroyed_sandbox(websocket, context_id):
...
+                "target": {
+                    "context": context_id
+                },
+                "expression": "window.w.terminate()",
+                "awaitPromise": True
+            }
+        })
+
+    # Wait for confirmation that worker was destroyed
+    while True:
```
sadym-chromium, line 234: `wait_for_event`

src/bidiMapper/domains/context/BrowsingContextProcessor.ts
jrandolf: Clarity

src/bidiMapper/domains/context/BrowsingContextProcessor.ts
jrandolf: I'd rather have declarations close to their use. It simplifies lookup. I usually hoist them to the beginning of the class if they get used in more than one area.

src/bidiMapper/domains/context/BrowsingContextProcessor.ts
jrandolf: Creation is a subjective term. I prefer not to create an additional function simply for separating functionality since this requires additional lookup when reading the function.

In particular, splitting a function should be determined by two things:

 1. Number of times it is used is greater than 1.
 2. There is more than one functionality, and
 3. The function is long.

At least two of the factors above should be satisfied to justify splitting a function. In this case, 1 and 3 are not satisfied so I'd prefer not splitting it.

jrandolf: Woops, read this wrong. Done.

tests/script/test_realm.py
sadym-chromium: I don't think the `while True` here is required, as you terminate only one realm, so there should be only one event. Just `await wait_for_event` and assert the event is expected.

sadym-chromium (approved):

PR #1625, opened 2023-12-14
feat: implement evaluate for dedicated workers

jrandolf:
Implements script evaluation for workers.

### DIFF
--- a/tests/script/test_evaluate.py
+++ b/tests/script/test_evaluate.py
@@ -16,7 +16,9 @@ from unittest.mock import ANY
 
 import pytest
 from anys import ANY_STR
-from test_helpers import execute_command, get_tree
+from syrupy.filters import props
+from test_helpers import (execute_command, get_tree, send_JSON_command,
+                          subscribe, wait_for_command, wait_for_event)
 
 
 @pytest.mark.asyncio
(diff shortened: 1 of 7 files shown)


tests/script/test_evaluate.py
```diff
@@ -395,3 +397,66 @@ async def test_scriptEvaluate_realm(websocket, context_id):
...
+                "target": {
+                    "realm": realm
+                },
+                "expression": "self.postMessage('hello world')",
+                "awaitPromise": True
+            }
+        })
+
+    # Wait for the response.
+    while True:
```
Lightning00Blade, line 458: Can you extract this and the `execute_command` logic into separate helper and reuse it?

tests/script/test_evaluate.py
```diff
@@ -395,3 +397,66 @@ async def test_scriptEvaluate_realm(websocket, context_id):
...
+                "target": {
+                    "context": context_id
+                },
+                "expression": "new Promise(resolve => window.w.addEventListener('message', ({data}) => resolve(data), {once: true}))",
+                "awaitPromise": True
+            }
+        })
+
+    # Post a message from the worker.
+    await execute_command(
```
Lightning00Blade, line 445: This should be `send_JSON_command`, as I am not sure if the order of completion will always be correct -> You may get the resolve before the command is completed.

Lightning00Blade (approved):

PR #1626, opened 2023-12-14
chore: prepare for changing `sharedId` format.

sadym-chromium:
Mapper accepts both new and old `sharedId` format. It generates old or new format based on option `sharedIdWithFrame` with `false` meaning old format by default.
Added new capability `sharedIdWithFrame` in NodeJS runner for testing purposes.

http://go/chromedriver:weak-map


nechaev-chromium (approved):
Thank you, Maxim!

jrandolf (approved):

PR #1636, opened 2023-12-18
chore: increase test_scriptEvaluate_dedicated_worker timeout

jrandolf:
Sometimes the web worker may take some time before it's destroyed, so we need to increase the timeout for this test.

### DIFF
--- a/tests/script/test_evaluate.py
+++ b/tests/script/test_evaluate.py
@@ -399,7 +399,7 @@ async def test_scriptEvaluate_realm(websocket, context_id):
     } == result
 
 
-@pytest.mark.timeout(30)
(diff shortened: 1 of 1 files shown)


tests/script/test_evaluate.py
```diff
@@ -399,6 +399,7 @@ async def test_scriptEvaluate_realm(websocket, context_id):
     } == result
 
 
+@pytest.mark.timeout(30)
```
Lightning00Blade, line 402: Let's not triple it if possible, let's double it and then see if we can do something if we need to increase it again.
```suggestion
@pytest.mark.timeout(20)
```

Lightning00Blade (approved):

PR #1645, opened 2023-12-27
feat: implement permissions

OrKoN:
Spec: https://www.w3.org/TR/permissions/#automation-webdriver-bidi
Tests: https://github.com/web-platform-tests/wpt/pull/43802


src/protocol/chromium-bidi.ts
```diff
@@ -87,7 +88,13 @@ export namespace Network {
   }
 }
 
-export type Command = (WebDriverBidi.Command | Cdp.Command) & {
+export type Command = (
+  | WebDriverBidi.Command
+  | Cdp.Command
+  | ({
+      id: WebDriverBidi.JsUint;
```
sadym-chromium, line 95: Is `id` added because the permissions is an extension, and it relies on the BiDi spec to have the `id`, so it is not generated in types?

sadym-chromium:
Please add some e2e tests

src/protocol/chromium-bidi.ts
OrKoN: Yes, the ID is part of the main spec and it is not defined explicitly in the permissions spec CDDL.

OrKoN:
Do you know any Chrome-specific permissions tests we could add?

src/protocol/chromium-bidi.ts
OrKoN: Added

OrKoN:
Added a happy path e2e test.

tests/permissions/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+    """ Queries a permission via the script.callFunction command."""
+
+    result = await execute_command(
+        websocket, {
+            "method": "script.callFunction",
+            "params": {
+                "functionDeclaration": """() => {
+                    return navigator.permissions.query({ name: '%s' })
+                      .then(val => val.state, err => err.message)
+                  }""" % name,
```
sadym-chromium, line 37:
```suggestion
                "functionDeclaration": f"""() => {
                    return navigator.permissions.query({ name: '{name}' })
                      .then(val => val.state, err => err.message)
                  }""",
```

sadym-chromium (approved):

tests/permissions/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+
+async def query_permission(websocket, context_id, name):
+    """ Queries a permission via the script.callFunction command."""
+
+    result = await execute_command(
+        websocket, {
+            "method": "script.callFunction",
+            "params": {
+                "functionDeclaration": f"""() => {
+                    return navigator.permissions.query({ name: '{name}' })
```
sadym-chromium, line 35:
```suggestion
                    return navigator.permissions.query({{ name: '{name}' }})
```

tests/permissions/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+    """ Set a permission via the permissions.setPermission command."""
+    return await execute_command(
+        websocket, {
+            'method': 'permissions.setPermission',
+            'params': {
+                'origin': origin,
+                'descriptor': descriptor,
+                'state': state
+            }
+        })
```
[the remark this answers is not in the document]
OrKoN: Inlining it in every call would reduce the readability, IMO? WPT test suite seems to have made the opposite trade-offs

tests/permissions/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+    """ Queries a permission via the script.callFunction command."""
+
+    result = await execute_command(
+        websocket, {
+            "method": "script.callFunction",
+            "params": {
+                "functionDeclaration": f"""() => {
+                    return navigator.permissions.query({{ name: '{name}' }})
+                      .then(val => val.state, err => err.message)
+                  }""",
```
sadym-chromium, line 37:
```suggestion
                "functionDeclaration": f"""() => {{
                    return navigator.permissions.query({{ name: '{name}' }})
                      .then(val => val.state, err => err.message)
                  }}""",
```

sadym-chromium: Ok, maybe string format was more readable option )))

PR #1659, opened 2024-01-04
build: use JS for cross compatibility

### LINKED ISSUE #876: Make tooling cross-platform (Linux, macOS, Windows)

Lightning00Blade:
Will leave comments on other places.
Closes #876


.eslintignore
```diff
@@ -9,6 +9,6 @@ lib
 logs/
 node_modules/
 wptreport*.json
-wpt/
+./wpt/
```
Lightning00Blade, line 12: This is more correct as `tools/wpt` which I was thinking of moving things was ignored because of it.

.prettierignore
```diff
@@ -1,11 +1,10 @@
-.eslintrc.js
```
Lightning00Blade, line 1: This was done due to the `keep-sorted` rule not being set-up properly back in the day. I fixed that a while ago.

tools/run-bidi-server.mjs
```diff
@@ -58,7 +58,7 @@ if (CHANNEL === 'local') {
 
 // DEBUG = (empty string) is allowed.
 const DEBUG = process.env.DEBUG ?? 'bidi:*';
-const DEBUG_COLORS = process.env.DEBUG_COLORS || 'yes';
+const DEBUG_COLORS = process.env.DEBUG_COLORS || 'false';
```
Lightning00Blade, line 61: Setting it to `false` by default as it makes the log generated files unreadable.

Lightning00Blade:
No they use the `npm run X` format.

tools/run-bidi-server.mjs
Lightning00Blade: image

OrKoN (approved):

PR #1660, opened 2024-01-04
chore: EsLint properly lints tooling

Lightning00Blade:
This PR separates the TypeScript rule and the JS rules, and ands `.mjs` to the list of files.
Also to remove some EsLint warning moved the generated files under a folder.


.eslintignore
```diff
@@ -1,14 +1,14 @@
...
 chromium-bidi.iml
+*.py.ini
 lib
 logs/
 node_modules/
 wptreport*.json
 ./wpt/
-/src/protocol/webdriver-bidi.ts
-/src/protocol-parser/webdriver-bidi.ts
+/src/protocol/cddl/
```
OrKoN, line 13: the folder does not contain cddl, let's rename to `generated`

.eslintrc.js
```diff
@@ -127,4 +78,55 @@ module.exports = {
...
+  overrides: [
+    {
+      files: ['*.ts'],
+      extends: [
+        'plugin:@typescript-eslint/eslint-recommended',
+        'plugin:@typescript-eslint/recommended',
+        'plugin:@typescript-eslint/recommended-requiring-type-checking',
+      ],
+      parserOptions: {
+        project: './tsconfig.base.json',
```
Lightning00Blade, line 90: This contains all the configuration that the parser needs, that way we can use TS (if needed) outside of the project and EsLint will not complain.

OrKoN (approved):

PR #1663, opened 2024-01-05
chore: remove deps for Args parsing

Lightning00Blade:


sadym-chromium:
Please make sure you update all the usages, eg in [tools/run-bidi-server.mjs](https://github.com/GoogleChromeLabs/chromium-bidi/blob/82d13df43b04e5e31118e21d1b6ba13bce9eb87c/tools/run-bidi-server.mjs#L81).

Lightning00Blade:
That should be update here - https://github.com/GoogleChromeLabs/chromium-bidi/pull/1663/files#diff-06b8a2692cf19ff55e4aa295e81ac271dcd639e3149221ce9f8e6b6f18f71411R87

sadym-chromium:
> That should be update here - https://github.com/GoogleChromeLabs/chromium-bidi/pull/1663/files#diff-06b8a2692cf19ff55e4aa295e81ac271dcd639e3149221ce9f8e6b6f18f71411R87

e2e headless CI job runs tests in headful mode now.

Lightning00Blade:
I think instead of using the build in Node, we should use only `yargs` as we use it in multiple places and allows for faster development.

PR #1664, opened 2024-01-05
refactor: move events under session

Lightning00Blade:
BiDi does not have a `event` module so we should keep event relevant logic under the Session module (domain - we should properly rename it module at this point.)


sadym-chromium (approved):
Makes sense.

PR #1665, opened 2024-01-05
refactor: remove I prefix form interfaces

Lightning00Blade:


Lightning00Blade:
We don't enforce it so its not present in all cases (consistency is the important part for me). Google style recommends against this naming convention - https://google.github.io/styleguide/tsguide.html#naming-style. Also the main TypeScript repo don't use it (thought like they say its for TypeScript convetion) - https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines

sadym-chromium (approved):

PR #1668, opened 2024-01-08
build(spec): update WebDriverBidi types

sadym-chromium:
Based on https://github.com/GoogleChromeLabs/chromium-bidi/pull/1638.

* Update BiDi types
* Add temp required field `owner` to worker realms.
* Fixed e2e tests and made them more informational by introducing a `wait_for_filtered_event` helper.


tests/script/test_evaluate.py
```diff
@@ -418,16 +418,22 @@ async def test_scriptEvaluate_dedicated_worker(websocket, context_id, html,
             }
         })
 
-    # Wait for worker to be created
-    while True:
```
sadym-chromium, line 422: If the event format is changed (eg added a new field), this test failed with timeout instead of providing a meaningful message.

sadym-chromium: I split the assertion into 2 parts:
* Get the event.
* Assert the event.

jrandolf (approved):

PR #1669, opened 2024-01-08
chore: update VS Code folder to use templates

Lightning00Blade:
We should not provide the files directly, as people may have different set-ups and custom workspace configuration.
If you need to maintain the current behavior  just copy the files without the `.template` part.


OrKoN (approved):

Lightning00Blade:
CC @sadym-chromium  @jrandolf, please create be advice is your IDE does not auto-format or similar it may be due to this change.

PR #1678, opened 2024-01-09
[WIP] WPT sharding

sadym-chromium:
Run only a small subset of WPT tests in CI. Preparation for WPT sharding.


sadym-chromium:
Done in https://github.com/GoogleChromeLabs/chromium-bidi/pull/1682

PR #1686, opened 2024-01-10
chore: speed up `eslint` and `prettier` when ran via `npm run format`

Lightning00Blade:
This should speed up running format as previously both prettier and eslint were looking at folders thay should not


package.json
```diff
@@ -9,12 +9,12 @@
     "e2e-headful": "wireit",
     "e2e-headless": "wireit",
     "e2e": "npm run e2e-headless --",
-    "eslint": "eslint --ext js --ext mjs --ext ts --fix .",
     "flake8": "flake8 examples/ tests/",
     "format": "npm run pre-commit --",
+    "format:eslint": "eslint --ext js --ext mjs --ext ts --fix .",
```
Lightning00Blade, line 14: Convetion for Package json - https://eslint.org/docs/latest/contribute/package-json-conventions
We should consider using it for the other commands as well.

sadym-chromium (approved):

PR #1703, opened 2024-01-15
chore: allow generating types for any CDDL file

OrKoN:


package.json
```diff
@@ -33,6 +33,7 @@
   },
   "wireit": {
     "bidi-types": {
+      "clean": "if-file-deleted",
```
Lightning00Blade, line 36: I think you need to specify `files` for this to have any affect, which we can't do due to cddl living in different places for each use. Also I think we want to delete the files when we run this command (default behaviour).

Lightning00Blade (approved):
LGTM + minor comment.

PR #1709, opened 2024-01-16
refactor: separate realm types

jrandolf:
This PR separates the realm class into distinct subclasses based on the spec leading to better separation of functionality.


src/bidiMapper/domains/log/LogManager.ts
```diff
@@ -90,6 +90,11 @@ export class LogManager {
           cdpSessionId: this.#cdpTarget.cdpSessionId,
           executionContextId: params.executionContextId,
         });
+        if (realm === undefined) {
+          // Ignore exceptions not attached to any realm.
+          return;
+        }
```
sadym-chromium, line 96: Maybe at least log them?

src/bidiMapper/domains/log/LogManager.ts
```diff
@@ -140,31 +148,39 @@ export class LogManager {
           cdpSessionId: this.#cdpTarget.cdpSessionId,
           executionContextId: params.exceptionDetails.executionContextId,
         });
+        if (realm === undefined) {
+          // Ignore exceptions not attached to any realm.
+          return;
+        }
```
sadym-chromium, line 154: Same here. We probably should at least log it.

src/bidiMapper/domains/log/LogManager.ts
```diff
@@ -140,31 +148,39 @@ export class LogManager {
...
+                type: 'event',
+                method: ChromiumBidi.Log.EventNames.LogEntryAdded,
+                params: {
+                  level: Log.Level.Error,
+                  source: {
+                    realm: realm.realmId,
+                    ...(realm instanceof WindowRealm
+                      ? {context: realm.browsingContext.id}
+                      : {}),
+                  },
```
sadym-chromium, line 170: NIT: realm can have a method `source`, returning the realm and context if needed

src/bidiMapper/domains/script/Realm.ts
```diff
@@ -171,7 +154,10 @@ export class Realm {
 
     if (deepSerializedValue.type === 'node') {
       if (Object.hasOwn(bidiValue, 'backendNodeId')) {
-        let navigableId = this.navigableId;
+        let navigableId = 'UNKNOWN';
+        if (this instanceof WindowRealm) {
```
sadym-chromium, line 158: This line does not look right. The base class behavior depends on if the current instance of some specific derived  type.

sadym-chromium: You can get back method `get navigableId()` returning `UNKNOWN` and override it in the `WindowRealm` instead. BTW, having the `UNKNOWN` navigable id  worth logging as warn.

jrandolf: I'd rather not have unnecessary abstract methods. `instanceof` is meant for this purpose.

src/bidiMapper/domains/script/Realm.ts
```diff
@@ -833,16 +812,176 @@ export class Realm {
...
         },
-      },
-      this.browsingContextId
+        browsingContext.id
+      );
+    }
+  }
+}
+
+export class WindowRealm extends Realm {
```
sadym-chromium, line 831: Move to a separate file please

src/bidiMapper/domains/script/Realm.ts
```diff
@@ -833,16 +812,176 @@ export class Realm {
...
+      argumentsLocalValues,
+      awaitPromise,
+      resultOwnership,
+      serializationOptions,
+      userActivation
+    );
+  }
+}
+
+export class DedicatedWorkerRealm extends Realm {
```
sadym-chromium, line 934: Move to a separate file please

src/bidiMapper/domains/script/Realm.ts
```diff
@@ -265,47 +238,39 @@ export class Realm {
...
-        return {
-          // TODO: add proper owners.
-          //  https://github.com/GoogleChromeLabs/chromium-bidi/issues/1667
-          owners: [] as any as [Script.Realm],
-          realm: this.realmId,
-          origin: this.origin,
-          type: this.type,
-        };
-    }
+  abstract get associatedBrowsingContexts(): BrowsingContextImpl[];
```
sadym-chromium, line 254: Can a realm have more then 1 associated browsing context?

sadym-chromium (changes requested):
As long as you separated window and worker realms, please move their specific logic to those classes as well

src/bidiMapper/domains/script/Realm.ts
jrandolf: Yes, for example a shared worker can be associated with multiple browsing contexts.

src/bidiMapper/domains/script/Realm.ts
jrandolf: We've ran into issues before in Puppeteer due to separating classes from their base because of circular imports. It's possible to separate the classes into a separate file, but you would need an entry point that takes care of the JS hoisting. Tl;dr, it's much harder for people to work with the code if we try separating them.

src/bidiMapper/domains/script/Realm.ts
jrandolf: See other comment.

src/bidiMapper/domains/script/Realm.ts
OrKoN: If Realm does not depend on WindowRealm it should be no problem separating it.

src/bidiMapper/domains/script/Realm.ts
sadym-chromium: ack

src/bidiMapper/domains/script/Realm.ts
sadym-chromium: Resolving to continue conversation in https://github.com/GoogleChromeLabs/chromium-bidi/pull/1709/files#r1453290301

src/bidiMapper/domains/script/Realm.ts
```diff
@@ -833,16 +735,256 @@ export class Realm {
...
+export class DedicatedWorkerRealm extends Realm {
+  readonly #ownerRealm: WindowRealm | DedicatedWorkerRealm;
+
+  constructor(
+    cdpClient: CdpClient,
+    eventManager: EventManager,
+    executionContextId: Protocol.Runtime.ExecutionContextId,
+    logger: LoggerFn | undefined,
+    origin: string,
+    ownerRealm: WindowRealm | DedicatedWorkerRealm,
```
sadym-chromium, line 950: Why cannot it be just any realm?

src/bidiMapper/domains/script/RealmStorage.ts
```diff
@@ -62,13 +62,21 @@ export class RealmStorage {
       }
       if (
         filter.browsingContextId !== undefined &&
-        filter.browsingContextId !== realm.browsingContextId
+        (!(realm instanceof WindowRealm) ||
+          filter.browsingContextId !== realm.browsingContext.id)
```
sadym-chromium, line 66: `realm.associatedBrowsingContexts.includes(filter.browsingContextId)`?

src/bidiMapper/domains/script/Realm.ts
```diff
@@ -833,16 +735,256 @@ export class Realm {
...
+  }
+
+  override get associatedBrowsingContexts(): [BrowsingContextImpl] {
+    let realm = this.#ownerRealm;
+    // The root of any realm chain is a window realm.
+    while (!(realm instanceof WindowRealm)) {
+      realm = realm.#ownerRealm;
+    }
+
+    return [realm.browsingContext];
```
sadym-chromium, line 976: `return this.#ownerRealm.associatedBrowsingContexts`?

sadym-chromium, line 976:
```suggestion
    return this.#ownerRealm.associatedBrowsingContexts;
```

src/bidiMapper/domains/script/Realm.ts
jrandolf: Even if it does not, I'd rather keep same classes together to prevent a potential problem in the future like the locators issue.

src/bidiMapper/domains/script/Realm.ts
sadym-chromium, line 950:
```suggestion
    ownerRealm: Realm,
```

src/bidiMapper/domains/script/Realm.ts
```diff
@@ -833,16 +735,256 @@ export class Realm {
...
+      awaitPromise,
+      resultOwnership,
+      serializationOptions,
+      userActivation
+    );
+  }
+}
+
+export class DedicatedWorkerRealm extends Realm {
+  readonly #ownerRealm: WindowRealm | DedicatedWorkerRealm;
```
sadym-chromium, line 942:
```suggestion
  readonly #ownerRealm: Realm;
```

src/bidiMapper/domains/context/BrowsingContextProcessor.ts
```diff
@@ -434,19 +429,17 @@ export class BrowsingContextProcessor {
...
         cdpTarget.cdpClient,
         this.#eventManager,
-        this.#sharedIdWithFrame,
-        this.#logger
+        id,
+        this.#logger,
+        serializeOrigin(origin),
+        this.#realmStorage.getRealm({
+          browsingContextId,
+        }) as WindowRealm,
```
sadym-chromium, line 440:
```suggestion
        this.#realmStorage.getRealm({
          browsingContextId,
        }),
```

src/bidiMapper/domains/script/RealmStorage.ts
jrandolf: `associatedBrowsingContexts` are not necessarily the owners of the given realm. The only realm that has a browsing context is `WindowRealm`. Every other realm is indirectly related. Let me rename this to better imply this.

sadym-chromium, line 66:
```suggestion
        (filter.browsingContextId !== undeinfed && !realm.associatedBrowsingContexts.includes(filter.browsingContextId))
```

src/bidiMapper/domains/context/BrowsingContextProcessor.ts
jrandolf: This doesn't help because `getRealm` is not typed.

src/bidiMapper/domains/script/Realm.ts
jrandolf: It is generally only `WindowRealm` and `DedicatedWorkerRealm`, but I guess the distinction doesn't matter.

src/bidiMapper/domains/context/BrowsingContextProcessor.ts
sadym-chromium: Why `DedicatedWorkerRealm` should care of the owner realm type?

jrandolf: This comment is obsolete anyway. We already changed it to be generic in other places.

src/bidiMapper/domains/context/BrowsingContextProcessor.ts
```diff
@@ -434,19 +425,17 @@ export class BrowsingContextProcessor {
...
         cdpTarget.cdpClient,
         this.#eventManager,
-        this.#sharedIdWithFrame,
-        this.#logger
+        id,
+        this.#logger,
+        serializeOrigin(origin),
+        this.#realmStorage.getRealm({
+          browsingContextId,
+        }),
```
sadym-chromium, line 436: You probably want to have window realm here:
```suggestion
        this.#realmStorage.getRealm({
          browsingContextId,
          type: 'window'
        }),
```

jrandolf: Only window realms carry a browsing context directly, so this is clear by definition.

src/bidiMapper/domains/context/BrowsingContextProcessor.ts
```diff
@@ -348,27 +349,31 @@ export class BrowsingContextProcessor {
       params
     );
 
+    const createCdpTarget = () => {
```
sadym-chromium, line 352: Move to private class method `#createCdpTarget`

src/bidiMapper/domains/script/WindowRealm.ts
```diff
@@ -0,0 +1,205 @@
...
+      .getAllContexts()
+      .find((context) => context.navigableId === navigableId);
+    return maybeBrowsingContext?.id ?? 'UNKNOWN';
+  }
+
+  get browsingContext(): BrowsingContextImpl {
+    return this.#browsingContextStorage.getContext(this.#browsingContextId);
+  }
+
+  override get indirectBrowsingContexts(): [BrowsingContextImpl] {
```
sadym-chromium, line 68: It can be a bit confusing in case of window realm, where it's direct. `associatedBrowsingContexts` was more clear. Consider rename back.

jrandolf: Sure.

src/bidiMapper/domains/script/RealmStorage.ts
sadym-chromium: I think the model of `associatedBrowsingContexts` is perfectly fine, as you can request any realm associated with the given BrowsingContext. In filter you can specify a `type`, if you want a `window` realm.

src/bidiMapper/domains/script/RealmStorage.ts
```diff
@@ -62,13 +63,21 @@ export class RealmStorage {
...
       ) {
         return false;
       }
       if (
         filter.navigableId !== undefined &&
-        filter.navigableId !== realm.navigableId
+        (!(realm instanceof WindowRealm) ||
+          filter.navigableId !== realm.browsingContext.navigableId)
```
sadym-chromium, line 74: A list of `associatedNavigables` can be a part of `realm` as well.

sadym-chromium: Actually, as long as `filter.navigableId` is not used anywhere, it can be simply removed.

src/bidiMapper/domains/script/RealmStorage.ts
```diff
@@ -62,13 +63,21 @@ export class RealmStorage {
...
+      ) {
+        return false;
+      }
+      if (
+        filter.sandbox !== undefined &&
+        (!(realm instanceof WindowRealm) || filter.sandbox !== realm.sandbox)
```
sadym-chromium, line 80: `filter.sandbox` is used in [`BrowsingContextImpl`](https://github.com/GoogleChromeLabs/chromium-bidi/blob/477d1f28c39f02777a4801acdcaea09311f1fdc1/src/bidiMapper/domains/context/BrowsingContextImpl.ts#L272), and the `sandbox` filtering logic can be removed there.

src/bidiMapper/domains/script/Realm.ts
```diff
@@ -171,7 +154,10 @@ export class Realm {
 
     if (deepSerializedValue.type === 'node') {
```
Lightning00Blade, line 144: Node type of `node` should throw if we try to evaluate them in a non `WindowRealm`, There is no DOM for other realms so no node should exist.

src/bidiMapper/domains/script/RealmStorage.ts
jrandolf: > Actually, as long as `filter.navigableId` is not used anywhere, it can be simply removed.

Seems like it!

src/bidiMapper/domains/script/Realm.ts
jrandolf: In that case, it automatically throws in the CDP call.

src/bidiMapper/domains/script/RealmStorage.ts
sadym-chromium: There can other from `window` realms associated with the given browsing context, and I would expect the filter `{browsingContextId: SOME_VALUE}` to return all of them.

src/bidiMapper/domains/script/Realm.ts
Lightning00Blade: So this check should live on the WindowRealm and not on the Realm base class

src/bidiMapper/domains/script/RealmStorage.ts
sadym-chromium: Eg BiDi command [`script.getRealms`](https://w3c.github.io/webdriver-bidi/#command-script-getRealms) should return not only window realms, but workers as well.

src/bidiMapper/domains/script/RealmStorage.ts
jrandolf: Not sure what you mean here. Why can the sandbox logic be removed?

src/bidiMapper/domains/script/RealmStorage.ts
jrandolf: That warrants it. Changed.

src/bidiMapper/domains/context/BrowsingContextProcessor.ts
sadym-chromium: https://github.com/GoogleChromeLabs/chromium-bidi/pull/1709/files#r1453401191

src/bidiMapper/domains/context/BrowsingContextProcessor.ts
```diff
@@ -430,23 +413,42 @@ export class BrowsingContextProcessor {
...
-        undefined,
+      const realm = new DedicatedWorkerRealm(
         cdpTarget.cdpClient,
         this.#eventManager,
-        this.#sharedIdWithFrame,
-        this.#logger
+        id,
+        this.#logger,
+        serializeOrigin(origin),
+        this.#realmStorage.getRealm({
```
sadym-chromium, line 447: What if the browsing context has JS sandboxes?

src/bidiMapper/domains/script/RealmStorage.ts
jrandolf: sgtm

src/bidiMapper/domains/context/BrowsingContextProcessor.ts
jrandolf: In theory it's possible, but we currently don't support it :/ But you're right, it should be defaultRealm

src/bidiMapper/domains/context/BrowsingContextProcessor.ts
```diff
@@ -430,25 +420,42 @@ export class BrowsingContextProcessor {
...
+      this.#preloadScriptStorage,
+      this.#networkStorage,
+      this.#acceptInsecureCerts,
+      this.#logger
+    );
+  }
+
   #workers = new Map<string, Realm>();
-  #handleWorkerTarget(cdpTarget: CdpTarget, browsingContextId: string) {
+  #handleWorkerTarget(cdpTarget: CdpTarget, realm: Realm) {
```
sadym-chromium, line 445: NIT
```suggestion
  #handleWorkerTarget(cdpTarget: CdpTarget, ownerRealm: Realm) {
```

src/bidiMapper/domains/script/Realm.ts
```diff
@@ -169,94 +141,32 @@ export class Realm {
...
+        bidiValue[i] = this.serializeForBiDi(bidiValue[i], internalIdMap);
       }
     }
     if (['object', 'map'].includes(deepSerializedValue.type)) {
       for (const i in bidiValue) {
         bidiValue[i] = [
-          this.#deepSerializedToBiDi(bidiValue[i][0], internalIdMap),
-          this.#deepSerializedToBiDi(bidiValue[i][1], internalIdMap),
+          this.serializeForBiDi(bidiValue[i][0], internalIdMap),
+          this.serializeForBiDi(bidiValue[i][1], internalIdMap),
```
sadym-chromium, line 158: I wonder in case of `window` realm, this would call window's `serializeForBiDi` or the base realm's

src/bidiMapper/domains/script/RealmStorage.ts
jrandolf: Sorry, I reverted it because we use the filter for the default realm.

sadym-chromium:
Please address the e2e failure and make sure it's not flaky

jrandolf:
> Please address the e2e failure and make sure it's not flaky. I wonder why the tests are failing and would get another view after it is addressed.

This is unrelated to this PR. It has to do with the browser implementation. I can disable this test if you want.

sadym-chromium:
> > Please address the e2e failure and make sure it's not flaky. I wonder why the tests are failing and would get another view after it is addressed.
> 
> This is unrelated to this PR. It has to do with the browser implementation. I can disable this test if you want.

Oh, no worries then. It used to be a worker-related test.

src/bidiMapper/domains/context/BrowsingContextProcessor.ts
```diff
@@ -415,7 +393,19 @@ export class BrowsingContextProcessor {
...
+          targetInfo,
+          sessionId
+        );
+        this.#handleWorkerTarget(
+          cdpTarget,
+          this.#realmStorage.getRealm({
+            browsingContextId: browsingContext.id,
+            type: 'window',
+            sandbox: undefined,
+          })
```
sadym-chromium, line 407: nit: I'd prefer exposing `browsingContext.defaultRealm` here to hide the logic.

src/bidiMapper/domains/script/RealmStorage.ts
sadym-chromium: The default realm logic can be moved to BrowsingContext. But this is optional.

src/bidiMapper/domains/context/BrowsingContextProcessor.ts
jrandolf: This logic will eventually change since worker realms can be nested. I don't think this change matters too much.

sadym-chromium (approved):

sadym-chromium:
Looks like this PR caused Puppeteer tests failing
https://github.com/puppeteer/puppeteer/pull/11795

PR #1713, opened 2024-01-17
tests: add more screenshot tests

jrandolf-2:
Adds a screenshot fixture and adds more test.


OrKoN:
Looks like tests fail

src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -771,12 +771,12 @@ export class BrowsingContextImpl {
     switch (params.origin) {
       case 'document': {
         script = String(() => {
-          const element = document.documentElement;
+          const rect = document.body.getBoundingClientRect();
```
OrKoN, line 774: Would not it make the problem worse meaning given that the body can be made of size 0x0 while the content is still visible?

OrKoN (changes requested):

src/bidiMapper/domains/context/BrowsingContextImpl.ts
jrandolf: Yea, that's a good argument.

OrKoN:
@sadym-chromium assert_screenshot looks useful, let's revert the mapper changes though. WDYT?

sadym-chromium:
> @sadym-chromium assert_screenshot looks useful, let's revert the mapper changes though. WDYT?

agree

PR #1732, opened 2024-01-19
build: call `cddlconv` direcly

sadym-chromium:
Temp work-around to fix issue with generating permission types https://github.com/GoogleChromeLabs/chromium-bidi/actions/runs/7582257569/job/20651521640

### DIFF
--- a/.github/workflows/update-bidi-types.yml
+++ b/.github/workflows/update-bidi-types.yml
@@ -88,15 +88,16 @@ jobs:
           name: permissions-cddl
       - name: Generate TypeScript types for the main spec
         run: npm run bidi-types -- --cddl-file all.cddl
-      - name: Generate TypeScript types for Permissions spec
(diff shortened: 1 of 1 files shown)


.github/workflows/update-bidi-types.yml
```diff
@@ -89,7 +89,7 @@ jobs:
       - name: Generate TypeScript types for the main spec
         run: npm run bidi-types -- --cddl-file all.cddl
       - name: Generate TypeScript types for Permissions spec
-        run: npm run bidi-types -- --cddl-file permissions.cddl --ts-file src/protocol/generated/webdriver-bidi-permissions.ts --zod-file src/protocol-parser/generated/webdriver-bidi-permissions.ts
+        run: cddlconv -f zod permissions.cddl > src/protocol/generated/webdriver-bidi-permissions.ts
```
OrKoN, line 92: this is the wrong destination and we need the FILE_HEADER. We also need TS types.

OrKoN (changes requested):

PR #1735, opened 2024-01-19
test: extend e2e with chromedriver

sadym-chromium:
* extend e2e with chromedriver
* align js runner with BiDi protocol
* provide headless in capabilities

Blocked by:
1. Chromedriver converts `id` in response to double instead of `js-uint`.
2. JS runner disregards `headless` args from capabilities.


Lightning00Blade:
@sadym-chromium is his still needed? If not please close.

PR #1740, opened 2024-01-22
refactor: create an instance

Lightning00Blade:
Comments on places where changes were made.

### DIFF
--- a/src/bidiServer/WebSocketServer.ts
+++ b/src/bidiServer/WebSocketServer.ts
@@ -55,333 +55,337 @@ type SessionOptions = {
 };
 
 export class WebSocketServer {
-  static #sessions = new Map<string, Session>();
+  #sessions = new Map<string, Session>();
+  #port: number;
+  #channel: ChromeReleaseChannel;
+  #headless: boolean;
+  #verbose: boolean;
 
-  /**
-   * @param bidiPort Port to start ws server on.
-   * @param channel
-   * @param headless
-   * @param verbose
-   */
-  static run(
-    bidiPort: number,
+  #server: http.Server;
+  #wsServer: websocket.server;
+
+  constructor(
+    port: number,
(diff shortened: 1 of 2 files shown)


src/bidiServer/WebSocketServer.ts
```diff
@@ -55,333 +55,337 @@ type SessionOptions = {
...
+    this.#channel = channel;
+    this.#headless = headless;
+    this.#verbose = verbose;
+
+    this.#server = http.createServer(this.#onRequest.bind(this));
+    this.#wsServer = new websocket.server({
+      httpServer: this.#server,
+      autoAcceptConnections: false,
+    });
+    this.#wsServer.on('request', this.#onWsRequest.bind(this));
```
Lightning00Blade, line 83: Extract into separate functions.

src/bidiServer/WebSocketServer.ts
```diff
@@ -55,333 +55,337 @@ type SessionOptions = {
...
+    }
 
-    const wsServer: websocket.server = new websocket.server({
-      httpServer: server,
-      autoAcceptConnections: false,
-    });
+    // https://w3c.github.io/webdriver-bidi/#transport, step 2.
+    if (request.url === '/session') {
+      const bodyArray: Uint8Array[] = [];
+      const body = await new Promise<Buffer>((resolve, reject) => {
```
Lightning00Blade, line 106: Created a promise to reduce nesting.

wpt
OrKoN, line 1: exclude wpt?

Lightning00Blade: Strange my IDE usually does this automatically.  Removing.

sadym-chromium (approved):
Even though the reviewing of the whole changes are quite challenging, I assume it is no-op change. As long as the NodeJS Mapper runner is used only in CI, I believe it's safe to proceed.

PR #1744, opened 2024-01-23
feat: respect partition

sadym-chromium:
Apparently, the partition is a required field, not related to "partitioned" cookies. This PR intended to respect this fact.

The origin is a protocol + domain, no port of path.

**Open question:**
* When cookies are requested for the given origin, we should return only cookies accessible from the provided origin. Currently, we get all the cookies, and filter out those accessible by checking by `domain` and [`CookieSourceScheme`](https://chromedevtools.github.io/devtools-protocol/tot/Network/#type-CookieSourceScheme) fields.
  * Is there a way to avoid getting all the cookies?
  * Is there a way to check the cookie availability from the given origin? Eg [`Network.getCookies`](https://chromedevtools.github.io/devtools-protocol/tot/Network/#method-getCookies) does not seem to work, as it filters our cookies by `path`.

* Is there a way to say if the cookie was not set? Eg setting cookie with domain `.com` successes. But it is obviously is not set.

TODO:
* respect port

### DIFF
--- a/src/bidiMapper/domains/storage/StorageProcessor.ts
+++ b/src/bidiMapper/domains/storage/StorageProcessor.ts
@@ -30,6 +30,7 @@ import type {LoggerFn} from '../../../utils/log.js';
 import {LogType} from '../../../utils/log.js';
 import type {BrowsingContextStorage} from '../context/BrowsingContextStorage.js';
 import {NetworkProcessor} from '../network/NetworkProcessor.js';
+
 const HTTPS_SCHEME = 'https:';
 
 /**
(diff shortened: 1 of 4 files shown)


OrKoN:
I think it should not be required. 

> not related to "partitioned" cookies.

Could you elaborate? only partitioned cookies get added to a separate storage partition. Regular cookies exist in one partitions, e.g., the default user contexts.

sadym-chromium:
> Could you elaborate? only partitioned cookies get added to a separate storage partition. Regular cookies exist in one partitions, e.g., the default user contexts.

The PR is in draft, and I try to figure out what exactly is the expected behavior here.

PR #1748, opened 2024-01-25
chore: update deps

Lightning00Blade:
Updates Rollup as Chromium is now at Node 20.


tests/network/test_network.py
```diff
@@ -20,6 +20,12 @@
                           subscribe, wait_for_event)
 
 
+def compute_response_headers_size(headers) -> int:
```
Lightning00Blade, line 23: Drive-by: extract to top.

OrKoN (approved):
LGTM % precommit

PR #1759, opened 2024-01-26
feat: add CPD specific field in cookies

sadym-chromium:
Extend [`network.Cookie`](https://w3c.github.io/webdriver-bidi/#type-network-Cookie) with CDP-specific fields both for setting and getting cookies.
* Move the `cdp-to-bidi` logic to network helper.
* Use the unified `cdp-to-bidi` in network and in storage domains.


src/bidiMapper/domains/network/Cookie.ts
```diff
@@ -0,0 +1,159 @@
...
+const CDP_SPECIFIC_PREFIX = 'goog:';
+
+/**
+ * Utility class for presenting BiDi `network.Cookie` and it's conversion with
+ *  CDP's `Network.Cookie` and `Network.CookieParam`.
+ *  * https://w3c.github.io/webdriver-bidi/#type-network-Cookie
+ *  * https://chromedevtools.github.io/devtools-protocol/tot/Network/#type-Cookie
+ *  * https://chromedevtools.github.io/devtools-protocol/tot/Network/#type-CookieParam
+ */
+export class Cookie {
```
Lightning00Blade, line 35: Note: Usually only static Classes are forbidden/discouraged. as they are close to name spacing. Personally I have no problem, but we should align as a team on this and if we think it make sense disable relative rules for EsLint.

Lightning00Blade: https://google.github.io/styleguide/tsguide.html#container-classes

sadym-chromium: Moved to `NetworkUtils`

OrKoN (approved):

src/bidiMapper/domains/network/NetworkUtils.ts
```diff
@@ -21,7 +21,11 @@
  */
 import type {Protocol} from 'devtools-protocol';
 
-import type {Network} from '../../../protocol/protocol.js';
+import {
+  InvalidArgumentException,
+  UnsupportedOperationException,
+} from '../../../protocol/ErrorResponse';
```
Lightning00Blade, line 27: We need the rule that forces the use of extensions as this may brake in Puppeteer under some cases.

OrKoN: Good point, we should always add an extension for running in the browser

sadym-chromium: missed it. Automatic check would be awesome

PR #1768, opened 2024-01-29
fix: add pointer props to input up events

jrandolf-2:


OrKoN:
Which issue does it fix?

src/bidiMapper/domains/input/ActionDispatcher.ts
```diff
@@ -319,6 +326,12 @@ export class ActionDispatcher {
               {
                 x,
                 y,
+                ...getRadii(width ?? 1, height ?? 1),
```
OrKoN, line 329: let's add a test?

PR #1780, opened 2024-01-30
refactor: use single type for channels

Lightning00Blade:
If I understand correctly `Channel` on the subscribe level has nothing to do with `Script.Channel`.


sadym-chromium:
Maybe rename `Channel` to something like `BidiPlusChannel` or `TransportChannel`?

Lightning00Blade:
@sadym-chromium Some other changes as well, so please review.

sadym-chromium (approved):

PR #1781, opened 2024-01-30
feat: support user context in cookies API

sadym-chromium:
Get `userContext` from partition or from the linked browsing context.

### DIFF
--- a/src/bidiMapper/domains/storage/StorageProcessor.ts
+++ b/src/bidiMapper/domains/storage/StorageProcessor.ts
@@ -50,9 +50,7 @@ export class StorageProcessor {
     const cdpResponse = await this.#browserCdpClient.sendCommand(
       'Storage.getCookies',
       {
-        ...(partitionKey.userContext === undefined
-          ? {}
-          : {browserContextId: partitionKey.userContext}),
+   browserContextId: partitionKey.userContext,
       }
     );
 
--- a/src/bidiMapper/domains/storage/StorageProcessor.ts
+++ b/src/bidiMapper/domains/storage/StorageProcessor.ts
@@ -50,7 +50,7 @@ export class StorageProcessor {
     const cdpResponse = await this.#browserCdpClient.sendCommand(
       'Storage.getCookies',
       {
-   browserContextId: partitionKey.userContext,
+        browserContextId: partitionKey.userContext,
       }
     );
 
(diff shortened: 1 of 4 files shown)


src/bidiMapper/domains/storage/StorageProcessor.ts
```diff
@@ -79,6 +83,9 @@ export class StorageProcessor {
     try {
       await this.#browserCdpClient.sendCommand('Storage.setCookies', {
         cookies: [cdpCookie],
+        ...(partitionKey.userContext === undefined
+          ? {}
+          : {browserContextId: partitionKey.userContext}),
```
OrKoN, line 88:
```suggestion
        {browserContextId: partitionKey.userContext}
```

tests/conftest.py
```diff
@@ -377,3 +380,14 @@ async def iframe_id(websocket, context_id: str, html_iframe_same_origin, html):
     await goto_url(websocket, iframe_id, html("<h1>FRAME</h1>"))
 
     return iframe_id
+
+
+@pytest_asyncio.fixture
+async def user_context(websocket):
```
OrKoN, line 386:
```suggestion
async def create_user_context(websocket):
```

tests/conftest.py
```diff
@@ -377,3 +380,14 @@ async def iframe_id(websocket, context_id: str, html_iframe_same_origin, html):
...
+    """Create a new user context and return its id."""
+    result = await execute_command(websocket, {
+        "method": "browser.createUserContext",
+        "params": {}
+    })
+
+    return result['userContext']
```
OrKoN, line 393: shall we add auto-remove?

tests/conftest.py
sadym-chromium: The fixture is supposed to be used as a parameter in the test, and will already provide the result.

tests/conftest.py
sadym-chromium: We don't clean up the browser state between tests, so no need in this yet.

tests/conftest.py
OrKoN: There is a fixture called create_user_context in WPT and I think to makes sense to match the style, also `create_context` would be a similar feature.

sadym-chromium: Currently, the `create_context` returns a function, which can be called in test. Al long as this fixture follows the naming pattern used in this file, I suggest moving this out of scope of this PR.

OrKoN: well, I am not sure there is a pattern in this file :) there is `context_id` fixture, should this one be called `user_context_id`?

OrKoN (approved):

tests/conftest.py
sadym-chromium: `user_context_id` SGTM. Done.

PR #1785, opened 2024-01-30
fix: enable cdp Network domain when needed

Lightning00Blade:
It seems like some E2E use CDP methods and they are now wrong, will address them before merging but please review as is as WPT test pass.

### DIFF
--- a/src/bidiMapper/domains/context/CdpTarget.ts
+++ b/src/bidiMapper/domains/context/CdpTarget.ts
@@ -137,7 +137,8 @@ export class CdpTarget {
     // Check if the network domain is enabled globally.
     const enabledNetwork =
       this.#eventManager.subscriptionManager.isSubscribedToModule(
-        BiDiModule.Network
+        BiDiModule.Network,
+        this.#id
       );
 
     try {
--- a/src/bidiMapper/domains/context/CdpTarget.ts
+++ b/src/bidiMapper/domains/context/CdpTarget.ts
@@ -183,18 +183,18 @@ export class CdpTarget {
     });
   }
 
-  async toggleNetworkIfNeeded(enable: boolean): Promise<void> {
+  async toggleNetworkIfNeeded(enabled: boolean): Promise<void> {
+    if (enabled === this.#networkDomainEnabled) {
+      return;
+    }
+
+    this.#networkDomainEnabled = enabled;
     try {
-      if (enable && !this.#networkDomainEnabled) {
-        this.#networkDomainEnabled = true;
-        await this.#cdpClient.sendCommand('Network.enable');
-      } else if (!enable && this.#networkDomainEnabled) {
-        this.#networkDomainEnabled = false;
-        await this.#cdpClient.sendCommand('Network.disable');
-      }
-    } catch (error) {
-      this.#networkDomainEnabled = !this.#networkDomainEnabled;
-      throw error;
+      await this.#cdpClient.sendCommand(
+        this.#networkDomainEnabled ? 'Network.enable' : 'Network.disable'
+      );
+    } catch (err) {
+      this.#networkDomainEnabled = !enabled;
     }
   }
 
--- a/src/bidiMapper/domains/context/CdpTarget.ts
+++ b/src/bidiMapper/domains/context/CdpTarget.ts
@@ -18,6 +18,7 @@
 import type {Protocol} from 'devtools-protocol';
 
 import type {CdpClient} from '../../../cdp/CdpClient.js';
+import {BiDiModule} from '../../../protocol/chromium-bidi.js';
 import {Deferred} from '../../../utils/Deferred.js';
 import type {LoggerFn} from '../../../utils/log.js';
 import type {Result} from '../../../utils/result.js';
@@ -38,8 +39,9 @@ export class CdpTarget {
   readonly #preloadScriptStorage: PreloadScriptStorage;
   readonly #networkStorage: NetworkStorage;
 
-  readonly #targetUnblocked = new Deferred<Result<void>>();
+  readonly #unblocked = new Deferred<Result<void>>();
   readonly #acceptInsecureCerts: boolean;
+  #networkDomainEnabled = false;
 
   static create(
     targetId: Protocol.Target.TargetID,
@@ -94,7 +96,7 @@ export class CdpTarget {
 
   /** Returns a deferred that resolves when the target is unblocked. */
   get unblocked(): Deferred<Result<void>> {
-    return this.#targetUnblocked;
+    return this.#unblocked;
   }
 
   get id(): Protocol.Target.TargetID {
@@ -132,6 +134,12 @@ export class CdpTarget {
    * Enables all the required CDP domains and unblocks the target.
    */
   async #unblock() {
+    // Check if the network domain is enabled globally.
+    const enabledNetwork =
+      this.#eventManager.subscriptionManager.isSubscribedToModule(
+        BiDiModule.Network
+      );
+
     try {
       await Promise.all([
         this.#cdpClient.sendCommand('Runtime.enable'),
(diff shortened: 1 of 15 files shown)


OrKoN (commented):
I would not change the subscribe/unsubscribe methods to be async as it might lead to not atomic changes and race conditions. Let's instead emit an event from the event manager when there is a network change and listen for that event in the CDP target. If we need to wait for side effects to complete let's do it in some other way. WDYT?

OrKoN (approved):

sadym-chromium:
> I would not change the subscribe/unsubscribe methods to be async as it might lead to not atomic changes and race conditions. Let's instead emit an event from the event manager when there is a network change and listen for that event in the CDP target. If we need to wait for side effects to complete let's do it in some other way. WDYT?

There can be a message cache, and the `subscribe` command should be finished AFTER all the cached events are sent. So the `subscribe` method can be async. But we should be careful with not sending new events BEFORE the subscription command is finished.

src/bidiMapper/domains/context/CdpTarget.ts
```diff
@@ -40,6 +41,7 @@ export class CdpTarget {
 
   readonly #targetUnblocked = new Deferred<Result<void>>();
   readonly #acceptInsecureCerts: boolean;
+  #networkDomainEnabled = false;
```
sadym-chromium, line 44: Maybe make it deferred?

src/bidiMapper/domains/context/CdpTarget.ts
```diff
@@ -173,6 +174,27 @@ export class CdpTarget {
...
+    const enableNetwork =
+      this.#eventManager.subscriptionManager.isSubscribedToModule(
+        this.#id,
+        BiDiModule.Network
+      );
+
+    try {
+      if (enableNetwork && !this.#networkDomainEnabled) {
+        this.#networkDomainEnabled = true;
+        await this.#cdpClient.sendCommand('Network.enable');
```
sadym-chromium, line 187: Maybe more atomic?
```suggestion
        await this.#cdpClient.sendCommand('Network.enable').catch(e=>{
            this.#networkDomainEnabled = false;
            throw e;
        |});
```

src/bidiMapper/domains/session/EventManager.ts
```diff
@@ -188,17 +192,31 @@ export class EventManager extends EventEmitter<EventManagerEventsMap> {
         }
       }
     }
+
+    await this.toggleModulesIfNeeded();
```
sadym-chromium, line 196: This should be done before `this.#subscriptionManager.subscribe`. Otherwise the events will start coming before the subscription is done.

src/bidiMapper/domains/context/CdpTarget.ts
Lightning00Blade: I think the suggestion here is NoOp.

I am not sure about this the current implementation does the the following.
If changes -> enable/disable and update flag.
If no changes -> no error can be thrown so we never go to catch.
If command fails -> flip to previous state.

If second change comes while we are waiting for response we should be good as well as both commands will be send to the same process in Chrome and will be in queue from what i understand.

src/bidiMapper/domains/context/CdpTarget.ts
```diff
@@ -173,6 +174,27 @@ export class CdpTarget {
     });
   }
 
+  async toggleNetworkIfNeeded(): Promise<void> {
+    const enableNetwork =
+      this.#eventManager.subscriptionManager.isSubscribedToModule(
+        this.#id,
```
sadym-chromium, line 180: We imply the fact there is always a BrowsingContext with the very same `#id`. Maybe move the `this.#eventManager.subscriptionManager.isSubscribedToModule` to `BrowsingContext.toggleDomainsIfNeeded()`, and call `cdpTarget.toggleNetwork(enable: bool)` only when needed?

src/bidiMapper/domains/session/EventManager.ts
Lightning00Blade: Hmm I guess so but that makes the current implementation not work as it relies on the Maps in SessionManager.

src/bidiMapper/domains/context/CdpTarget.ts
sadym-chromium: Easier concurrency handling

src/bidiMapper/domains/context/CdpTarget.ts
Lightning00Blade: Sound OK.

src/bidiMapper/domains/session/EventManager.ts
sadym-chromium: > Hmm I guess so but that makes the current implementation not work as it relies on the Maps in SessionManager.

Right. Then we can try this approach.

src/bidiMapper/domains/context/CdpTarget.ts
sadym-chromium: We can put concurrent enabling/disablling commands in a queue which will reduce a chance of race condition.

sadym-chromium: This suggestion is optional though

src/bidiMapper/domains/context/CdpTarget.ts
sadym-chromium: Agree that your solution will work, but I'm afraid it's easier to screw up with it later on. Up for you.

src/bidiMapper/domains/context/CdpTarget.ts
```diff
@@ -132,6 +134,12 @@ export class CdpTarget {
    * Enables all the required CDP domains and unblocks the target.
    */
   async #unblock() {
+    // Check if the network domain is enabled globally.
+    const enabledNetwork =
```
sadym-chromium, line 138: If the browsing context is OOPiF, there already can be browsing context - specific subscriptions, so it should be checked as well

Lightning00Blade: I think the change should not be related to this PR, if you look at the current implementation we already always subscribe.

src/bidiMapper/domains/context/CdpTarget.ts
```diff
@@ -159,20 +169,35 @@ export class CdpTarget {
...
     }
 
-    this.#targetUnblocked.resolve({
+    this.#unblocked.resolve({
       kind: 'success',
       value: undefined,
     });
   }
 
+  async toggleNetworkIfNeeded(enable: boolean): Promise<void> {
```
OrKoN, line 186:
```suggestion
  async toggleNetworkIfNeeded(enabled: boolean): Promise<void> {
```

src/bidiMapper/domains/context/CdpTarget.ts
```diff
@@ -159,20 +169,35 @@ export class CdpTarget {
...
+    try {
+      if (enable && !this.#networkDomainEnabled) {
```
OrKoN, line 188:
```
      if (enabled === this.#networkDomainEnabled) {return}
      this.#networkDomainEnabled = enabled;
      try {     
         await this.#cdpClient.sendCommand(this.#networkDomainEnabled ? 'Network.enable' : 'Network.disable'); 
      } catch (err) {
          this.#networkDomainEnabled = !enabled;
      }
```

src/bidiMapper/domains/session/EventManager.ts
```diff
@@ -188,17 +192,31 @@ export class EventManager extends EventEmitter<EventManagerEventsMap> {
...
     for (const name of eventNames) {
       assertSupportedEvent(name);
     }
     this.#subscriptionManager.unsubscribeAll(eventNames, contextIds, channel);
+    await this.toggleModulesIfNeeded();
+  }
+
+  async toggleModulesIfNeeded(): Promise<void> {
+    // TODO: Only update changed subscribers
+    await Promise.all(
```
OrKoN, line 213: Go over all targets or or all browsing contexts instead of only top-level contexts

src/bidiMapper/domains/session/SubscriptionManager.spec.ts
```diff
@@ -486,4 +486,57 @@ describe('SubscriptionManager', () => {
...
+      );
+
+      expect(
+        subscriptionManager.isSubscribedToModule(
+          ChromiumBidi.BiDiModule.Network,
+          SOME_CONTEXT
+        )
+      ).to.equal(true);
+    });
+    it('should return true specific context subscription', () => {
```
OrKoN, line 505: Is it possible to write a test for a specific frame tree?

OrKoN (approved):

PR #1788, opened 2024-01-30
feat: allow Chrome-specific params in user contexts

OrKoN:
The proxy servers are somewhat difficult to test. Does the pytest server support a proxy mode? Any test suggestions for vendor specific attributes?

### DIFF
--- a/src/bidiMapper/CommandProcessor.ts
+++ b/src/bidiMapper/CommandProcessor.ts
@@ -153,7 +153,7 @@ export class CommandProcessor extends EventEmitter<CommandProcessorEventsMap> {
       case 'browser.close':
         return this.#browserProcessor.close();
       case 'browser.createUserContext':
-        return await this.#browserProcessor.createUserContext();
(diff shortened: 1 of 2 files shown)


OrKoN:
@sadym-chromium do you have an advice for on how to test this best?

sadym-chromium:
> @sadym-chromium do you have an advice for on how to test this best?

We can implement the proxy along with our http server in pytest: https://stackoverflow.com/a/4412630/22971202

OrKoN:
@sadym-chromium it does not look like this is a correct implementation of the proxy. Also https://github.com/csernazs/pytest-httpserver/issues/39 is not available yet.

OrKoN:
I will look into making a werkzeug proxy.

OrKoN:
Alright it does not seem to work as expected

OrKoN:
I suggest that we test it via Puppeteer that has a flexible proxy server. WDYT?

PR #1836, opened 2024-02-05
chore: fix not gc after detach

Lightning00Blade:
We still have a reference to the `NetworkManager` in the listener, this may/should prevent it from being garbage collected.


src/bidiMapper/domains/network/NetworkManager.ts
```diff
@@ -87,13 +87,21 @@ export class NetworkManager {
...
+    };
+
     cdpTarget.browserCdpClient.on(
       'Target.detachedFromTarget',
-      (params: Protocol.Target.DetachedFromTargetEvent) => {
-        if (cdpTarget.cdpClient.sessionId === params.sessionId) {
-          networkManager.#networkStorage.disposeRequestMap();
-        }
-      }
+      detachedListener
```
OrKoN, line 104: Is the issue hypothetical or real? I don't think there is a difference between the before and the after version? The listener still retains the target and the network manager.

Lightning00Blade: Hypothetical, but from my understanding `detachedListener` loses scope after the return from `create` and the only reference lives inside the `EventEmitter`. Also we explicitly call the `off` method on the `EventEmitter` now.
Graph of ref should be `EventEmitter` -> `detachedListener` -> `networkManager`, so calling `off` should break the first chain.

OrKoN: It also looses scope when it is inline, calling the `off` should release it just as well.

Lightning00Blade: How do we get the reference to call `off` for your case?

PR #1838, opened 2024-02-06
refactor: re-work network interception to be non-flaky 

Lightning00Blade:
- the PR moves the interception logic to NetworkStorage since interception is tightly coupled with storage. It removes the NetworkManager class which only provided basic event listeners.
- NetworkStorage is multi-target so that it is able to control interception across all targets including the browser target.
- The Fetch/Network domains are enabled on the page/frame target levels to intercept all requests from the targets.
- The code is modified to await for interception to conclude before sending BiDi events.
- Hopefully, it solves existing flakiness in the interception.
- TODO: credentials
- TODO: re-enable disabled tests


src/bidiMapper/CommandProcessor.ts
```diff
@@ -91,7 +91,7 @@ export class CommandProcessor extends EventEmitter<CommandProcessorEventsMap> {
     this.#parser = parser;
     this.#logger = logger;
 
-    const networkStorage = new NetworkStorage();
+    const networkStorage = new NetworkStorage(browserCdpClient);
```
OrKoN, line 94: I think the cdp client should be in the networkprocessor and not the storage. Any reason for giving to the storage?

Lightning00Blade: `Processor` in Mapper is closer to `CommandProcessor` so `NetworkProcessor` is an entry point for network module commands.

src/bidiMapper/domains/network/NetworkRequest.ts
```diff
@@ -53,9 +51,10 @@ export class NetworkRequest {
    * The identifier for a request resulting from a redirect matches that of the
    * request that initiated it.
    */
-  #requestId: Network.Request;
+  #id: Network.Request;
+
+  #fetchId: Protocol.Fetch.RequestId | undefined = undefined;
```
OrKoN, line 56:
```suggestion
  #fetchId?: Protocol.Fetch.RequestId;
```

src/bidiMapper/domains/network/NetworkRequest.ts
```diff
@@ -182,13 +188,31 @@ export class NetworkRequest {
...
+    }
     this.#emitEventsIfReady();
   }
 
   onRequestWillBeSentExtraInfoEvent(
     event: Protocol.Network.RequestWillBeSentExtraInfoEvent
   ) {
     this.#request.extraInfo = event;
+    if (
+      this.#networkStorage.requestBlockedBy(
```
Lightning00Blade, line 208: I think we should use the CdpTarget to find out if interception is enabled, it's the source of truth if an even is going to be sent.

OrKoN: It's not about the interception being enabled, it's about knowing if the request should be intercepted irrespective of the underlying CDP status

Lightning00Blade:
Puppeteer also fails with the latest Canary so we should be OK, I believe the other test to be flaky 
https://github.com/puppeteer/puppeteer/actions/runs/7905564059/job/21578469250

OrKoN (approved):

OrKoN:
I think it is ready to be landed.

PR #1846, opened 2024-02-09
fix: limit the mapper tab to 200 lines of logs

### LINKED ISSUE #1839: Out of Memory when BiDi is enabled
The DOM of the BiDi-CDP Mapper tab does grow unlimited, this results in a out of memory issue in this tab.
Depending on the available RAM it might take some time to run into this issue.

Minimal code to reproduce the issue using selenium is listed below.
When running real code this usually happens faster, as more things are logged. 

```
        var options = new org.openqa.selenium.chrome.ChromeOptions();
        options.setCapability("webSocketUrl", true);

        WebDriver driver = new org.openqa.selenium.chrome.ChromeDriver(options);

        for (int i = 0; i < 1_000_000_000; i++) {
            driver.navigate().refresh();
        }
```

OrKoN:
Issue #1839

### DIFF
--- a/src/bidiTab/mapperTabPage.ts
+++ b/src/bidiTab/mapperTabPage.ts
@@ -93,4 +93,7 @@ export function log(logPrefix: LogPrefix, ...messages: unknown[]) {
   lineElement.className = 'pre';
   lineElement.textContent = [logPrefix, ...messages].map(stringify).join(' ');
   typeLogContainer.appendChild(lineElement);
+  if (typeLogContainer.childNodes.length > 200) {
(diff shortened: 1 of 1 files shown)


Lightning00Blade (approved):
LGTM

Should we not do any DOM appends if we are in a production mode?

OrKoN:
I don't think there is a production mode in chromedriver.

Lightning00Blade:
The concept there is "--verbose" else don't log anything that's not errors/warns. It's a micro optimisation to be fair but we should consider. How about we keep the issue open and assign it to me to look into that at some point?

sadym-chromium (approved):

PR #1850, opened 2024-02-13
fix: fix navigation events

jrandolf-2:
This PR swaps out the deferred promises for event emission and uses a combination of `init` and `requestWillBeSent` for `navigationStarted`. We also introduce random UUIDs for fragment navigation.

Closes https://github.com/GoogleChromeLabs/chromium-bidi/issues/1858


src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -350,41 +352,32 @@ export class BrowsingContextImpl {
...
+        }
+
         this.#eventManager.registerEvent(
           {
             type: 'event',
             method: ChromiumBidi.BrowsingContext.EventNames.FragmentNavigated,
             params: {
               context: this.id,
-              navigation: null,
+              navigation: navigation ?? navigation ?? uuidv4(),
```
OrKoN, line 373:
```suggestion
              navigation: navigation ?? uuidv4(),
```

Lightning00Blade:
```suggestion
              navigation: navigation || uuidv4(),
```

src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -350,41 +352,32 @@ export class BrowsingContextImpl {
         this.#url = params.url;
         this.#navigation.withinDocument.resolve(params);
 
+        let url = this.#url;
+        let navigation = this.#fragmentIds.get(this.#url);
+        if (navigation === undefined) {
+          for (const [fragmentUrl, fragmentId] of [
```
Lightning00Blade, line 358: This can be done with `[...<MAP>].at(-1)` as we want the last item right?

src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -576,6 +569,21 @@ export class BrowsingContextImpl {
...
+    if (this.#loaderId) {
+      this.#eventManager.registerEvent(
+        {
+          type: 'event',
+          method: ChromiumBidi.BrowsingContext.EventNames.NavigationStarted,
+          params: {
+            context: this.id,
+            navigation: this.#loaderId,
+            timestamp: BrowsingContextImpl.getTimestamp(),
+            url: '',
```
Lightning00Blade, line 581: Can we do better than empty string?

src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -576,6 +569,21 @@ export class BrowsingContextImpl {
     this.#resetLifecycleIfFinished();
 
     this.#loaderId = loaderId;
+    if (this.#loaderId) {
+      this.#eventManager.registerEvent(
+        {
+          type: 'event',
+          method: ChromiumBidi.BrowsingContext.EventNames.NavigationStarted,
```
sadym-chromium, line 576: The PR title is misleading. `#documentChanged` is called after CDP call `Page.navigate` or after `Page.lifecycleEvent: init`, and has nothing to do with `requestWillBeSent`.

jrandolf:
We need to wait till https://github.com/w3c/webdriver-bidi/issues/662 resolves. Currently, CDP doesn't have an event that emits only for non-same-document navigation. The closest event is `Page.frameRequestedNavigation`, but this is only for renderer-initiated navigation.

https://chromium-review.googlesource.com/c/chromium/src/+/5300245 is implements an event that captures _all_ navigations (which is still technically not spec-compliant). If https://github.com/w3c/webdriver-bidi/issues/657 resolves to emit `navigationStarted` for all navigations, then the CL should work as is (+ tests), but if not, the CL needs to be modified to remove fragment navigations.

PR #1852, opened 2024-02-14
chore: use verbose logging by default for PR

Lightning00Blade:


package.json
```diff
@@ -48,7 +48,7 @@
       ]
     },
     "e2e:headless": {
-      "command": "pipenv run python -m pytest",
+      "command": "pipenv run python -m pytest --verbose",
```
Lightning00Blade, line 51: This will print the Test name to console when it start.

OrKoN (commented):
I am not sure we should output more by default. For now, you can use manual dispatch on your PR to get more details. I would propose that we event reduce the current output and only output test results.

OrKoN (approved):

PR #1862, opened 2024-02-20
chore: use new headless

jrandolf:


OrKoN (commented):
Looks like some tests need to be updated. Also, is it used for both WPT and e2e tests?

PR #1872, opened 2024-02-21
chore: fix empty setFiles case

jrandolf:


src/bidiMapper/domains/input/InputProcessor.ts
```diff
@@ -119,17 +112,50 @@ export class InputProcessor {
       );
     }
 
+    /**
+     * The zero-length array is a special case, it seems that
+     * DOM.setFileInputFiles does not actually update the files in that case, so
```
Lightning00Blade, line 117: Deferring to Alex as I am unsure what the guidance for changes to Chromium for Tooling are, though it feels like we should fix it there.
Link to impl - https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/html/forms/file_input_type.cc;l=476;drc=f4a00cc248dd2dc8ec8759fb51620d47b5114090;bpv=0;bpt=1

OrKoN (approved):

PR #1805, opened 2024-01-31
feat: support user context in permissions

OrKoN:
This PR adds `goog:` prefixed user context to allow use cases required by Puppeteer. Once permissions spec issue is resolved https://github.com/w3c/permissions/pull/438 we can remove the prefix.

PR #1873, opened 2024-02-22
test: apply comments to set_permission

OrKoN:
Raised in https://github.com/GoogleChromeLabs/chromium-bidi/pull/1805

### DIFF
--- a/tests/permissions/test_set_permission.py
+++ b/tests/permissions/test_set_permission.py
@@ -39,7 +39,7 @@ async def test_permissions_set_permission(websocket, context_id, example_url):
 @pytest.mark.asyncio
 @pytest.mark.skip(reason="See chromium-bidi/issues#1610")
 async def test_permissions_set_permission_in_user_context(
-        websocket, context_id, example_url):
+        websocket, context_id, example_url, create_context):
     await goto_url(websocket, context_id, example_url)
 
     user_context = await execute_command(websocket, {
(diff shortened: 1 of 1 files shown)


tests/permissions/test_set_permission.py
```diff
@@ -39,48 +39,47 @@ async def test_permissions_set_permission(websocket, context_id, example_url):
...
-        websocket, {
-            "method": "browsingContext.create",
-            "params": {
-                "type": "tab",
-                "userContext": user_context["userContext"]
-            }
-        })
-
+    another_browsing_context = await create_context(
+        user_context_id=user_context["userContext"])
```
sadym-chromium, line 51: I don't think we need anything except browsing context id, so something like this would be even more readable:
```suggestion
    another_browsing_context_id = (await create_context(
        user_context_id=user_context["userContext"]))['context']
```

tests/permissions/test_set_permission.py
```diff
@@ -39,48 +39,47 @@ async def test_permissions_set_permission(websocket, context_id, example_url):
...
 @pytest.mark.skip(reason="See chromium-bidi/issues#1610")
 async def test_permissions_set_permission_in_user_context(
-        websocket, context_id, example_url):
+        websocket, context_id, example_url, create_context):
     await goto_url(websocket, context_id, example_url)
 
     user_context = await execute_command(websocket, {
         "method": "browser.createUserContext",
         "params": {}
     })
```
sadym-chromium, line 48:
```suggestion
    user_context_id = (await execute_command(websocket, {
        "method": "browser.createUserContext",
        "params": {}
    }))["userContext"]
```

sadym-chromium (approved):
Consider extracting not the objects,  but the specific IDs, as mentioned in the comment

sadym-chromium (approved):
thanks!

tests/permissions/test_set_permission.py
OrKoN: Why does create_context already return an ID?

sadym-chromium (approved):

PR #1876, opened 2024-02-22
ci: shard e2e tests

sadym-chromium:
Use `pytest-shard` to shard e2e tests, decreasing e2e run time for headful Mac from 15 to 5 minutes.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -83,9 +83,21 @@ jobs:
       - name: Install python dependencies
         run: pipenv install
       - name: Run E2E tests
+        if: matrix.os == 'ubuntu-latest' && matrix.head == 'headful'
         timeout-minutes: 20
         run: >
           xvfb-run --auto-servernum
+          npm run e2e:${{ matrix.head }}
+          --
+          --this-chunk=${{ matrix.this_chunk }}
+          --total-chunks=${{ matrix.total_chunks }}
+        env:
+          BROWSER_BIN: ${{ steps.browser.outputs.executablePath }}
+          VERBOSE: ${{ github.event.inputs.verbose }}
+      - name: Run E2E tests
+        if: matrix.os != 'ubuntu-latest' || matrix.head != 'headful'
+        timeout-minutes: 20
+        run: >
           npm run e2e:${{ matrix.head }}
           --
           --this-chunk=${{ matrix.this_chunk }}
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -83,27 +83,16 @@ jobs:
       - name: Install python dependencies
         run: pipenv install
       - name: Run E2E tests
-        if: matrix.os == 'ubuntu-latest' && matrix.head == 'headful'
         timeout-minutes: 20
         run: >
           xvfb-run --auto-servernum
           npm run e2e:${{ matrix.head }}
-        # For verbose logging, set `DEBUG: 'bidi:mapper:*'` and `VERBOSE: true`.
+          --
+          --this-chunk=${{ matrix.this_chunk }}
+          --total-chunks=${{ matrix.total_chunks }}
         env:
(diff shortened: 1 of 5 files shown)


.github/workflows/e2e.yml
```diff
@@ -45,6 +45,7 @@ jobs:
         # TODO(#876): Add Windows CI.
         os: [ubuntu-latest, macos-latest]
         head: [headful, headless]
+        shard_id: [0, 1, 2]
```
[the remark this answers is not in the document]
sadym-chromium: This can be a bit confusing, as the shard_id is 0-based

Lightning00Blade:
This will make logs a little bit more tedious to look at, considering you will have to see multiple runs and files if there are failures. 
How much more complicated is it to fix the runner and run a single browser rather then spawning a new one for each test, I think that will drop it much lower than 5 min on its own.

sadym-chromium:
The scenario I think is the following:
1. E2e test failed on some shard.
2. Developer look at the GitHub action log. If it's not clear what went wrong, they download artefacts if the required shard.
3. They analyze the log. Ideally there should be the CDP logs as well.

sadym-chromium:
> This will make logs a little bit more tedious to look at, considering you will have to see multiple runs and files if there are failures. 
> How much more complicated is it to fix the runner and run a single browser rather then spawning a new one for each test, I think that will drop it much lower than 5 min on its own.

The scenario I think is the following:
1. E2e test failed on some shard.
2. Developer look at the GitHub action log. If it's not clear what went wrong, they download artefacts if the required shard.
3. They analyze the log. Ideally there should be the CDP logs as well.

OrKoN:
@sadym-chromium are all tests from one .py guaranteed to be in one shard?

sadym-chromium:
> @sadym-chromium are all tests from one .py guaranteed to be in one shard?

"Shards tests based on a hash of their test name". I read this as "no, it's not guaranteed".

OrKoN:
Okay, so at least sharding is persistent and will be the same if you run a subset of tests. Could you please update documentation on how to reproduce a specific shard failure locally?

tools/run-e2e.mjs
```diff
@@ -34,6 +34,10 @@ process.chdir(packageDirectorySync());
 const argv = parseCommandLineArgs();
 const LOG_FILE = createLogFile('e2e');
 const PYTEST_PREFIX = 'PyTest';
+
+const PYTEST_TOTAL_CHUNKS = process.env.PYTEST_TOTAL_CHUNKS || '1';
```
OrKoN, line 38: Let's make it command line arguments?

sadym-chromium:
> How much more complicated is it to fix the runner and run a single browser rather then spawning a new one for each test, I think that will drop it much lower than 5 min on its own.

The nodejs runner is used by e2e and wpt tests. In order to keep ability to run both, we need to adjust the runner for both of them. Specifically:
1. E2e relies on the fact the state is brand-new, without any events subscriptions, network interceptions etc. This would require to add proper teardown for all our tests.
2. Wpt implements teardown via WebDriver Classic. So to properly implement it, we require to implement WebDriver Classic in the nodejs runner, which is quite a lot of work for a CI-only code wrapper.

One of the options is to get rid of the nodejs runner, and run e2e via chromedriver. But that can be harder to debug, and make it way harder to test changes in the runner. E.g. adding new capabilities.

tools/run-e2e.mjs
```diff
@@ -35,6 +35,17 @@ process.chdir(packageDirectorySync());
 const argv = parseCommandLineArgs();
 const LOG_FILE = createLogFile('e2e');
 const PYTEST_PREFIX = 'PyTest';
+
+let PYTEST_TOTAL_CHUNKS = process.env.PYTEST_TOTAL_CHUNKS || '1';
```
Lightning00Blade, line 39: You can put this in the default for the Parsing of args.

Lightning00Blade (approved):

PR #1878, opened 2024-02-22
build: output third party notices

OrKoN:
Context https://bugs.chromium.org/p/chromedriver/issues/detail?id=4729


sadym-chromium (approved):

OrKoN:
Force merging because the change could not possibly affect tests.

PR #1879, opened 2024-02-22
test: increase e2e test timeout to fight the flakiness

sadym-chromium:
Out of last `12` failures of e2e in main, `5` are because of the `test_realm_dedicated_worker` timeout. This is because of the worker's realm is not destroyed immediately. Increasing timeout of the test `test_realm_dedicated_worker` should address the flakiness.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -4,6 +4,13 @@ addopts = --doctest-modules
 minversion = 7.0
 testpaths = tests
 
+# https://pypi.org/project/pytest-rerunfailures/
+# Try each test up to 2 times.
+# We prefer to configure this in PYTEST_ADDOPTS instead of pytest.ini.
+# So that this configuration is only applied in CI environments.
+# It's still listed here for documentation and discoverability purposes.
+# reruns = 1
+
 # https://pypi.org/project/pytest-timeout/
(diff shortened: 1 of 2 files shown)


OrKoN (commented):
I don't think we should increase general timeout as it might lead to the overall runtime reduction. Let's increase it only for the tests that need it with an explanation why.

sadym-chromium:
> I don't think we should increase general timeout as it might lead to the overall runtime reduction. Let's increase it only for the tests that need it with an explanation why.

It can lead to overall timeout reduction in case of tests reaching the timeout, which is not expected to happen very often.

sadym-chromium:
> For that test, is there evidence that it will succeed with the increased timeout or if it will time out anyway after 20 seconds?

It's not reproducible on my laptop, so we can use only empiric data.

sadym-chromium:
> I don't think we should increase general timeout as it might lead to the overall runtime reduction. Let's increase it only for the tests that need it with an explanation why.

Updated the timeout for the specific test.

OrKoN (approved):

OrKoN:
Landed in hopes it helps with this test timing out in other PRs

PR #1882, opened 2024-02-22
fix: add `cdp.resolveRealm` command

jrandolf:
`cdp.resolveRealm` resolves a realm ID to an executionContextId.

This is needed for Puppeteer to adopt nodes from ARIA queries.

This is a workaround for https://crbug.com/326258571

### DIFF
--- a/src/bidiMapper/domains/cdp/CdpProcessor.ts
+++ b/src/bidiMapper/domains/cdp/CdpProcessor.ts
@@ -15,21 +15,25 @@
  * limitations under the License.
  */
 
-import type {Cdp} from '../../../protocol/protocol.js';
+import {UnknownErrorException, type Cdp} from '../../../protocol/protocol.js';
 import type {CdpClient, CdpConnection} from '../../BidiMapper.js';
 import type {BrowsingContextStorage} from '../context/BrowsingContextStorage.js';
+import type {RealmStorage} from '../script/RealmStorage.js';
 
 export class CdpProcessor {
   readonly #browsingContextStorage: BrowsingContextStorage;
+  readonly #realmStorage: RealmStorage;
   readonly #cdpConnection: CdpConnection;
   readonly #browserCdpClient: CdpClient;
 
   constructor(
     browsingContextStorage: BrowsingContextStorage,
+    realmStorage: RealmStorage,
     cdpConnection: CdpConnection,
     browserCdpClient: CdpClient
   ) {
     this.#browsingContextStorage = browsingContextStorage;
+    this.#realmStorage = realmStorage;
     this.#cdpConnection = cdpConnection;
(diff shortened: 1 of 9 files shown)


OrKoN:
I wonder if instead we should accept realm in BiDi+ and automatically replace it with an execution context id. @sadym-chromium WDYT?

jrandolf:
> I wonder if instead we should accept realm in BiDi+ and automatically replace it with an execution context id. @sadym-chromium WDYT?

I think that's pretty clever. +1

src/bidiMapper/domains/cdp/CdpProcessor.ts
```diff
@@ -50,6 +56,19 @@ export class CdpProcessor {
     const client = params.session
       ? this.#cdpConnection.getCdpClient(params.session)
       : this.#browserCdpClient;
+    if (params.method === 'DOM.resolveNode') {
+      const args = params.params as {
+        executionContextUniqueId?: string;
```
OrKoN, line 61: Let's call it `bidi:realmId`?

sadym-chromium:
@jrandolf can you please elaborate a bit the original issue? Why do you need `DOM.resolveNode` CDP command in BiDi mode?

sadym-chromium:
> realm in BiDi+

If Puppeteer in BiDi needs to be aware of the CDP `executionContextId`, we can add a BiDi+ command `cdp.resolveRealm`:
```
CdpResolveRealmCommand = {
   method: "cdp.resolveRealm",
   params: CdpResolveRealmParameters,
}

CdpResolveRealmParameters = {
   bidiRealm: script.Realm,
}

CdpResolveRealmResult = {
   cdpExecutionContextId: text,
}
```

OrKoN:
@sadym-chromium it’d be nice to avoid many round trips to support a11y queries over bidi+

sadym-chromium:
> @sadym-chromium it’d be nice to avoid many round trips to support a11y queries over bidi+

In case of puppeteer + mapper, this will be an internal call, without a roundtrip to the browser. But I'm not sure I'm the specific scenario this is required to

jrandolf:
> > @sadym-chromium it’d be nice to avoid many round trips to support a11y queries over bidi+
> 
> In case of puppeteer + mapper, this will be an internal call, without a roundtrip to the browser. But I'm not sure what is the specific scenario this is required to

When an A11y query is performed in Puppeteer, we get a backendNodeId. If the query is performed in a sandboxed realm, we need to `DOM.resolveNode` against the sandboxed realm instead of the default one.

sadym-chromium (approved):
LGTM, but please add an e2e test

OrKoN (approved):

sadym-chromium (approved):
Thanks!

PR #1896, opened 2024-02-26
ci: don't run wpt headful mapper

sadym-chromium:
The node js mapper runner is not used by real users,  and is there only for development purposes. The node js runner re-runs browser for each individual test, which makes it run for a long time and causes timeouts. Having that, there is no value in running headful mapper wpt tests in CI. Expectations are kept, as manual run is still possible.


Lightning00Blade (approved):

sadym-chromium:
[Test "WPT detailed" run](https://github.com/GoogleChromeLabs/chromium-bidi/actions/runs/8050070200)

PR #1897, opened 2024-02-26
ci: better e2e logs

Lightning00Blade:
Creates a separate E2E runner file. That way we can sync the output of `bidi-server.mjs` and `pytest`, which make is easier to see where a test starts and end in the logs.

E2E runs will only print PyTest output to reduce the time need to see what fails. The message logs will be uploaded as artifacts which can be downloaded. The `VERBOSE` is set to `true` always so we get the `CDP` logs as well.

### DIFF
--- a/tools/run-e2e.mjs
+++ b/tools/run-e2e.mjs
@@ -1,0 +1,131 @@
+#!/usr/bin/env node
+
+/**
+ * Copyright 2023 Google LLC.
+ * Copyright (c) Microsoft Corporation.
+ *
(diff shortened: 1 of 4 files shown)


tools/run-e2e.mjs
```diff
@@ -0,0 +1,131 @@
+#!/usr/bin/env node
+
+/**
+ * Copyright 2023 Google LLC.
```
Lightning00Blade, line 4:
```suggestion
 * Copyright 2024 Google LLC.
```

OrKoN (approved):

PR #1898, opened 2024-02-26
ci: BiDi server logs only when test fail

Lightning00Blade:
When running `npm run e2e` we only collect `Chromium-BiDi` log when the test fails rather then all the time. 
This reduces the overall size of logs to just a few KB compare to 7-8MB  previously, also this improves readability.

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,6 +284,12 @@ npm run e2e:headful
 npm run e2e:headless
 ```
 
+This commands will run `./tools/run-e2e.mjs`, which will log the PyTest output to console,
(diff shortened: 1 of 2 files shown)


OrKoN (approved):

sadym-chromium:
Could you please add more detailed description? And consider update readme

sadym-chromium (approved):
It would be great to log CDP as well in case of failing tests, but it is obviously out of scope of the PR

Lightning00Blade:
@sadym-chromium It does log CDP logs - https://github.com/GoogleChromeLabs/chromium-bidi/blob/main/tools/bidi-server.mjs#L118

PR #1907, opened 2024-02-27
test: add test filter to e2e

Lightning00Blade:
Adds the ability to filter by folder, test file or test name.

### DIFF
--- a/tools/bidi-server.mjs
+++ b/tools/bidi-server.mjs
@@ -58,6 +58,10 @@ export function parseCommandLineArgs() {
       type: 'boolean',
       default: process.env.HEADLESS === 'true',
     })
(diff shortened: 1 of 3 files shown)


tools/bidi-server.mjs
```diff
@@ -58,6 +58,10 @@ export function parseCommandLineArgs() {
       type: 'boolean',
       default: process.env.HEADLESS === 'true',
     })
+    .option('k', {
+      describe: 'If running though e2e specifies the test name to execute',
```
sadym-chromium, line 62: Not sure if this is a typo or expected to be so

sadym-chromium (approved):
Please mention this option it in readme

PR #1915, opened 2024-02-28
feat: implement `storage.deleteCookies`

sadym-chromium:
Implement `storage.deleteCookies`.

WPT tests are on their way in https://github.com/GoogleChromeLabs/chromium-bidi/pull/1914

### DIFF
--- a/src/bidiMapper/domains/storage/StorageProcessor.ts
+++ b/src/bidiMapper/domains/storage/StorageProcessor.ts
@@ -42,6 +42,46 @@ export class StorageProcessor {
     this.#logger = logger;
   }
 
+  async deleteCookies(
+    params: Storage.DeleteCookiesParameters
+  ): Promise<Storage.DeleteCookiesResult> {
+    const partitionKey = this.#expandStoragePartitionSpec(params.partition);
+
+    const cdpResponse = await this.#browserCdpClient.sendCommand(
+      'Storage.getCookies',
+      {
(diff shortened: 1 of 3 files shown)


src/bidiMapper/domains/storage/StorageProcessor.ts
```diff
@@ -42,6 +42,46 @@ export class StorageProcessor {
...
+          c.partitionKey === partitionKey.sourceOrigin
+      )
+      .filter((cdpCookie) => {
+        const bidiCookie = cdpToBiDiCookie(cdpCookie);
+        return this.#matchCookie(bidiCookie, params.filter);
+      })
+      .map((cookie) => ({
+        ...cookie,
+        // Set expiry to pass date to delete the cookie.
+        expires: 1,
```
sadym-chromium, line 73: I believe it's a valid passed date.

OrKoN:
Sorry for editing, I thought it was Puppeteer :D

src/bidiMapper/domains/storage/StorageProcessor.ts
sadym-chromium: Another approach would be to remove all the cookies via `Storage.clearCookies` and set those didn't have to be deleted. But this would create a race condition with a concurrent read.

OrKoN (approved):
Thanks, LGTM

PR #1922, opened 2024-02-29
ci: add puppeteer label to release pr

Lightning00Blade:
This may create a loop if GitHub is not smart enough.


sadym-chromium (approved):
Let's check it out!

PR #1932, opened 2024-03-04
ci: show chromedriver headful results in wptreport site

OrKoN:
Proposal: let's switch the dashboard to report numbers from headful instead of headless as it is a configuration that matches the user experience the most?

### DIFF
--- a/.github/workflows/wpt.yml
+++ b/.github/workflows/wpt.yml
@@ -53,7 +53,7 @@ jobs:
     steps:
       - uses: actions/download-artifact@c850b930e6ba138125429b7e5c93fc707a7f8427 # v4.1.4
         with:
-          name: chromedriver-headless-artifacts
+          name: chromedriver-headful-artifacts
       - name: Prepare Pages
         run: |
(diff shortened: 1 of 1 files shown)


sadym-chromium:
> Proposal: let's switch the dashboard to report numbers from headful instead of headless as it is a configuration that matches the user experience the most?

FWIW, headless mode was chosen for report as it is used in [wpt.fyi](https://wpt.fyi/results/webdriver/tests/bidi?label=master&label=experimental&aligned).

OrKoN:
> > Proposal: let's switch the dashboard to report numbers from headful instead of headless as it is a configuration that matches the user experience the most?
> 
> FWIW, headless mode was chosen for report as it is used in [wpt.fyi](https://wpt.fyi/results/webdriver/tests/bidi?label=master&label=experimental&aligned). I don't recall where this can be checked though.

ah I see, where can I confirm that? the wpt.fyi does not show the headless label.

sadym-chromium (approved):

PR #1935, opened 2024-03-04
ci: don't commit for cancelled runs

Lightning00Blade:
https://docs.github.com/en/actions/learn-github-actions/expressions#always
Currently if you cancel a run it will commit expectations which is bad as it make the PR useless.

### DIFF
--- a/.github/workflows/wpt.yml
+++ b/.github/workflows/wpt.yml
@@ -70,7 +70,7 @@ jobs:
   wpt-auto-commit:
     name: WPT auto-commit expectations
     needs: wpt
-    if: ${{ always() && github.event.inputs.auto-commit == 'true' }}
+    if: ${{ !cancelled() && github.event.inputs.auto-commit == 'true' }}
     runs-on: ubuntu-latest
     # Give GITHUB_TOKEN write permission to commit and push.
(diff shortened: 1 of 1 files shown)


.github/workflows/wpt.yml
```diff
@@ -171,7 +171,7 @@ jobs:
           name: ${{ matrix.kind }}-${{ matrix.head }}-wpt-metadata
           path: wpt-metadata/${{ matrix.kind }}/${{ matrix.head }}
       - name: Upload artifacts
-        if: always()
+        if: success() || failure()
```
sadym-chromium, line 174: Let's still upload artifacts, it can be useful

sadym-chromium (approved):

sadym-chromium:
Let's still upload artifacts

.github/workflows/wpt.yml
sadym-chromium, line 174: Let's still upload artifacts, it can be useful even if the run is canceled
```suggestion
        if: always()
```

Lightning00Blade: Give a use case and I will revert. 
Usually a run gets cancelled because of 2 reason a superseding was created or the user manually cancelled it. I both seem like you don't need the info and its a waste of resources.

PR #1941, opened 2024-03-05
build: disable the minification

nechaev-chromium:
Chromium third party policies disallow any minified code: [adding_to_third_party.md](https://chromium.googlesource.com/chromium/src/+/HEAD/docs/adding_to_third_party.md#Binaries_obfuscated-or-minified-code).
This code disables the code minification that was introduced earlier.


sadym-chromium:
@Lightning00Blade do we still need `plugin-terser` package?

Lightning00Blade:
No we do not need it. It was only used here. 
So we should remove it from the `package.json` (both `terser` and `@rollup/plugin-terser`).

nechaev-chromium:
> No we do not need it. It was only used here. So we should remove it from the `package.json` (both `terser` and `@rollup/plugin-terser`).

Done!

Lightning00Blade (approved):
LGTM + side effect of this we should get better stack traces locally and on GitHub Actions if something fails.

PR #1943, opened 2024-03-05
[DON'T MERGE] chore: change to ESM

Lightning00Blade:


tsconfig.base.json
```diff
@@ -24,6 +24,7 @@
     "strictNullChecks": true,
     "strictPropertyInitialization": true,
     "target": "ESNext",
-    "useUnknownInCatchVariables": true
+    "useUnknownInCatchVariables": true,
+    "removeComments": true
```
OrKoN, line 28: does this keep license?

Lightning00Blade: Nope. It removes all. 
Was looking at size diff which is around 30Kb unminified ESM compared to the unminified CJS

Lightning00Blade:
Closing for now will revisit at some point.

PR #1944, opened 2024-03-05
chore: move deps and add rule

Lightning00Blade:
This should prevent our production code from pulling `devDependencies`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -216,11 +216,11 @@
     "websocket": "1.0.34",
     "wireit": "0.14.4",
     "ws": "8.16.0",
-    "yargs": "17.7.2",
-    "zod": "3.22.4"
+    "yargs": "17.7.2"
   },
   "dependencies": {
     "mitt": "3.0.1",
-    "urlpattern-polyfill": "10.0.0"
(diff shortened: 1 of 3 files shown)


OrKoN (approved):

package.json
```diff
@@ -216,11 +216,11 @@
...
     "wireit": "0.14.4",
     "ws": "8.16.0",
-    "yargs": "17.7.2",
-    "zod": "3.22.4"
+    "yargs": "17.7.2"
   },
   "dependencies": {
     "mitt": "3.0.1",
-    "urlpattern-polyfill": "10.0.0"
+    "urlpattern-polyfill": "10.0.0",
```
OrKoN, line 223: hmm, now I remember. These modules are only bundled, therefore, they were devDependencies since you do not actually need them once you install from npm.

OrKoN: back to devDependencies?

Lightning00Blade: I am not sure as in theory you can run the mapper instance and send commands that way, and that has Zod as production dependency. That's all bundled in the output folder.

PR #1945, opened 2024-03-05
feat: add support for `contexts` in `addInterception`

Lightning00Blade:
This PR move the logic to for when to enable the Fetch Domain to the CdpTarget. That way we can manage interception more gradually. 

Also fixes an issue where if we `network.addInterception` then `network.subscribe`, the interception would have not been enabled.
Also fixes an issue if a CdpTarget is created later but network interception was enable globally before that.

### DIFF
--- a/src/bidiMapper/domains/context/CdpTarget.ts
+++ b/src/bidiMapper/domains/context/CdpTarget.ts
@@ -176,10 +176,10 @@ export class CdpTarget {
 
     if (
       // Only toggle interception when Network is enabled
-      !this.#networkDomainEnabled &&
-      this.#fetchDomainStages.request === stages.request &&
-      this.#fetchDomainStages.response === stages.response &&
-      this.#fetchDomainStages.auth === stages.auth
+      !this.#networkDomainEnabled ||
+      (this.#fetchDomainStages.request === stages.request &&
+        this.#fetchDomainStages.response === stages.response &&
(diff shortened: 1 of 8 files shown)


src/bidiMapper/domains/context/CdpTarget.ts
```diff
@@ -135,9 +143,9 @@ export class CdpTarget {
           ignore: this.#acceptInsecureCerts,
         }),
         // TODO: enable Network domain for OOPiF targets.
-        enabledNetwork
-          ? this.#cdpClient.sendCommand('Network.enable')
-          : undefined,
+        this.toggleNetworkIfNeeded(enabledNetwork).then(async () => {
+          return await this.toggleFetchIfNeeded();
```
OrKoN, line 147: should not we enable the fetch at the same time as the network? I think this might be flaky since the target will unblock before the fetch is configured (if needed)

OrKoN (approved):
LGTM % 1 comment

src/bidiMapper/domains/context/CdpTarget.ts
Lightning00Blade: Agree, let me change it, also now I see there was a small bug, that we never set the `networkDomainEnabled` var in call.

Lightning00Blade: Done.

PR #1947, opened 2024-03-05
fix: workaround issue with Script.Target

Lightning00Blade:
Limitation to Zod.union is that i check the first one and cuts the object in its shape if it passes so we only get the `Realm`.
This works around this until we add a `type: 'realm'` and `type: 'context'` to the Script.Target https://github.com/w3c/webdriver-bidi/issues/635


OrKoN:
Related: https://github.com/google/cddlconv/issues/23

OrKoN (approved):

PR #1951, opened 2024-03-06
test: fix url pattern validation

OrKoN:
- Removes pattern string computation (matching is done by the URLPattern)
- Adds validation based on https://w3c.github.io/webdriver-bidi/#type-network-UrlPattern
- As a follow-up we could improve error messages
- Another follow up is to fix the remaining test cases


src/bidiMapper/domains/network/NetworkProcessor.ts
```diff
@@ -296,6 +295,40 @@ export class NetworkProcessor {
...
+          }
+          if (urlPattern.port) {
+            urlPattern.port = unescapeURLPattern(urlPattern.port);
+          }
+          if (urlPattern.pathname) {
+            urlPattern.pathname = unescapeURLPattern(urlPattern.pathname);
+            if (urlPattern.pathname[0] !== '/') {
+              urlPattern.search = `/${urlPattern.search}`;
+            }
+            if (
```
Lightning00Blade, line 315: Nit: Let's move the check that that may throw first.

src/bidiMapper/domains/network/NetworkProcessor.ts
```diff
@@ -296,6 +295,40 @@ export class NetworkProcessor {
...
+            ) {
+              throw new InvalidArgumentException('Forbidden characters');
+            }
+          }
+          if (urlPattern.search) {
+            urlPattern.search = unescapeURLPattern(urlPattern.search);
+            if (urlPattern.search[0] !== '?') {
+              urlPattern.search = `?${urlPattern.search}`;
+            }
+            if (urlPattern.search.includes('#')) {
```
Lightning00Blade, line 327: Same here.

src/bidiMapper/domains/network/NetworkProcessor.ts
OrKoN: That would not be according to the spec

src/bidiMapper/domains/network/NetworkProcessor.ts
OrKoN: Same here.

src/bidiMapper/domains/network/NetworkProcessor.ts
Lightning00Blade: Like I said nit. But if we throw anyway we should not do extra work - tldr spec is wrong.

Lightning00Blade (approved):

PR #1954, opened 2024-03-06
fix: don't block on unsubscribed events

Lightning00Blade:


src/bidiMapper/domains/session/SubscriptionManager.ts
```diff
@@ -170,8 +170,8 @@ export class SubscriptionManager {
    *
    * @returns
    */
-  isSubscribedToModule(
-    module: ChromiumBidi.BiDiModule,
+  isSubscribedTo(
```
OrKoN, line 173: Do we have unit tests for the new scenarios here? please add

OrKoN (approved):

PR #1957, opened 2024-03-06
ci: gather all WPT tests to generate report

sadym-chromium:
When generating WPT test report, get all the tests from the WPT, and update the status of those actually run. This allows to get detailed information about skipped tests and gather more detailed data on the passing rate.

If the approach is accepted, the WPT path should be passed in parameters.

Experimental run: https://github.com/GoogleChromeLabs/chromium-bidi/actions/runs/8174790184


Lightning00Blade:
@sadym-chromium The run is failing due to no `pytest` being found.

Lightning00Blade:
@sadym-chromium Run still fails due to not correct set-up.

sadym-chromium:
> @sadym-chromium Run still fails due to not correct set-up.

I know, WIP

sadym-chromium:
Closing, in favor of removing skips from the expectations at all

PR #1959, opened 2024-03-06
docs: fix script

Lightning00Blade:


sadym-chromium (approved):
this change looks like only renaming. Is it so?

Lightning00Blade:
Its a rename but if fixes it as my previous change broke it.

OrKoN (approved):

tools/wpt-report-builder/formatter.mjs
```diff
@@ -217,22 +217,21 @@ function generateHtml(map, commitHash) {
         })
 
         const expand = document.querySelector('.expand');
-        let toggleState = 'collapse';
+        let expandState = 'collapse';
```
OrKoN, line 220: any reason both of those states are not boolean?

Lightning00Blade: I use to change the name of button as well :sweat_smile:

PR #1970, opened 2024-03-07
feat: start implementing `browsingContext.locateNodes`

sadym-chromium:
Start implementing `browsingContext.locateNodes`. First part. Implement css selector.

other params are out of scope of the PR and should be implemented incrementally afterwards.


sadym-chromium:
@Lightning00Blade it's ready for review now

src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -41,6 +41,8 @@ import type {EventManager} from '../session/EventManager.js';
 import type {BrowsingContextStorage} from './BrowsingContextStorage.js';
 import type {CdpTarget} from './CdpTarget.js';
 
+import ResultOwnership = Script.ResultOwnership;
```
Lightning00Blade, line 44: This looks wrong as it uses import rather then `const` - not sure how that works.
But I am also not sure the benefit of having a shorter name in a single file gives us.

src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -1012,6 +1014,99 @@ export class BrowsingContextImpl {
 
     await this.#cdpTarget.toggleNetworkIfNeeded(enableNetwork);
   }
+
+  async locateNodes(
+    params: BrowsingContext.LocateNodesParameters
+  ): Promise<BrowsingContext.LocateNodesResult> {
+    // TODO: test sandboxing.
+    const realm = await this.getOrCreateSandbox(params.sandbox);
```
Lightning00Blade, line 1022: Nit: Let's do validation first and then any creation.

src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -1012,6 +1014,99 @@ export class BrowsingContextImpl {
...
+          params.locator.value
+        );
+      default:
+        throw new UnsupportedOperationException(
+          `locateNodes does not support ${params.locator.type} locator type.`
+        );
+    }
+  }
+
+  async #locateNodesByCssSelector(
```
Lightning00Blade, line 1059: How about we move this to the `WindowRealm` class?

src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -1012,6 +1014,99 @@ export class BrowsingContextImpl {
...
+    realm: Realm,
+    selector: string
+  ): Promise<{nodes: any[]}> {
```
Lightning00Blade, line 1062: Let's use correct type here and cast the value if needed.

src/bidiMapper/domains/context/BrowsingContextImpl.ts
sadym-chromium: It was IDE's auto import. Inlined.

src/bidiMapper/domains/context/BrowsingContextImpl.ts
sadym-chromium: good point. Done

src/bidiMapper/domains/context/BrowsingContextImpl.ts
sadym-chromium: I guess the implementation should be implemented via `DOM.querySelectorAll`, as otherwise the elements form the shadow tree will not be searched. And if so, it is more relevant to the BrowsingContext rather then window realm.

src/bidiMapper/domains/context/BrowsingContextImpl.ts
```diff
@@ -1012,6 +1012,102 @@ export class BrowsingContextImpl {
...
+        throw new UnsupportedOperationException(
+          `locateNodes does not support ${params.locator.type} locator type.`
+        );
+    }
+  }
+
+  async #locateNodesByCssSelector(
+    realm: Realm,
+    selector: string
+  ): Promise<{nodes: Script.NodeRemoteValue[]}> {
```
Lightning00Blade, line 1061:
```suggestion
  ): Promise<BrowsingContext.LocateNodesResult> {
```

Lightning00Blade (approved):
LGTM + 1 comment

PR #1982, opened 2024-03-08
docs: print delta since 2023-12-31 in filtered HTML report

mathiasbynens:
This amends https://googlechromelabs.github.io/chromium-bidi/2023 to include a note of the form

> Since 2023-12-31, the total number of subtests went from 2807 to 2807 (+0.00%), and the number of passing subtests went from 2456 to 2659 (+8.27%).

(Happy to iterate on the phrasing + the way we present this as follow-ups. E.g. a more minimal way to represent this would be to use the terms “coverage” for the former and “compliance” for the latter. As in, “coverage improved by X%, compliance by Y%”.)


tools/wpt-report-builder/filter-2023.mjs
```diff
@@ -207,5 +207,13 @@ export const apply2023Filter = (reportData) => {
   const filteredReportData = structuredClone(reportData);
   filteredReportData.results = filteredResults;
   filteredReportData.isFiltered = true;
+  // The status at 2023-12-31 gives us a baseline to compare against
+  // to track our progress during 2024.
+  filteredReportData.baseline = {
+    date: '2023-12-31',
+    passing: 2456,
+    // TODO: Why does https://wpt.fyi/results/webdriver/tests/bidi?sha=2a639a9fe4cdefd4ecd124a3f30caf631150eea5&label=master&max-count=1&q=label%3Achromium-bidi-2023 show a greater number of total subtests (2894)?
```
mathiasbynens, line 215: https://googlechromelabs.github.io/chromium-bidi/2023 is listing 2807 total subtests right now. It seems weird that the number would have gone down compared to the beginning of the year. Any idea what’s going on?

Until we figure this out, I’ve gone ahead and used the smaller value here.

Lightning00Blade: This may be due to the change in timeout multiplier, that made the `/webdriver/tests/bidi/input/perform_actions/invalid.py` timeout on CI, so we disable it that roughly ~400-450 test that are currently not being ran.

Lightning00Blade (approved):

sadym-chromium (approved):

PR #2006, opened 2024-03-14
test: add dedicated e2e runner for new and old headless modes

sadym-chromium:
Add ability to e2e tests run both with an **old** and a **new** headless. Do that on Ubuntu, and run only headful on macos. The CI combinations are:
* ubuntu, headful **4m 37s**
* ubuntu, new headless **3m 0s**
* ubuntu, old-headless **1m 56s**
* macos, headful **8m 49s**

Closes #1862

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -44,7 +44,7 @@ jobs:
       matrix:
         # TODO(#876): Add Windows CI.
         os: [ubuntu-latest, macos-latest]
-        head: [headful, headless]
+        head: [headful, 'new-headless', 'old-headless']
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout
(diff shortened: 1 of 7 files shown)


sadym-chromium:
> LGTM, but let's think if we want to run both `headful` and `new-headless` for MacOs its slow due to start-up of browser and prone to flakiness.

Updated description with an open question. WDYT?

Lightning00Blade:
Yeah I think the data validates my suspicion, I would be OK running all for `linux` as they are under 5 min. But would disable the slowest one `macOS new-headless` .

sadym-chromium:
Moving open question to comments:

Open question: should we run in CI only `new`, only `old`, or both headless modes? I'm inclined to run only `new`, with second option of running both.
* Why `new`? To detect potential issues in differences between headful and new headless implementation from the Mapper perspective.
* Why not `new`? Because new headless is very close to the headful and the chance of different behavior there is minimal.
* Why `old`? To detect issues with deprecated but still being in use old headless mode.
* Why not `old`? Because it is deprecated.

Runtimes:
* ubuntu-latest headful **4m 37s** (-26s)
* ubuntu-latest, new-headless **3m 0s**
* ubuntu-latest, old-headless **1m 56s**
* macos-latest, headful **8m 49s**
* macos-latest, new-headless **11m 49s**
* macos-latest, old-headless **6m 20s**

sadym-chromium:
https://github.com/GoogleChromeLabs/chromium-bidi/issues/2056

OrKoN:
@sadym-chromium I have a suggestion to link by mentioning each PR in the description of https://github.com/GoogleChromeLabs/chromium-bidi/issues/2056 (too many comments/notifications otherwise :) )

PR #2007, opened 2024-03-14
chore: remove headful mapper WPT expectations

sadym-chromium:
Remove the headful mapper WPT expectations, as they are not run in CI and are not updated. Keep the folder to be able to run it locally with existing scripts.


Lightning00Blade (approved):
LGTM, let's add it git ignore, so we don't commit to it.

sadym-chromium:
> LGTM, let's add it git ignore, so we don't commit to it.

done

PR #2008, opened 2024-03-14
chore: provide browser bin via capabilities or env var

sadym-chromium:
Required for running e2e via chromedriver, as chromedriver uses browser binary form capabilities.

### DIFF
--- a/tools/path-getter/path-getter.mjs
+++ b/tools/path-getter/path-getter.mjs
@@ -18,40 +18,54 @@
 import {execSync} from 'child_process';
 import {join} from 'path';
 
+import {Browser, computeSystemExecutablePath} from '@puppeteer/browsers';
+
 /**
  * Either return a value of `BROWSER_BIN` environment variable or gets the path
- * to the browser version defined in `.browser` file, downloading the required
- * version if needed.
+ * to the browser specified in `CHANNEL` environment variable if specified. If
+ * no channel or browser bin is provided in environment variable,
+ * returns defined in `.browser` file, downloading the required version if
+ * needed.
  * @return {string}
  */
 export function installAndGetChromePath() {
-  let BROWSER_BIN = process.env.BROWSER_BIN;
-  if (!BROWSER_BIN) {
-    BROWSER_BIN = execSync(
-      ['node', join('tools', 'install-browser.mjs')].join(' ')
-    )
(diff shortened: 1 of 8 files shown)


tools/path-getter/path-getter.mjs
```diff
@@ -18,40 +18,54 @@
...
+ * needed.
  * @return {string}
  */
 export function installAndGetChromePath() {
-  let BROWSER_BIN = process.env.BROWSER_BIN;
-  if (!BROWSER_BIN) {
-    BROWSER_BIN = execSync(
-      ['node', join('tools', 'install-browser.mjs')].join(' ')
-    )
+  if (process.env.BROWSER_BIN) return process.env.BROWSER_BIN;
```
Lightning00Blade, line 32:
```suggestion
  if (process.env.BROWSER_BIN) {
    return process.env.BROWSER_BIN;
  }
```

tools/path-getter/path-getter.mjs
```diff
@@ -18,40 +18,54 @@
...
  */
 export function installAndGetChromeDriverPath() {
-  let CHROMEDRIVER_BIN = process.env.CHROMEDRIVER_BIN;
-  if (!CHROMEDRIVER_BIN) {
-    CHROMEDRIVER_BIN = execSync(
-      ['node', join('tools', 'install-browser.mjs'), '--chromedriver'].join(' ')
-    )
-      .toString()
-      .trim();
+  if (process.env.CHROMEDRIVER_BIN) return process.env.CHROMEDRIVER_BIN;
```
Lightning00Blade, line 56:
```suggestion
  if (process.env.CHROMEDRIVER_BIN) { return process.env.CHROMEDRIVER_BIN;}
```

Lightning00Blade (approved):
LGTM + 2 comments.

tools/path-getter/path-getter.mjs
sadym-chromium: eslint and prettier allow this syntax

Lightning00Blade: I may be wrong but adding brackets is encouraged by Google style.

sadym-chromium:
https://github.com/GoogleChromeLabs/chromium-bidi/issues/2056

PR #2013, opened 2024-03-15
test: enable skipped tests

Lightning00Blade:
Will skip the test for `mapper` as they spawn a browser and that makes it slow

### DIFF
--- a/wpt-metadata/chromedriver/headless/webdriver/tests/bidi/input/perform_actions/invalid.py.ini
+++ b/wpt-metadata/chromedriver/headless/webdriver/tests/bidi/input/perform_actions/invalid.py.ini
@@ -1,2 +1,0 @@
-[invalid.py]
-  disabled: Issue-1903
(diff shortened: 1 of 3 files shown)


wpt-metadata/chromedriver/headless/webdriver/tests/bidi/input/perform_actions/invalid.py.ini
sadym-chromium, line 1: It's a bit ugly, but from the risks perspective I think it's ok to run the invalid tests only in one configuration. Obvious choice is the fastest, chromedriver headless.

sadym-chromium (approved):
LGTM % consider enabling only for a single config

wpt-metadata/chromedriver/headless/webdriver/tests/bidi/input/perform_actions/invalid.py.ini
Lightning00Blade: We need the `chromedriver-headful` result to be correct  - https://github.com/GoogleChromeLabs/chromium-bidi/pull/1932

PR #2019, opened 2024-03-18
[WIP] test: add chromedriver e2e to CI

sadym-chromium:
TODO:
* [ ] Fix nodejs test runner. For now it hangs.
* [ ] Skip known to fail tests
* [ ] Revisit the configurations


sadym-chromium:
I was able to reproduce the hang locally on macbook. It looks like the problem is in the nodejs wrapper `run-e2e`, as running  tests via python command works perfectly well, and via script it hangs at ~65% of tests.

sadym-chromium:
Closing for now

PR #2021, opened 2024-03-18
fix: CDP quirkiness with intercept

Lightning00Blade:


sadym-chromium (approved):
I assume #2022 will be rebased on this one

PR #2022, opened 2024-03-18
fix: get headers from Paused events

Lightning00Blade:


src/bidiMapper/domains/network/NetworkRequest.ts
```diff
@@ -514,14 +533,14 @@ export class NetworkRequest {
       this.#response.extraInfo = undefined;
     }
 
-    // TODO: get headers from Fetch.requestPaused
-    const headers = bidiNetworkHeadersFromCdpNetworkHeaders(
-      this.#response.info?.headers
-    );
+    const headers = [
+      ...bidiNetworkHeadersFromCdpNetworkHeaderEntries(
+        this.#response.paused?.responseHeaders
```
sadym-chromium, line 538: I wonder why no e2e is failing after this change

sadym-chromium (approved):

src/bidiMapper/domains/network/NetworkRequest.ts
Lightning00Blade: We never enable interception for most test. Naturally the event will not be present.

PR #2029, opened 2024-03-19
fix: get logs back

sadym-chromium:
The logs were missing because of the wrong options were sent to `child_process.spawn`


Lightning00Blade (approved):

Lightning00Blade:
Update the description to overwrite the `release-please` as this was not a user facing fix.

PR #2038, opened 2024-03-20
chore: enable clip tests

jrandolf:


Lightning00Blade:
Closing as we will need to re-investigate what the problem is.

PR #2053, opened 2024-03-21
chore: remove `iterator` and `generator` heuristics

sadym-chromium:
After the `iterator` is [removed from the spec](https://github.com/w3c/webdriver-bidi/pull/691), there is no need in this heuristics.


Lightning00Blade:
Remove the wpt change please. And link the issue of the spec.

OrKoN (approved):

sadym-chromium:
> Remove the wpt change please. And link the issue of the spec.

Updated

tests/script/test_serialization.py
```diff
@@ -187,209 +187,194 @@ async def test_serialization_deserialization(websocket, context_id,
 @pytest.mark.asyncio
 @pytest.mark.parametrize(
     "js_string, expected_serialized",
-    [
```
sadym-chromium, line 190: this was a `yapf` automatic reformat. Effectively only `iterator` tests are changed.

PR #2062, opened 2024-03-22
test: deflake e2e setup timeouts

sadym-chromium:
The flakiness is reproducible by running several (40) instances of the tests: https://github.com/GoogleChromeLabs/chromium-bidi/actions/runs/8388959559

Increasing the whole test and `session.new` timeouts fixed the repro case: https://github.com/GoogleChromeLabs/chromium-bidi/actions/runs/8389011170


Lightning00Blade:
> Avoiding timeouts in Fixtures
The timeout applies to the entire test including any fixtures which may need to be setup or torn down for the test (the exact affected fixtures depends on which scope they are and whether other tests will still use the same fixture). If the timeouts really are too short to include fixture durations, firstly make the timeouts larger ;). If this really isn’t an option a timeout_func_only boolean setting exists which can be set in the pytest ini configuration file, as documented in pytest --help.
For the decorated function, a decorator will override timeout_func_only = true in the pytest ini file to the default value. If you need to keep this option for a decorated test, you must specify the option explicitly again:

From [pytest-timeout](https://pypi.org/project/pytest-timeout/)

Should we do that as well?

sadym-chromium:
> Should we do that as well? And not increase the timeout for the test?

The spec says:
> If the timeouts really are too short to include fixture durations, firstly make the timeouts larger ;)

Lightning00Blade (approved):
LGTM, as my comment is non-blocker for this PR.

PR #2066, opened 2024-03-22
refactor: move defaults to Realm

Lightning00Blade:
Moves the defaults to the `Realm` class rather then the `ScriptProcessor`, so we can use them from anywhere.
Also some small changes that I will mark as comments.

### DIFF
--- a/src/bidiMapper/domains/network/NetworkProcessor.ts
+++ b/src/bidiMapper/domains/network/NetworkProcessor.ts
@@ -22,7 +22,6 @@ import {
   NoSuchRequestException,
   InvalidArgumentException,
 } from '../../../protocol/protocol.js';
-import {assert} from '../../../utils/assert.js';
 import type {BrowsingContextStorage} from '../context/BrowsingContextStorage.js';
 
 import type {NetworkRequest} from './NetworkRequest.js';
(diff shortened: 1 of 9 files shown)


src/bidiMapper/domains/network/NetworkProcessor.ts
```diff
@@ -160,12 +157,6 @@ export class NetworkProcessor {
 
       username = credentials.username;
       password = credentials.password;
-      // TODO: This should be invalid argument exception.
-      // Spec may need to be updated.
-      assert(
```
Lightning00Blade, line 165: This should be verified by `zod` and TS tell me it's always true.

src/bidiMapper/domains/script/ScriptProcessor.ts
```diff
@@ -78,23 +78,17 @@ export class ScriptProcessor {
   async removePreloadScript(
     params: Script.RemovePreloadScriptParameters
   ): Promise<EmptyResult> {
-    const bidiId = params.script;
+    const {script: id} = params;
```
Lightning00Blade, line 81: Simplify name.

src/bidiMapper/domains/script/ScriptProcessor.ts
```diff
@@ -105,14 +99,12 @@ export class ScriptProcessor {
     const realm = await this.#getRealm(params.target);
     return await realm.callFunction(
```
Lightning00Blade, line 100: We should consider just passing in an object easier to use the default.

PR #2070, opened 2024-03-25
feat: implement readonly capabilities

OrKoN:
Note that e2e or chromedriver tests would not work here since they override session.new call. This PR does not implement capability matching since this is not required by Puppeteer yet.

### DIFF
--- a/src/bidiMapper/domains/session/SessionProcessor.ts
+++ b/src/bidiMapper/domains/session/SessionProcessor.ts
@@ -15,6 +15,7 @@
  * limitations under the License.
  */
 
+import type {CdpClient} from '../../../cdp/CdpClient.js';
 import type {BidiPlusChannel} from '../../../protocol/chromium-bidi.js';
 import type {
   ChromiumBidi,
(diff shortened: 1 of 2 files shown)


OrKoN:
Note that e2e or chromedriver tests would not work here since they override session.new call

src/bidiMapper/domains/session/SessionProcessor.ts
```diff
@@ -26,15 +27,37 @@ import type {EventManager} from './EventManager.js';
...
+  async create(params: Session.NewParameters): Promise<Session.NewResult> {
+    // Since mapper exists, there is a session already.
+    // Still the mapper can handle capabilities for us.
+    // Currently, only Puppeteer calls here but, eventually, every client
+    // should delegrate capability processing here.
+    const version = await this.#browserCdpClient.sendCommand('Browser.getVersion');
+    return {
+      sessionId: 'unknown',
+      capabilities: {
+        acceptInsecureCerts: false,
```
Lightning00Blade, line 50: Can we get this value from the options?

Lightning00Blade (approved):

src/bidiMapper/domains/session/SessionProcessor.ts
OrKoN: Not really, since we would need to implement the capability matching algorithm. I will look into that later since it is not needed for Puppeteer.

Issue #2035, opened 2024-03-20
Productionisation of Network Interception


Lightning00Blade:
Tracking bug for thing left to do to achieve 100% WPT test coverage, reduce complexity, and reduce flakiness. 
Design document - [go/webdriver:network-interception](http://goto.google.com/webdriver:network-interception)

- [x] Simplify logic for Network module (domain) by having a single point of entry (#1871)
- [x] Remove blocking behavior for Network events (#1871) 
- [x] Enable Fetch and Network events when needed (#1785, #1945)
- [x] Support for `contexts` (Puppeteer requirement) (#1945)
- [x] Investigate cost related to interceptions #1080
- [x] Emit `authRequired` events (#1871)
- [x] `network.continueResponse` work for AuthRequired events
- [x] `network.provideResponse` supports all paused phases
- [x] Provide `headers` from all possible requests (#2022)
- [x] Clarify network interception edge case https://github.com/w3c/webdriver-bidi/issues/663
- [ ] Fix remaining network interception edge cases https://github.com/GoogleChromeLabs/chromium-bidi/issues/2147

[Full list of PRs](https://github.com/GoogleChromeLabs/chromium-bidi/pulls?q=is%3Apr+author%3ALightning00Blade+is%3Aclosed+in%3Adescription+%23network-interception-2024)

Lightning00Blade:
Closing will track on edge case on it's own issue.

PR #2081, opened 2024-03-27
fix: don't block data url events when interception is enabled

Lightning00Blade:
#2035

### DIFF
--- a/src/bidiMapper/domains/network/NetworkStorage.spec.ts
+++ b/src/bidiMapper/domains/network/NetworkStorage.spec.ts
@@ -216,6 +216,20 @@ describe('NetworkStorage', () => {
       const event = await getEvent('network.beforeRequestSent');
       expect(event).to.exist;
     });
+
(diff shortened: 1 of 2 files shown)


OrKoN (approved):

Lightning00Blade:
This should not be needed I miss read some Puppereer logs. Will close once I verify.

src/bidiMapper/domains/network/NetworkStorage.spec.ts
```diff
@@ -216,6 +216,20 @@ describe('NetworkStorage', () => {
       const event = await getEvent('network.beforeRequestSent');
       expect(event).to.exist;
     });
+
+    it.only('should work with data url and global interception', async () => {
```
OrKoN, line 220:
```suggestion
    it('should work with data url and global interception', async () => {
```

OrKoN (approved):

PR #2084, opened 2024-03-27
chore: rename folder from `domains` to `modules`

Lightning00Blade:
`domains` is the CDP separation but BiDi uses `modules` give this and the fact we split them based on the BiDi implementation we should rename if for consistency.


OrKoN:
Would you mind delaying renaming this until we have no in-flight PRs including domains?

OrKoN:
You would also need to update all imports :)

Lightning00Blade:
Yeah, but I forgot to push the second commit, having the rename in one and the change in another make it (more likely) the file to keep its history.
What is the PR that are inflight? That we think we will merge in the coming days?

OrKoN:
The bluetooth stuff, it will probably take longer.

Lightning00Blade:
Happy to rebase it myself after this lands.

OrKoN (approved):

PR #2091, opened 2024-03-28
test: use `--headless=new` for Mapper runs as well

Lightning00Blade:
Tests are expected to Timeout, due to the increase time to spawn the browser.
But we should get closer to ChromeDriver. This means that we should be able to reproduce all Mapper failures, if the not it signals that there may be an issue with ChromeDriver. 
Closes https://github.com/GoogleChromeLabs/chromium-bidi/issues/949


OrKoN (approved):

Lightning00Blade:
Closing this as this still is not possible without disabling the only mapper runs.

PR #2134, opened 2024-04-17
fix: wait for the first pain of the page beofre taking screenshot

sadym-chromium:
When taking screenshot right after the navigation the viewport can be not yet resized, which makes the result of `window.visualViewport` incorrect. Waiting for the `firstPaint` lifecycle event should prevent this race condition.

Probably this is related to this spec of [`browsingContext.captureScreenshot`](https://w3c.github.io/webdriver-bidi/#command-browsingContext-captureScreenshot):
> 4. Immediately after the next invocation of the [run the animation frame callbacks](https://html.spec.whatwg.org/multipage/imagebitmap-and-animations.html#run-the-animation-frame-callbacks) algorithm for document...

Alternatively, the `wait: complete` condition can be adjusted, but this can not work in background not-active pages.

This fixes WPT tests `test_clip_box_outside_of_window_viewport[viewport]` and `test_clip_element_outside_of_window_viewport[viewport]`.

Example CDP logs BEFORE the change (simplified):
```
> window.innerHeight: 469

cdp:RECV ◂ { "method": "Page.frameResized", "params": {} }
cdp:RECV ◂ { "method": "Page.lifecycleEvent", "params": { "name": "firstPaint" } }
cdp:RECV ◂ { "method": "Page.lifecycleEvent", "params": { "name": "firstMeaningfulPaintCandidate" } }

> window.innerHeight: 417
```

Open question: should it be `firstMeaningfulPaintCandidate` instead?

### DIFF
--- a/src/bidiMapper/modules/context/BrowsingContextImpl.ts
+++ b/src/bidiMapper/modules/context/BrowsingContextImpl.ts
@@ -61,6 +61,8 @@ export class BrowsingContextImpl {
 
   #lifecycle = {
     DOMContentLoaded: new Deferred<void>(),
+    // Needed for the screenshot capturing.
+    firstPaint: new Deferred<void>(),
     load: new Deferred<void>(),
   };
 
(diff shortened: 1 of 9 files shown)


sadym-chromium:
The `firstPaint` event is not guaranteed to be emitted. The flakiness comes from the resize caused by the previous screenshot.

OrKoN:
"wait for the first pain" 🤣

PR #2148, opened 2024-04-22
feat: implement accessibility locator

OrKoN:
This PR implements https://w3c.github.io/webdriver-bidi/#locate-nodes-using-accessibility-attributes


src/bidiMapper/modules/context/BrowsingContextImpl.ts
```diff
@@ -1226,10 +1227,105 @@ export class BrowsingContextImpl {
...
+
+                  if (match) {
+                    if (
+                      maxNodeCount !== 0 &&
+                      returnedNodes.length === maxNodeCount
+                    ) {
+                      break;
+                    }
+
+                    returnedNodes.push(contextNode);
```
sadym-chromium, line 1293: I guess we don't need to make another loop if the `maxNodeCount` is reached.
```suggestion
                    returnedNodes.push(contextNode);

                                        if (
                      maxNodeCount !== 0 &&
                      returnedNodes.length === maxNodeCount
                    ) {
                      break;
                    }
```

OrKoN: I agree, but note that this is not per spec

OrKoN: wait, it's actually wrong, we need to check before we append.

sadym-chromium: oh, right. So you have to `return` instead of `break`. Otherwise it will have lots of unnecessary DOM traversing.

OrKoN: Return would not be sufficient too, because it is in the recursive function

OrKoN: added an abort flag

sadym-chromium (approved):

PR #2158, opened 2024-04-24
ci: remove second check

Lightning00Blade:
We already checked for that the first `if` for the Job.

### DIFF
--- a/.github/workflows/wpt.yml
+++ b/.github/workflows/wpt.yml
@@ -117,7 +117,7 @@ jobs:
   wpt:
     name: ${{ matrix.kind }}-${{ matrix.head }}
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/wpt.yml
```diff
@@ -86,10 +86,6 @@ jobs:
     needs: wpt
     if: ${{ !cancelled() && (github.event.inputs.auto-commit == 'true' || contains(github.event.pull_request.labels.*.name, 'update-expectations')) }}
     runs-on: ubuntu-latest
-    # Give GITHUB_TOKEN write permission to commit and push.
```
Lightning00Blade, line 89: No-Op - we are using our own token.

OrKoN (approved):

PR #2189, opened 2024-05-03
[Experiment] do not throttle Mapper tab in WPT

sadym-chromium:
Check if the flag influences the WPT performance.


sadym-chromium:
9:45 vs 10:30, 7% faster

PR #2202, opened 2024-05-13
chore: update WPT expectations quickly

sadym-chromium:
* `wpt-quick` now updates the expectations by label or a flag if run manually.
* `wpt` still updates expectations if run manually.
* `tools/run-wpt.mjs` supports the wpt report path as a CLI argument. This is required for running update expectations job in a `find` loop.

The update expectations job now is done in 8 minutes instead of 22.

Job without updating expectations: https://github.com/GoogleChromeLabs/chromium-bidi/actions/runs/9080607825
Job with updating expectations: https://github.com/GoogleChromeLabs/chromium-bidi/actions/runs/9080522732

### DIFF
--- a/tools/run-wpt.mjs
+++ b/tools/run-wpt.mjs
@@ -51,7 +51,9 @@ function usage() {
       [VERBOSE==<true | default: false>]
       [WPT_REPORT=<default: 'wptreport.json'>]
       [WPT_METADATA=<default: 'wpt-metadata/$\{ 'chromedriver' | 'mapper' }/$\{ 'headless' | 'headful' }>]
-      ${process.argv[1]} [webdriver/tests/bidi/[...]]`
+      ${process.argv[1]}
+      [--wpt-report WPT_REPORT]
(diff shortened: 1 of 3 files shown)


tools/run-wpt.mjs
```diff
@@ -62,17 +64,25 @@ if (
   usage();
   process.exit(0);
 }
+const restArgs = process.argv.slice(2);
+
+// Weather to run the tests. `false` value is useful for updating expectations
+// based on the WPT report.
+const RUN_TESTS = process.env.RUN_TESTS || 'true';
```
OrKoN, line 71: let's convert env variables into predictable booleans instead of strings, e.g. `(process.env.RUN_TESTS ?? 'true') === 'true'`

OrKoN (approved):

tools/run-wpt.mjs
sadym-chromium: This requires updating all the variables to prevent confusion, so I'd keep it out of scope of the PR

PR #2207, opened 2024-05-15
feat: support for graphemes in key input

sadym-chromium:
Allow graphemes in key input according to step 6 of https://w3c.github.io/webdriver/#dfn-process-a-key-action.
Use [`Intl.Segmenter`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter) to detect grapheme borders.

Follow up: update WPT tests.

### DIFF
--- a/src/bidiMapper/modules/input/ActionDispatcher.ts
+++ b/src/bidiMapper/modules/input/ActionDispatcher.ts
@@ -570,7 +570,7 @@ export class ActionDispatcher {
     if (!isSingleGrapheme(rawKey)) {
       // https://w3c.github.io/webdriver/#dfn-process-a-key-action
       // WebDriver spec allows a grapheme to be used.
-      throw new InvalidArgumentException(`Invalid key value: ${action.value}`);
+      throw new InvalidArgumentException(`Invalid key value: ${rawKey}`);
     }
     const isGrapheme = isSingleComplexGrapheme(rawKey);
     const key = getNormalizedKey(rawKey);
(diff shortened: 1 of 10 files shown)


src/bidiMapper/modules/input/ActionDispatcher.ts
```diff
@@ -878,3 +902,8 @@ function getRadii(
     radiusY: height ? height / 2 : 0.5,
   };
 }
+
+function splitGraphemes(value: string) {
```
OrKoN, line 906: let's add some unit tests for the grapheme detection and split

src/bidiMapper/modules/input/ActionDispatcher.ts
```diff
@@ -878,3 +902,8 @@ function getRadii(
...
+  const segmenter = new Intl.Segmenter(undefined, {granularity: 'grapheme'});
```
OrKoN, line 907: Is it dependent on the current locale? would not it lead to incorrect results and differences between node vs browser?

src/bidiMapper/modules/input/ActionDispatcher.ts
sadym-chromium: good point. Added.

src/bidiMapper/modules/input/ActionDispatcher.ts
sadym-chromium: According to https://www.unicode.org/reports/tr29/#Grapheme_Cluster_Boundaries, there can be some edge-cases in different locales. So I stick to `en` for consistency.

OrKoN: I see that undefined is passed to Intl.Segmenter and the current locale is not guaranteed to be en_US. Can you explicitly set it here?

src/bidiMapper/modules/input/ActionDispatcher.ts
OrKoN: Not seeing the tests yet

src/bidiMapper/modules/input/ActionDispatcher.ts
sadym-chromium: haven't pushed yet

OrKoN (approved):

src/bidiMapper/modules/input/ActionDispatcher.ts
sadym-chromium: pushed now

PR #2227, opened 2024-05-22
chore: let emitted value be inferred

Lightning00Blade:
Only remove types that can be inferred.

### DIFF
--- a/src/bidiMapper/modules/context/BrowsingContextImpl.ts
+++ b/src/bidiMapper/modules/context/BrowsingContextImpl.ts
@@ -322,142 +322,129 @@ export class BrowsingContextImpl {
   }
 
   #initListeners() {
-    this.#cdpTarget.cdpClient.on(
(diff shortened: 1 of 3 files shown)


src/bidiMapper/modules/context/BrowsingContextImpl.ts
sadym-chromium, line 1: It looks like this file was only reformatted. Could you please split the reformat and the functional change to ease review?

Lightning00Blade: There are no functional changes.

sadym-chromium (approved):

src/bidiMapper/modules/context/BrowsingContextImpl.ts
sadym-chromium: That explains why I couldn't see them )

Issue #2133, opened 2024-04-17
CI should run `interop` test to prevent regressions on our end 


Lightning00Blade:
Currently we run only test that are in the `webdriver/tests/bidi` we should have a separate job for running the `webdriver/tests/interop` . We should also have a separate view on the Web site.

- [x] Run `webdriver/tests/interop` in CI as separate job.
- [x] Have a page that shows `webdriver/tests/interop` tests

whimboo:
Just another idea... maybe you could run all the tests under `webdriver/tests` and exclude `classic`?

Lightning00Blade:
Done by #2350

PR #2237, opened 2024-05-24
ci: add Interop test to `main` CI

Lightning00Blade:
This add the changes only to the CI we run on the `main` branch a follow up will add it to `wpt-quick`.
New Website image:
image

Issue: #2133

### DIFF
--- a/tools/wpt-report-builder/builder.mjs
+++ b/tools/wpt-report-builder/builder.mjs
@@ -61,19 +61,36 @@ function readReport(filePath) {
 }
 
 const jsonPath = process.argv[2];
-const outputPath = process.argv[3];
-const filteredOutputPath = process.argv[4];
+const jsonInteropPath = process.argv[3];
+const outputPath = process.argv[4];
+const filteredOutputPath = process.argv[6];
 const reportData = readReport(jsonPath);
+const reportInteropData = readReport(jsonInteropPath);
 const filteredReportData = apply2023Filter(reportData);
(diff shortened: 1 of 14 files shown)


tools/wpt-report-builder/builder.mjs
```diff
@@ -61,19 +61,36 @@ function readReport(filePath) {
 }
 
 const jsonPath = process.argv[2];
-const outputPath = process.argv[3];
-const filteredOutputPath = process.argv[4];
+const jsonInteropPath = process.argv[3];
```
sadym-chromium, line 64: Optional:
Would it be possible to make the script API more generic, like this: `node tools/wpt-report-builder/builder.mjs --report-path {outputPath} --filter-2023 {true | false} {list-of-json-reports}`? The important part is providing several report paths. This would allow generating report in sharded WPT run.

sadym-chromium (approved):
LGTM, but it would be very helpful for future to make the builder.mjs a bit more generic

tools/wpt-report-builder/builder.mjs
Lightning00Blade: Can you open an issue about this, this would require some refactoring, and most likely using `yargs` to handle the argument parts.

sadym-chromium: sure: https://github.com/GoogleChromeLabs/chromium-bidi/issues/2240

PR #2254, opened 2024-05-29
refactor: use same error as CDP

Lightning00Blade:


src/bidiMapper/modules/network/NetworkProcessor.ts
```diff
@@ -461,6 +458,14 @@ export class NetworkProcessor {
       }
     });
   }
+
+  static handleInterceptionError(error: any) {
```
sadym-chromium, line 462: The naming is a bit confusing, as `handle` implies it is not re-thrown. Perhaps `wrapInterceptionErrorToBiDi`? Or return an error from this method, and re-throw it when used.

Lightning00Blade: Update. I think `ToBiDi` is redundant and not fully true as in case of other error we don't wrap it.

src/bidiMapper/modules/network/NetworkProcessor.ts
```diff
@@ -106,11 +106,7 @@ export class NetworkProcessor {
         postData: getCdpBodyFromBiDiBytesValue(body),
       });
     } catch (error) {
-      // https://source.chromium.org/chromium/chromium/src/+/main:content/browser/devtools/protocol/fetch_handler.cc;l=169
-      if ((error as any)?.message.includes('Invalid header')) {
-        throw new InvalidArgumentException('Tried setting invalid header');
-      }
-      throw error;
+      NetworkProcessor.wrapInterceptionError(error);
```
sadym-chromium, line 109: I would still be in favor of the helper function just wrapping exception, but throwing it from this place, not the wrapper function. Pure for readability.

sadym-chromium (approved):
Please consider throwing exception in the `catch` block, not in the helper function.

src/bidiMapper/modules/network/NetworkProcessor.ts
sadym-chromium, line 109:
```suggestion
      throw NetworkProcessor.wrapInterceptionError(error);
```

src/bidiMapper/modules/network/NetworkProcessor.ts
```diff
@@ -273,12 +273,9 @@ export class NetworkProcessor {
         body: getCdpBodyFromBiDiBytesValue(body),
       });
     } catch (error) {
-      // https://source.chromium.org/chromium/chromium/src/+/main:content/browser/devtools/protocol/fetch_handler.cc;l=169
-      if ((error as any)?.message.includes('Invalid header')) {
-        throw new InvalidArgumentException('Tried setting invalid header');
-      }
-      throw error;
+      NetworkProcessor.wrapInterceptionError(error);
```
sadym-chromium, line 276:
```suggestion
      throw NetworkProcessor.wrapInterceptionError(error);
```

Issue #2282, opened 2024-06-05
Implement reconnection


sadym-chromium:
Support launching Mapper on already existing browser instance.

Currently this feature is only supported when run via Puppeteer. Not chromedriver nor nodejs runner support re-connecting yet.

This includes:
* [x] Existing browsing contexts
  * [x] Top-level
  * [x] Nested
    * [x] Same-process
    * [x] OOPiF
* [x] Existing script realms
  * [x] Main realm
  * [x] Sandboxes

PR #2289, opened 2024-06-06
feat: restore frame tree when reconnecting to browser

sadym-chromium:
Addressing https://github.com/GoogleChromeLabs/chromium-bidi/issues/2282

### DIFF
--- a/src/bidiMapper/modules/cdp/CdpTarget.ts
+++ b/src/bidiMapper/modules/cdp/CdpTarget.ts
@@ -22,7 +22,9 @@ import {BiDiModule} from '../../../protocol/chromium-bidi.js';
 import type {ChromiumBidi} from '../../../protocol/protocol.js';
 import {Deferred} from '../../../utils/Deferred.js';
 import type {LoggerFn} from '../../../utils/log.js';
+import {LogType} from '../../../utils/log.js';
 import type {Result} from '../../../utils/result.js';
+import {BrowsingContextImpl} from '../context/BrowsingContextImpl.js';
 import type {BrowsingContextStorage} from '../context/BrowsingContextStorage.js';
 import {LogManager} from '../log/LogManager.js';
 import type {NetworkStorage} from '../network/NetworkStorage.js';
@@ -35,6 +37,7 @@ export class CdpTarget {
   readonly #id: Protocol.Target.TargetID;
   readonly #cdpClient: CdpClient;
   readonly #browserCdpClient: CdpClient;
+  readonly #realmStorage: RealmStorage;
   readonly #eventManager: EventManager;
 
   readonly #preloadScriptStorage: PreloadScriptStorage;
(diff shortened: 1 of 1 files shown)


src/bidiMapper/modules/cdp/CdpTarget.ts
```diff
@@ -133,9 +143,16 @@ export class CdpTarget {
    */
   async #unblock() {
     try {
+      await this.#cdpClient.sendCommand('Page.enable');
```
OrKoN, line 146: this is very likely not correct, you need to schedule the commands at the same time, as they might not reach the renderer while they are blocked, resulting in a deadlock

sadym-chromium: right, fixed

src/bidiMapper/modules/cdp/CdpTarget.ts
```diff
@@ -134,8 +145,17 @@ export class CdpTarget {
   async #unblock() {
     try {
       await Promise.all([
+        (async () => {
+          await this.#cdpClient.sendCommand('Page.enable');
+          // There can be some existing frames in the target, if reconnecting to an
+          // existing browser instance, e.g. via Puppeteer. Need to restore the browsing
+          // contexts for the frames to correctly handle further events, like
+          // `Runtime.executionContextCreated`.
+          const frameTree =
```
OrKoN, line 154: this is also incorrect: getFrameTree will only be sent once Page.enable is resolved, which is too late

OrKoN:
@sadym-chromium this should probably work https://github.com/GoogleChromeLabs/chromium-bidi/pull/2290

src/bidiMapper/modules/cdp/CdpTarget.ts
sadym-chromium: ack

src/bidiMapper/modules/cdp/CdpTarget.ts
```diff
@@ -134,8 +144,20 @@ export class CdpTarget {
...
         this.#cdpClient.sendCommand('Page.enable'),
+        // There can be some existing frames in the target, if reconnecting to an
+        // existing browser instance, e.g. via Puppeteer. Need to restore the browsing
+        // contexts for the frames to correctly handle further events, like
+        // `Runtime.executionContextCreated`.
+        // It's important to schedule this task together with enabling domains commands to
+        // prepare the tree before the events (e.g. Runtime.executionContextCreated) start
+        // coming.
+        this.#cdpClient
+          .sendCommand('Page.getFrameTree')
```
OrKoN, line 156: perhaps we could add a test to make sure that evaluate works right after a context creation (the puppeteer job does not report that scenario as fixed yet)

OrKoN:
Let's link issues that this contributes to/closes

src/bidiMapper/modules/cdp/CdpTarget.ts
sadym-chromium: The puppeteer tests started passing:
> **Remove** the following from the TestExpectations.json to ignore the error:
> * [launcher.spec] Launcher specs Puppeteer Puppeteer.connect should be able to reconnect to a disconnected browser"
> * [launcher.spec] Launcher specs Puppeteer Puppeteer.connect should be able to reconnect"

OrKoN (approved):

PR #2290, opened 2024-06-06
test: restore frame tree

OrKoN:


src/bidiMapper/modules/cdp/CdpTarget.ts
```diff
@@ -169,6 +186,36 @@ export class CdpTarget {
...
+        parentBrowsingContext.userContext,
+        parentBrowsingContext.cdpTarget,
+        this.#eventManager,
+        this.#browsingContextStorage,
+        this.#realmStorage,
+        // At this point, we don't know the URL of the frame yet, so it will be updated
+        // later.
+        frame.url,
+        this.#logger,
+        false
```
sadym-chromium, line 211: I believe we still have to emit `browsingContext.contextCreated` if the target is attached with frames.

OrKoN: I think that should be only on subscribe or when a new context actually created? anyway the tests didn't look happy with extra events

Issue #2301, opened 2024-06-10
Network requests are retained for longer than needed 


OrKoN:
See https://github.com/mozilla/pdf.js/issues/17961#issuecomment-2158609129

It looks like the network requests and related data are retained even after all events have been emitted and the request should be freed. This issue is discovered in the pdf.js project and the linked issue contains steps to reproduce.

@sadym-chromium @Lightning00Blade

PR #2302, opened 2024-06-10
fix: remove requests after they have been completed

OrKoN:
Refs #2301


OrKoN:
@Lightning00Blade wdyt? it looks like some e2e tests expect requests to linger in storage forever. What is the best way to fix this?

OrKoN:
I think we can update expected error messages.

PR #2320, opened 2024-06-14
fix: don't use promises for events

Lightning00Blade:
It becomes very hard to track promises when we use `.then`.
I would like to remove this and maybe try to create an EsLint rule that makes sure that `await` in `.on`  are always wrapped in `try/catch`.

### DIFF
--- a/src/bidiMapper/modules/log/LogManager.ts
+++ b/src/bidiMapper/modules/log/LogManager.ts
@@ -91,7 +91,7 @@ export class LogManager {
   }
 
   #initializeEntryAddedEventListener() {
-    this.#cdpTarget.cdpClient.on('Runtime.consoleAPICalled', (params) => {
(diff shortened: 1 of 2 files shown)


src/bidiMapper/modules/log/LogManager.ts
```diff
@@ -104,23 +104,15 @@ export class LogManager {
...
-            kind: 'success',
-            value: {
+      try {
+        const args = await Promise.all(
+          params.args.map((arg) => {
+            return realm.serializeCdpObject(arg, Script.ResultOwnership.None);
+          })
+        );
+        for (const browsingContext of realm.associatedBrowsingContexts) {
+          this.#eventManager.registerEvent(
```
OrKoN, line 114: I think this will change the order of events

Lightning00Blade: That's correct...

PR #2332, opened 2024-06-20
chore: get dependency revision from npm registry

sadym-chromium:
Add dependency revision in `lib/THIRD_PARTY_NOTICES` from npm registry. Helps addressing https://crbug.com/347863537

Tested manually by running `THIRD_PARTY_NOTICES_REVISION=true npm run build`.

### DIFF
--- a/rollup.config.mjs
+++ b/rollup.config.mjs
@@ -14,6 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import {spawnSync} from 'child_process';
 import path from 'path';
 
 import commonjs from '@rollup/plugin-commonjs';
@@ -52,6 +53,7 @@ export default {
               arr.push(`URL: ${url ?? 'N/A'}`);
               arr.push(`Version: ${dependency.version ?? 'N/A'}`);
               arr.push(`License: ${dependency.license ?? 'N/A'}`);
+              arr.push(`Revision: ${getRevision(dependency)}`);
               if (dependency.licenseText !== null) {
                 arr.push('');
                 arr.push(dependency.licenseText.replaceAll('\r', ''));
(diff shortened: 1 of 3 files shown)


rollup.config.mjs
```diff
@@ -73,3 +75,27 @@ export default {
...
+ */
+function getRevision(dependency) {
+  if (process.env.THIRD_PARTY_NOTICES_REVISION !== 'true') return 'N/A';
+  try {
+    // Chromium requires the revision to be specified for any dependency.
+    // Even though the revision is not available in the local npm package info,
+    // it can be fetched from the registry.
+    const npmRegistryUrl = `https://registry.npmjs.org/${dependency.name}/${dependency.version}`;
+    const dependencyRegistryInfo = JSON.parse(
+      execSync(`curl -s ${npmRegistryUrl}`).toString()
```
OrKoN, line 93: this is not cross-platform. Also, I am not sure if the THIRD_PARTY_NOTICES_REVISION is the right file for it?

OrKoN, line 93: let's use spawnSync with args so that there is no risk that package name is interpreted by the shell

sadym-chromium: > this is not cross-platform. 

This script is intended to be used only when bumping chromedriver manually, so I'm not sure if being cross-platform is really required. I'm not aware of the cross-platform tool for fetching url. As long as we require a sync way to fetch it (`rollup-plugin-license` does not support async),we can use package like [`sync-request`](https://www.npmjs.com/package/sync-request). WDYT?

> Also, I am not sure if the THIRD_PARTY_NOTICES_REVISION is the right file for it?

`THIRD_PARTY_NOTICES` seems to be a good place for it. Other suggestions?

sadym-chromium: NP, but this depends on the previous discussion.

OrKoN: alternative proposal is to download dependency revisions to a file ahead of the rollup call and read in the template. Not blocking though but if someone needs to roll being on Windows it might require extra steps to install curl probably.

OrKoN: another alternative, is to post-process the file with adding revisions.

OrKoN: The linked bug mentions that it expects the revision in README.chromium, is the README.chromium updated based on the THIRD_PARTY_NOTICES?

sadym-chromium: > The linked bug mentions that it expects the revision in README.chromium, is the README.chromium updated based on the THIRD_PARTY_NOTICES?

there is another script building README.chromium based on `THIRD_PARTY_NOTICES`. Let me experiment if I can add this logic in that script.

rollup.config.mjs
```diff
@@ -19,6 +19,7 @@ import path from 'path';
 import commonjs from '@rollup/plugin-commonjs';
 import {nodeResolve} from '@rollup/plugin-node-resolve';
 import license from 'rollup-plugin-license';
+import request from 'sync-request';
```
OrKoN, line 22: tbh I would prefer curl over sync-request

OrKoN (approved):

rollup.config.mjs
OrKoN: ah I see, if that does not work out, so current approach LGTM

sadym-chromium:
Addressed in chromium code base: https://crrev.com/c/5644823

PR #2338, opened 2024-06-24
refactor: move the methods to request

Lightning00Blade:
The main justification is we need to store the data from the interception as BiDi sends back the data for both request and response with the events. 

This PR does not introduce functional changes, Or are marked by comment.

### DIFF
--- a/src/bidiMapper/modules/network/NetworkRequest.ts
+++ b/src/bidiMapper/modules/network/NetworkRequest.ts
@@ -475,10 +475,6 @@ export class NetworkRequest {
       }
     }
 
-    if (this.interceptPhase !== Network.InterceptPhase.ResponseStarted) {
-      return;
-    }
-
     const responseHeaders: Protocol.Fetch.HeaderEntry[] | undefined =
       cdpFetchHeadersFromBidiNetworkHeaders(overrides.headers);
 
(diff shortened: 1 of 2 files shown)


src/bidiMapper/modules/network/NetworkRequest.ts
```diff
@@ -411,19 +419,25 @@ export class NetworkRequest {
...
       method: overrides.method,
-      headers: overrides.headers,
-      postData: overrides.postData,
+      headers,
+      postData,
     });
+
     // TODO: Store postData's size only
-    this.#request.overrides = {
+    this.#requestOverrides = {
```
Lightning00Blade, line 440: We store the overrides as they come from BiDi.

src/bidiMapper/modules/network/NetworkRequest.ts
```diff
@@ -630,12 +666,8 @@ export class NetworkRequest {
...
 
     let headers: Network.Header[] = [];
-    if (this.#request.overrides?.headers) {
-      headers = [
-        ...bidiNetworkHeadersFromCdpNetworkHeadersEntries(
-          this.#request.overrides?.headers
-        ),
-      ];
+    if (this.#requestOverrides?.headers) {
+      headers = this.#requestOverrides.headers;
```
Lightning00Blade, line 670: We just use them without re parsing them.

sadym-chromium (approved):

PR #2339, opened 2024-06-24
fix: request body size

Lightning00Blade:


sadym-chromium (approved):

sadym-chromium:
you probably want to add the flag `update-expectations`

sadym-chromium:
I believe yout force-push dropped the WPT expectations update commit

PR #2348, opened 2024-06-25
feat: body size for non-intercepted requests

Lightning00Blade:
I couldn't get an example with multiple `postDataEntries` to verify that we can just add things like that.

### DIFF
--- a/wpt-metadata/chromedriver/headful/webdriver/tests/bidi/input/set_files/invalid.py.ini
+++ b/wpt-metadata/chromedriver/headful/webdriver/tests/bidi/input/set_files/invalid.py.ini
@@ -1,0 +1,2 @@
+[invalid.py]
+  expected: TIMEOUT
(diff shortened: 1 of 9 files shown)


wpt-metadata/chromedriver/headful/webdriver/tests/bidi/input/set_files/invalid.py.ini
```diff
@@ -0,0 +1,2 @@
+[invalid.py]
+  expected: TIMEOUT
```
sadym-chromium, line 2: This one seems to be irrelevant

sadym-chromium (approved):

wpt-metadata/chromedriver/headful/webdriver/tests/bidi/input/set_files/invalid.py.ini
Lightning00Blade: Auto-updated. Will check one test is failing unexpectedly

PR #2349, opened 2024-06-25
fix: provide type in `browsingContext.userPromptClosed` event

sadym-chromium:


src/bidiMapper/modules/context/BrowsingContextImpl.ts
```diff
@@ -575,33 +578,39 @@ export class BrowsingContextImpl {
 
     this.#cdpTarget.cdpClient.on('Page.javascriptDialogClosed', (params) => {
       const accepted = params.result;
-
+      assert(
```
Lightning00Blade, line 581: Just return here else this will be `uncaugh` .

sadym-chromium: good point. Added log and return `unknown` to prevent clients from hanging waiting for the event

Lightning00Blade (approved):

PR #2351, opened 2024-06-25
feat!: respect `unhandledPromptBehavior` capability

sadym-chromium:
BREAKING CHANGE: default behavior changed from `ignore` to `dismiss`.

Failing WPT tests should be addressed by https://github.com/web-platform-tests/wpt/pull/46895.

* Puppeteer failures are expected, as Puppeteer has to be updated after Mapper release to pass `unhandledPromptBehavior: {default: 'ignore'}`.
* chromedriver has to pass the capability to Mapper.


OrKoN:
the commit is not conventional, although I am not sure it should be a breaking change since it is technically a fix

sadym-chromium:
> the commit is not conventional, although I am not sure it should be a breaking change since it is technically a fix

It's breaking in terms of some work would be required on Puppeteer side to keep API consistent.

OrKoN (approved):

PR #2366, opened 2024-07-01
chore: fix session.new method name

Lightning00Blade:


src/bidiMapper/CommandProcessor.ts
```diff
@@ -311,7 +311,7 @@ export class CommandProcessor extends EventEmitter<CommandProcessorEventsMap> {
       // Session domain
       // keep-sorted start block=yes
       case 'session.new':
-        return await this.#sessionProcessor.create(command.params);
+        return await this.#sessionProcessor.new(command.params);
```
OrKoN, line 314: can it be considered a reserved keyword by some tools?

sadym-chromium: it seems to be a valid method name

OrKoN (approved):

sadym-chromium (approved):

PR #2370, opened 2024-07-01
fix: support for cookies in ContinueRequest

Lightning00Blade:
I still need to see why the data is not surfacing on the request correctly, but the feature should now work.
The WPT test will pass once we fix the `request->headers`.

### DIFF
--- a/src/bidiMapper/modules/network/NetworkProcessor.ts
+++ b/src/bidiMapper/modules/network/NetworkProcessor.ts
@@ -88,8 +88,6 @@ export class NetworkProcessor {
       Network.InterceptPhase.BeforeRequestSent,
     ]);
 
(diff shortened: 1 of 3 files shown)


sadym-chromium (approved):
Don't we have WPT tests for this? And it would be great to add an e2e test

Lightning00Blade:
@sadym-chromium there are WPT test for this, but the data they expect is not correct from our side, I am looking into it, happy to wait to merge this after that investigation.

sadym-chromium:
> @sadym-chromium there are WPT test for this, but the data they expect is not correct from our side, I am looking into it, happy to wait to merge this after that investigation.

I think we can merge this and update the solution to finegrain it later

PR #2397, opened 2024-07-08
fix: default `beforeUnload` behavior is `accept`

sadym-chromium:


Lightning00Blade:
Maybe marks as `fix:` not `chore:`

PR #2406, opened 2024-07-11
feat: include vendor-prefixed additional request data

OrKoN:
- 'goog:postData'
- 'goog:hasPostData'
- 'goog:resourceType'

The testing is planned via Puppeteer.

### DIFF
--- a/src/bidiMapper/modules/network/NetworkRequest.ts
+++ b/src/bidiMapper/modules/network/NetworkRequest.ts
@@ -835,6 +835,10 @@ export class NetworkRequest {
       headersSize: computeHeadersSize(headers),
       bodySize: this.#bodySize,
       timings: this.#timings,
+      // @ts-expect-error CDP-specific attribute.
+      'goog:postData': this.#request.info?.request?.postData,
+      'goog:hasPostData': this.#request.info?.request?.hasPostData,
+      'goog:resourceType': this.#request.info?.type,
     };
   }
 
--- a/tests/network/test_add_intercept.py
+++ b/tests/network/test_add_intercept.py
@@ -16,8 +16,9 @@ import re
 
 import pytest
 from anys import ANY_DICT, ANY_LIST, ANY_NUMBER, ANY_STR
-from test_helpers import (ANY_TIMESTAMP, ANY_UUID, execute_command,
-                          send_JSON_command, subscribe, wait_for_event)
+from test_helpers import (ANY_TIMESTAMP, ANY_UUID, AnyExtending,
+                          execute_command, send_JSON_command, subscribe,
+                          wait_for_event)
 
 
 @pytest.mark.asyncio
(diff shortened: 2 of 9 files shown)


src/bidiMapper/modules/network/NetworkRequest.ts
```diff
@@ -835,6 +835,10 @@ export class NetworkRequest {
       headersSize: computeHeadersSize(headers),
       bodySize: this.#bodySize,
       timings: this.#timings,
+      // @ts-expect-error CDP-specific attribute.
+      'goog:postData': this.#request.info?.request?.postData,
+      'goog:hasPostData': this.#request.info?.request?.hasPostData,
+      'goog:resourceType': this.#request.info?.type,
```
sadym-chromium, line 841: Can you please add e2e test verifying the fields are in place

OrKoN: I plan to test using Puppeteer since tests already exist there. The breakage can be caught through Puppeteer tests.

sadym-chromium: Having existing pipeline, the Puppeteer tests can be run too late. A single test verifying the fields are in place would be enough.

sadym-chromium:
Were the `tests/browsing_context/capture_screenshot/gradient_with_alpha_channel.png` and `tests/browsing_context/capture_screenshot/gradient_without_alpha_channel.png` added by mistake or intentionally?

src/bidiMapper/modules/network/NetworkRequest.ts
OrKoN: Puppeteer tests are forced on release. What is the scenario when it can it be too late?

sadym-chromium: e.g. roll into chromedriver. Are there any issues with adding the field?

OrKoN: do we plan to roll unreleased chromedriver? do we run e2e tests during the mapper roll?

OrKoN: A single e2e test would not be feasible since the data is included into multiple events that include base event params, and I do not really see value in adding the tests since we already test using Puppeteer before every release.

sadym-chromium (approved):

src/bidiMapper/modules/network/NetworkRequest.ts
sadym-chromium: > do we plan to roll unreleased chromedriver? do we run e2e tests during the mapper roll?

We roll the latest revision to chromedriver.

OrKoN: hm.. this sounds unexpected to me, why are not we rolling the latest released version?

OrKoN: filed crbug.com/352663873

OrKoN: I assume the issue with tests is resolved. If crbug.com/352663873 results in still rolling unreleased versions, we could enable Puppeteer tests on each PR. It takes 5-6 minutes unsharded and with sharding it could be down to 1-3 minutes.

FelineAsshole:
Thank you very much. I have no idea how to write code. I literally happened to download git, then the nest day i had 3 projects going while watching a couple thousand in a perpetual future, it was super f'ed. worth it tho

PR #2451, opened 2024-07-30
feat: support ESM module

Lightning00Blade:
Also makes the MapperTab use ESM rather then CommonJS which is smaller.


sadym-chromium (approved):

sadym-chromium:
It looks like local e2e run is broken with this change

Issue #2415, opened 2024-07-15
[Feature] Mechanism to retrieve supported Chrome stable version


Lightning00Blade:
Puppeteer by default downloads the latest stable version.
We need a way to know with which is the first stable release `chromium-bidi` will work.
This is due to the fact that we use Canary for testing to reduce the developer friction.

Green00101:
I'd like to work on this. Is it still available?

Lightning00Blade:
Closing as obsolete.

PR #2483, opened 2024-08-07
fix: always provide url in `browsingContext.navigationStarted`

sadym-chromium:
Rely on `Page.frameRequestedNavigation` and `Page.frameScheduledNavigation`.


OrKoN:
Before landing this let's discuss https://github.com/GoogleChromeLabs/chromium-bidi/issues/2415 If this lands, we probably won't be able to update chromium-bidi in Puppeteer until 129 is out.

OrKoN:
blocked on https://github.com/GoogleChromeLabs/chromium-bidi/issues/2536 Could we mark it as a breaking change perhaps? until the roller is able to bump the versions

OrKoN (approved):

OrKoN:
Note that the label alone would not mark it as a breaking change IIRC.

sadym-chromium:
converted to draft to prevent accidental merge

OrKoN (approved):

Issue #2484, opened 2024-08-08
Update the workflows making changes


OrKoN:
- we should use pull requests to submit changes (https://github.com/peter-evans/create-pull-request push-to-fork)
- reach out to me to configure a token for browser-automation-bot

OrKoN:
@Lightning00Blade could you please take a look?

PR #2488, opened 2024-08-09
ci: push to fork

Lightning00Blade:
@OrKoN You can re-add the token back, so we get those rollers up. While I investigate the `update-expectations` one.

Ref #2484


.github/workflows/wpt-quick.yml
```diff
@@ -245,7 +245,8 @@ jobs:
       - name: Checkout
         uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
         with:
-          token: ${{ secrets.BROWSER_AUTOMATION_BOT_TOKEN }}
+          # Temporary disabled
```
OrKoN, line 248: I have set the token

PR #2492, opened 2024-08-12
chore: add timestamp to chromedriver logs

Lightning00Blade:


sadym-chromium (approved):

sadym-chromium:
The next logical step would be to get the log path in NodeJS runner from CLI, and unify the cd and njs runners

PR #2498, opened 2024-08-13
test: expectations update

Lightning00Blade:


Lightning00Blade:
This was a test run, and it seem that now it's working as intended,

PR #2524, opened 2024-08-22
fix: allow pen to hover over

sadym-chromium:
Even if the pen is not touching the surface, the move events should be dispatched.

### DIFF
--- a/src/bidiMapper/modules/input/ActionDispatcher.ts
+++ b/src/bidiMapper/modules/input/ActionDispatcher.ts
@@ -410,6 +410,8 @@ export class ActionDispatcher {
             );
             break;
           case Input.PointerType.Pen:
+            // Even if the pressed size is 0, we still need to send the event, as pen can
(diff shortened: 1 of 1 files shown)


Lightning00Blade (dismissed):
LGTM, can we add a comment explaining this.

Lightning00Blade (approved):

OrKoN:
@Lightning00Blade expectation update seems to fail https://github.com/GoogleChromeLabs/chromium-bidi/actions/runs/10506682096/job/29107080659?pr=2524 ?

sadym-chromium:
> @Lightning00Blade expectation update seems to fail https://github.com/GoogleChromeLabs/chromium-bidi/actions/runs/10506682096/job/29107080659?pr=2524 ?

the test is flaky

OrKoN:
Right, so should not the expectation update mark it as such or disable it? or can we manually mark it as flaky?

Issue #2539, opened 2024-08-28
Using `*` as locator returns no nodes


christian-bromann:
This behavior is observed in `stable` and `canary` (130.0.6682.2)

## Expected Behavior

If I call `browsingContext.locateNodes` with the following parameters:

```
{
  "locator": {"type":"css","value":"*"},
  "context":"4ab66b36-ae31-4c38-90bb-40d97bd48b36"
}
```

I would expect that it would fetch the `<head /> tag as it is the first element in the DOM tree.

## Actual Behavior

__Firefox:__ ✅ works as expected and returns

```
{
    "nodes": [
        {
            "type": "node",
            "sharedId": "d2030884-983d-45f9-85fd-10552d1e9f70",
            "value": {
                "nodeType": 1,
                "localName": "head",
                "namespaceURI": "http://www.w3.org/1999/xhtml",
                "childNodeCount": 0,
                "attributes": {},
                "shadowRoot": null
            }
        },
        {
            "type": "node",
            "sharedId": "9d66bb03-20fd-4e5d-bacf-68bbdda1d15d",
            "value": {
                "nodeType": 1,
                "localName": "body",
                "namespaceURI": "http://www.w3.org/1999/xhtml",
                "childNodeCount": 0,
                "attributes": {},
                "shadowRoot": null
            }
        }
    ]
}
```

__Chrome:__ ❌ fails even though `*` is a valid selector and returns

```ts
{
    "nodes": []
}
```

## Reproducible Example

See https://github.com/christian-bromann/bidi-locateNodes-error, clone the repo, install dependencies, and call `npm run wdio`

sadym-chromium:
e2e repro works: https://github.com/GoogleChromeLabs/chromium-bidi/pull/2543. Looking further

sadym-chromium:
UPD: the issue should be fixed by https://github.com/GoogleChromeLabs/chromium-bidi/pull/2543

sadym-chromium:
The root cause is in the fact Chromium BiDi used `document.body` instead of `document`, which excluded `head` form the selectors.

christian-bromann:
@sadym-chromium why would it still return no nodes given the selector `*` then? I would assume it would return the first node in the `document.body` tree.

sadym-chromium:
> Thanks for fixing this. @sadym-chromium why would it still return no nodes given the selector `*` then? I would assume it would return the first node in the `document.body` tree.

[...]

PR #2543, opened 2024-08-29
fix: css and xPath selector start node is document

sadym-chromium:
Addressing https://github.com/GoogleChromeLabs/chromium-bidi/issues/2539

According to the spec, the default `startNodes` of the `browsingContext.locateNodes` is `document`, not `document.body`.

### DIFF
--- a/src/bidiMapper/modules/context/BrowsingContextImpl.ts
+++ b/src/bidiMapper/modules/context/BrowsingContextImpl.ts
@@ -1255,7 +1255,7 @@ export class BrowsingContextImpl {
                 return [...element.querySelectorAll(cssSelector)];
               };
 
-              startNodes = startNodes.length > 0 ? startNodes : [document.body];
+              startNodes = startNodes.length > 0 ? startNodes : [document];
               const returnedNodes = startNodes
                 .map((startNode) =>
                   // TODO: stop search early if `maxNodeCount` is reached.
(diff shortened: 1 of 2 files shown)


src/bidiMapper/modules/context/BrowsingContextImpl.ts
```diff
@@ -1530,7 +1530,12 @@ export class BrowsingContextImpl {
                 }
               }
 
-              startNodes = startNodes.length > 0 ? startNodes : [document.body];
+              startNodes =
+                startNodes.length > 0
+                  ? startNodes
+                  : Array.from(document.documentElement.children).filter(
```
sadym-chromium, line 1536: IDK if anything in `head` have a11y selectors.

OrKoN: is this filtering in the spec?

sadym-chromium: Yes:
> 5. Let child nodes be an empty [list](https://infra.spec.whatwg.org/#list) and, for each node child in the [children (https://dom.spec.whatwg.org/#concept-tree-child) of context node:
>     1. If child implements [Element](https://dom.spec.whatwg.org/#element), [append](https://infra.spec.whatwg.org/#list-append) child to child nodes.

OrKoN (approved):

PR #2571, opened 2024-09-03
fix: emulate initial navigation

sadym-chromium:
There should be a `browsingContext.navigationStarted` for the initial pager navigation.


sadym-chromium:
https://github.com/web-platform-tests/wpt/issues/47942

PR #2574, opened 2024-09-04
feat: emit `browsingContext.navigationFailed` event

sadym-chromium:
Introduce `#currentNavigation` deferred, which represents the ongoing navigation.


jm6340812:
Arreglar mi cuenta

PR #2581, opened 2024-09-04
chore: heuristic in console log serialization

sadym-chromium:
This change saves some CDP round trips on console log events, which improves Mapper performance in case of tons of console log events are written. This seems to be a bottleneck for testdriverJS BiDi.


OrKoN (approved):
Shall we plan the work to add deep serialization into CDP messages? As an opt-in, when enabling the Log domain?

PR #2606, opened 2024-09-10
fix: do not throw an error when unsubscribing from event if the CDP session is closed

sadym-chromium:
This causes a false-negative WPT tests in case of testdriverJS BiDi.

### DIFF
--- a/src/bidiMapper/modules/cdp/CdpTarget.ts
+++ b/src/bidiMapper/modules/cdp/CdpTarget.ts
@@ -48,6 +48,7 @@ export class CdpTarget {
   readonly #unhandledPromptBehavior?: Session.UserPromptHandler;
   readonly #logger: LoggerFn | undefined;
 
+  #deviceAccessEnabled = false;
   #cacheDisableState = false;
   #networkDomainEnabled = false;
   #fetchDomainStages = {
@@ -288,6 +289,9 @@ export class CdpTarget {
     } catch (err) {
       this.#logger?.(LogType.debugError, err);
       this.#networkDomainEnabled = !enabled;
+      if (!this.#isExpectedError(err)) {
+        throw err;
+      }
     }
   }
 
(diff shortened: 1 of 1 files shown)


sadym-chromium:
> I don't think this is the way we want to go, we would want to create a property on the CDP target to know what the status update or not, we already do that for Cache and Network, and Fetch, we should do it here as well. I guess we are sending a `DeviceAccess.disable` and that is causing the error, having a cached prevent from sending it.

If the `Target.detachedFromTarget` is received BEFORE the command, the browsing context is not iterated. The problem happens when there is a race condition between command and `Target.detachedFromTarget` event. E.g. in this case the timing is the following:
1. Send command `DeviceAccess.disable`.
2. Received event `Target.detachedFromTarget`.
3. Received command response "Session with given id not found".

Having that, we can either add a heuruistic check the exception

Lightning00Blade:
Can you take a look at this - https://github.com/GoogleChromeLabs/chromium-bidi/pull/2609
I think this will solve the problem you are trying to fix here, but as it prevents sending usless command to CDP.

Lightning00Blade:
If we want we should have a general solution to the problem of unsubscribing as there are other commands that have similar behaviour.

sadym-chromium:
> Can you take a look at this - #2609 I think this will solve the problem you are trying to fix here, but as it prevents sending usless command to CDP.

I added a commit with test failing but expecting to pass. The PR does not address the race condition.

src/bidiMapper/modules/cdp/CdpTarget.ts
```diff
@@ -303,17 +307,49 @@ export class CdpTarget {
...
-    });
+    try {
+      await this.#cdpClient.sendCommand('Network.setCacheDisabled', {
+        cacheDisabled,
+      });
+    } catch (err) {
+      if (!this.#isExpectedError(err)) {
+        throw err;
+      }
+      this.#logger?.(LogType.debugError, err);
```
Lightning00Blade, line 318: I think the check for throwing should always be last as we want to log the error and reset the state of the command. Else we will have wrong state on the object.

Lightning00Blade (approved):

PR #2682, opened 2024-10-16
chore: upload git patch for updated WPT expectations

sadym-chromium:


.github/workflows/wpt.yml
```diff
@@ -260,6 +260,20 @@ jobs:
         run: |
           rm -rf wpt-metadata/chromedriver wpt-metadata/mapper
           mv all-artifacts/updated-wpt-metadata/* ./wpt-metadata/
+      - name: git diff
```
Lightning00Blade, line 263: I don't think this step is needed.

sadym-chromium: It was added for debug purposes

PR #2742, opened 2024-11-04
ci: download headless shell when needed

sadym-chromium:
Download headless shell when `headless=old` is used. Required for https://github.com/GoogleChromeLabs/chromium-bidi/pull/2719.

### DIFF
--- a/tools/run-e2e.mjs
+++ b/tools/run-e2e.mjs
@@ -171,7 +171,14 @@ if (argv.fileOrFolder) {
       : [argv.fileOrFolder]),
   );
 }
-if (process.env.HEADLESS === 'false' && !argv.k) {
+
+// Default headless is `new`.
+const HEADLESS = ['old', 'new', 'false'].includes(process.env.HEADLESS)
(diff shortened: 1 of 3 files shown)


tools/run-e2e.mjs
```diff
@@ -181,8 +188,9 @@ if (argv.k) {
 const e2eProcess = child_process.spawn('pipenv', e2eArgs, {
   stdio: ['inherit', 'pipe', 'pipe'],
   env: {
-    BROWSER_BIN: installAndGetChromePath(),
     ...process.env,
+    BROWSER_BIN: installAndGetChromePath(),
```
Lightning00Blade, line 192: I think this should stay where it is else, the provided BROWSER_BIN will be wrong in the case where you want to run custom binary.

sadym-chromium: `installAndGetChromePath` respects the `BROWSER_BIN` env variable, so I guess we should force to use one returned by `installAndGetChromePath`

PR #2748, opened 2024-11-04
ci: install headless shell when required

sadym-chromium:
Required for https://github.com/GoogleChromeLabs/chromium-bidi/pull/2719


Lightning00Blade (approved):
LGTM, but it would be nice to consolidate the install of all in `--all` argument so locally its easy to install as well.

PR #2762, opened 2024-11-08
fix: allow setting `bluetooth.simulateAdapter` multiple times

sadym-chromium:
Align with the spec (step 6) of [`bluetooth.simulateAdapter`](https://webbluetoothcg.github.io/web-bluetooth/#bluetooth-simulateAdapter-command).


sadym-chromium:
@alexnj there is no restrictions in [the spec](https://webbluetoothcg.github.io/web-bluetooth/#bluetooth-simulateAdapter-command) on setting bluetooth adapter state several times, but the CDP commands fail if done twice.

OrKoN (approved):

PR #2780, opened 2024-11-15
feat: implement `browser.getClientWindows`

sadym-chromium:
Addressing https://github.com/GoogleChromeLabs/chromium-bidi/issues/2777

PR #2783, opened 2024-11-15
fix: `browser.getClientWindows` returns a unique per window value

sadym-chromium:
Addressing issue pointed in https://github.com/GoogleChromeLabs/chromium-bidi/pull/2780#discussion_r1844312850

### DIFF
--- a/src/bidiMapper/modules/browser/BrowserProcessor.ts
+++ b/src/bidiMapper/modules/browser/BrowserProcessor.ts
@@ -113,7 +113,7 @@ export class BrowserProcessor {
       {targetId},
     );
     return {
-      // `activate` is not supported in CDP yet.
+      // `active` is not supported in CDP yet.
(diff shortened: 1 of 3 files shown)


src/bidiMapper/modules/browser/BrowserProcessor.ts
```diff
@@ -113,7 +113,7 @@ export class BrowserProcessor {
       {targetId},
     );
     return {
-      // Is not supported in CDP yet.
+      // `activate` is not supported in CDP yet.
```
OrKoN, line 116:
```suggestion
      // `active` is not supported in CDP yet.
```

FWIW I would default to true. In general, I am not sure how useful this data will be to WebDriver BiDi client given that the input can be sent to any browsing context.

OrKoN (approved):

PR #2787, opened 2024-11-18
chore: throw only errors

Lightning00Blade:


src/bidiMapper/modules/context/BrowsingContextImpl.ts
```diff
@@ -520,7 +520,7 @@ export class BrowsingContextImpl {
           this.id,
         );
         this.#pendingCommandNavigation.reject(
-          ChromiumBidi.BrowsingContext.EventNames.NavigationAborted,
+          new UnknownErrorException('navigation aborted'),
```
sadym-chromium, line 523: `new UnknownErrorException(ChromiumBidi.BrowsingContext.EventNames.NavigationAborted)`?

PR #2800, opened 2024-11-21
test: speed up `assert_no_more_messages`

sadym-chromium:
Apply heuristic. Run a script on each exising browsing context instead of waiting. If there are events, most likely they will be enqueued before the script is finished.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -21,9 +21,9 @@ from uuid import uuid4
 import pytest
 import pytest_asyncio
 import websockets
-from test_helpers import (execute_command, get_tree, goto_url,
+from test_helpers import (AnyExtending, execute_command, get_tree, goto_url,
                           merge_dicts_recursively, read_JSON_message,
-                          wait_for_event, wait_for_events)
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -323,10 +323,31 @@ async def read_sorted_messages(
 @pytest.fixture
 def assert_no_more_messages(websocket):
     """Assert that there are no more messages on the websocket."""
-    async def assert_no_more_messages(timeout: float | None):
-        with pytest.raises(asyncio.TimeoutError):
-            await asyncio.wait_for(read_JSON_message(websocket),
-                                   timeout=timeout)
+    async def assert_no_more_messages():
```
OrKoN, line 326: I am not sure it guarantees much for network events for example

sadym-chromium: neither does a timeout of a second

sadym-chromium: But we do need to check there are no more messages are coming. I guess this approach covers the risks it is expected to cover.

Issue #2660, opened 2024-10-07
NetworkUtils.matchUrlPattern cannot parse certain URLs.


OrKoN:
for example, `https://web-platform.test:8443/webdriver/tests/support/inline.py?doc=%3C%21doctype+html%3E%0A%3Cmeta+charset%3DUTF-8%3E%0A%3Cdiv+id%3D%27from-initial%27%3Einitial%3C%2Fdiv%3E&mime=text%2Fhtml&charset=UTF-8`

it seems there could be an issue with the pattern polyfill or we might be holding it wrong. Perhaps, we should implement the matching ourselves to avoid these issues?

cc @Lightning00Blade

OrKoN:
Unit tests https://github.com/GoogleChromeLabs/chromium-bidi/pull/2661

PR #2832, opened 2024-11-27
fix: implement the pattern matching according to the spec 

OrKoN:
It looks like the URL pattern spec has changed and more characters are considered to be special tokens. WebDriver BiDi standard does not seem to be, therefore, compatible with it at the moment. This change drops urlpattern-polyfill and replaces it with custom implementation according to the spec.

Closes https://github.com/GoogleChromeLabs/chromium-bidi/issues/2660


OrKoN:
Needs more work

sadym-chromium (approved):

Lightning00Blade (approved):
LGTM, there is some custom code in the rollup config could you please remove that as well?

PR #2836, opened 2024-11-28
chore: extract navigation tracking

sadym-chromium:
Extract navigation tracking logic to a dedicated class. Simplify the logic. Required to be able to make further adjustments to the navigation logic.

Out of scope: document tracking.

### DIFF
--- a/src/bidiMapper/modules/context/BrowsingContextImpl.ts
+++ b/src/bidiMapper/modules/context/BrowsingContextImpl.ts
@@ -260,10 +260,6 @@ export class BrowsingContextImpl {
     load: new Deferred<void>(),
   };
 
-  #navigation = {
-    withinDocument: new Deferred<void>(),
-  };
-
   #url: string;
   readonly #eventManager: EventManager;
   readonly #realmStorage: RealmStorage;
@@ -623,7 +619,6 @@ export class BrowsingContextImpl {
       }
       const timestamp = BrowsingContextImpl.getTimestamp();
       this.#url = params.url;
-      this.#navigation.withinDocument.resolve();
       this.#navigationTracker.navigatedWithinDocument(
         params.url,
         params.navigationType,
@@ -968,14 +963,6 @@ export class BrowsingContextImpl {
   #documentChanged(loaderId?: Protocol.Network.LoaderId) {
     if (loaderId === undefined || this.#loaderId === loaderId) {
       // Same document navigation. Document didn't change.
-      if (this.#navigation.withinDocument.isFinished) {
-        this.#navigation.withinDocument = new Deferred();
-      } else {
-        this.#logger?.(
-          BrowsingContextImpl.LOGGER_PREFIX,
-          'Document changed (navigatedWithinDocument)',
-        );
-      }
       return;
     }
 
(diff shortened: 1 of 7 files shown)


src/bidiMapper/modules/context/NavigationTracker.ts
```diff
@@ -0,0 +1,214 @@
...
+  // Tracks CDP events on the navigation. Used to distinguish new navigations from the
+  // initial one.
+  cdpStates = {
+    frameScheduledNavigation: false,
+    frameRequestedNavigation: false,
+    frameStartedLoading: false,
+  };
+
+  markFragmentNavigated() {
+    this.started.resolve();
```
OrKoN, line 49: so fragment navigations should not have other navigation events like navigationStarted as per spec.

src/bidiMapper/modules/context/NavigationTracker.ts
```diff
@@ -0,0 +1,214 @@
...
+      url: this.url,
+    };
+  }
+
+  markStarted(): void {
+    this.started.resolve();
+  }
+}
+
+export class NavigationTracker {
```
OrKoN, line 92: could we perhaps include unit tests to indicate all the expectations we have from the NavigationTracker (similar to what Nik has done for the network manager/storage)

src/bidiMapper/modules/context/NavigationTracker.ts
```diff
@@ -0,0 +1,214 @@
...
+  frameRequestedNavigation = (url: string): void => {
+    // Signals that it's a new navigation:
+    // 1. The `Page.frameStartedLoading` was already emitted.
+    // 2. The `Page.frameRequestedNavigation` was already emitted. Note that having
+    //    `Page.frameScheduledNavigation` emitted at this point does not signal the new
+    //    navigation.
+    // 3. The current navigation is the initial one, and the URL does not match
+    //    `about:blank`.
+    const isNewNavigation =
+      this.#currentNavigation.cdpStates.frameStartedLoading ||
```
OrKoN, line 202: how does cdpStates track if the events came from a current navigation or from one of the past navigations?

OrKoN (approved):
It's a little difficult to track if the behavior is the same. Assuming it is, this LGTM. I would suggest adding unit tests for various navigation tracker behaviors.

src/utils/Time.spec.ts
```diff
@@ -0,0 +1,33 @@
+/**
```
Lightning00Blade, line 1: Nit: non Class files should start with small letter.

src/bidiMapper/modules/context/BrowsingContextImpl.ts
```diff
@@ -419,8 +391,9 @@ export class BrowsingContextImpl {
       if (this.id !== params.frame.id) {
         return;
       }
-      this.#url = params.frame.url + (params.frame.urlFragment ?? '');
-      this.#pendingNavigationUrl = undefined;
+      const url = params.frame.url + (params.frame.urlFragment ?? '');
+      this.#url = url;
```
Lightning00Blade, line 395: In a follow up would it make sense to refactor the Url to come from the navigation tracking as it knows when its committed and when not.
So we change 
```ts
get #url(){
  return this.#navigationTracker.url;
}
```

sadym-chromium: I guess it should come from document tracker instead when it is implemented.

src/bidiMapper/modules/context/NavigationTracker.ts
Lightning00Blade: +1 to this.

Lightning00Blade (approved):

sadym-chromium:
closing in favor of https://github.com/GoogleChromeLabs/chromium-bidi/pull/2881

PR #2847, opened 2024-11-29
test: snapshot test navigation

sadym-chromium:
Some of the tests are flaky due to racy implementation. The behavior is not guaranteed to be correct, but required for consistency check.

### DIFF
--- a/README.md
+++ b/README.md
@@ -311,7 +311,7 @@ HEADLESS=new npm run e2e
 #### Updating snapshots
 
 ```sh
-npm run e2e -- --snapshot-update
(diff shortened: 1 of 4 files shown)


Lightning00Blade (approved):
LGTM, but my only concern is that we use the `read_sorted_messages` an in some cases we want to verify order as well, but I don't think it fully applies to this test.

sadym-chromium:
> LGTM, but my only concern is that we use the `read_sorted_messages` an in some cases we want to verify order as well, but I don't think it fully applies to this test.

Right. If we need to verify order, we can add a halper for reading X messages.

PR #2848, opened 2024-11-29
fix: set navigationInitiatedByCommand when calling traverseHistory

OrKoN:


sadym-chromium:
Do we have e2e or wpt test for this?

sadym-chromium:
I believe this is not required anymore

Issue #2856, opened 2024-12-02
Align navigation with the spec


sadym-chromium:
> [!IMPORTANT]
> ### 🚚 Moved to [crbug.com/411434924](https://crbug.com/411434924)
> Please star and follow **[crbug.com/411434924](https://crbug.com/411434924)** on the Chromium Issue Tracker for updates.

---

Currently, the navigation events are based on [some CDP heuristics](https://github.com/GoogleChromeLabs/chromium-bidi/blob/main/src/bidiMapper/modules/context/BrowsingContextImpl.ts#L85-L106), which do not always allow for proper events  emitting and commands results.

### Related documents
* [Chromium BiDi navigation overview](http://goto.google.com/webdriver:bidi-navigation)
* [Detect navigation started](http://goto.google.com/webdriver:detect-navigation-started)
* [Deterministic random values](http://go/webdriver:deterministic-random-values)

### WPT
* [x] [`browsingContext.navigate`](https://wpt.fyi/results/webdriver/tests/bidi/browsing_context/navigate?label=master&product=chrome%5Bexperimental%5D&product=firefox&aligned=)
* [x] [`browsingContext.navigationStarted`](https://wpt.fyi/results/webdriver/tests/bidi/browsing_context/navigation_started/navigation_started.py?label=master&product=chrome%5Bexperimental%5D&product=firefox&aligned=)
* [x] [browsingContext.navigationFailed](https://wpt.fyi/results/webdriver/tests/bidi/browsing_context/navigation_failed/navigation_failed.py?label=master&product=chrome%5Bexperimental%5D&product=firefox&aligned)
* [ ] Add tests for [`browsingContext.navigationAborted`](https://w3c.github.io/webdriver-bidi/#event-browsingContext-navigationAborted)

OrKoN:
Is https://github.com/GoogleChromeLabs/chromium-bidi/blob/main/src/bidiMapper/modules/cdp/CdpTarget.ts#L244C9-L248C53 still required?

sadym-chromium:
Moved to [crbug.com/411434924](https://crbug.com/411434924)
Please star and follow [crbug.com/411434924](https://crbug.com/411434924) on the Chromium Issue Tracker for updates.

PR #2879, opened 2024-12-10
chore: emulate `frameStartedNavigating` event

sadym-chromium:
Implementation of [Alternative 3. Rely on “Network.requestWillBeSent” CDP event](http://goto.google.com/webdriver:detect-navigation-started#bookmark=id.64balpqrmadv). Required for https://github.com/GoogleChromeLabs/chromium-bidi/issues/2856.

Emit `frameStartedNavigating` event on CdpTarget before `Network.requestWillBeSent`.


src/bidiMapper/modules/cdp/CdpTarget.ts
```diff
@@ -57,7 +57,9 @@ export class CdpTarget {
 
   #deviceAccessEnabled = false;
   #cacheDisableState = false;
-  #networkDomainEnabled = false;
+  // Even though the CDP `Network` event should be always enabled, some additional domains
+  // can be or be not required.
+  #networkBidiDomainEnabled = false;
```
OrKoN, line 62: do we need to rename it? I think we are still toggling the CDP's network domain and not the bidi network domain.

src/bidiMapper/modules/cdp/CdpTarget.ts
```diff
@@ -437,13 +441,13 @@ export class CdpTarget {
   }
 
   async #toggleNetwork(enable: boolean): Promise<void> {
-    this.#networkDomainEnabled = enable;
+    this.#networkBidiDomainEnabled = enable;
     try {
       await this.#cdpClient.sendCommand(
         enable ? 'Network.enable' : 'Network.disable',
```
OrKoN, line 447: could we force enable here instead of #unblock above?

src/cdp/CdpConnection.ts
```diff
@@ -154,6 +154,22 @@ export class MapperCdpConnection implements CdpConnection {
...
       );
+      if (message.method === 'Network.requestWillBeSent') {
+        /**
+         * Emulate CDP event 'Page.frameStartedNavigating' right before the
+         * `Network.requestWillBeSent`.
+         * http://go/webdriver:detect-navigation-started#bookmark=id.64balpqrmadv
+         */
+        const eventParams = message.params;
+        if (eventParams.loaderId === eventParams.requestId) {
+          client?.emit('Page.frameStartedNavigating', {
```
OrKoN, line 165: that would be visible to external clients right? Let's emit this event on an internal emitter only for the mapper implementation

src/bidiMapper/modules/cdp/CdpTarget.ts
sadym-chromium: I meant "Enable all required for BiDi Network Module"

sadym-chromium: Actually we don't need this field at all

sadym-chromium: reverted

src/bidiMapper/modules/cdp/CdpTarget.ts
sadym-chromium: right, done

src/cdp/CdpConnection.ts
sadym-chromium: The event will be exposed in `CdpClient` interface and to the clients as `goog:cdp.Page.frameStartedNavigating` event. Are you concerned about the latest, or both?

OrKoN: I am concerned that event will be observable by Puppeteer and by clients creating a custom CDP session through Puppeteer who might start depending on it or break due to unknown event/event name format. I do not think even prefixed events should be emitted there. What is the difficulty with using this event internally?

OrKoN: Does it even work with custom CDP connections like the one Puppeteer is using?

OrKoN (commented):
I see that 'Network.requestWillBeSent' is used once in chromium-bidi? what was the race-condition that you mentioned?

OrKoN:
WDYT if we implement it this way instead https://github.com/GoogleChromeLabs/chromium-bidi/pull/2880 (plus network domain enablement changes )?  Note that TargetEvents.FrameStartedNavigating should always fire before the network event since it is subscribed to the network event first.

sadym-chromium:
> WDYT if we implement it this way instead #2880 (plus network domain enablement changes )? Note that TargetEvents.FrameStartedNavigating should always fire before the network event since it is subscribed to the network event first.

done

OrKoN (changes requested):
EmulatedCdpMapping is not needed anymore?

sadym-chromium:
> EmulatedCdpMapping is not needed anymore?

right, my bad, artefact after merging. Fixed

OrKoN (approved):

OrKoN:
(No new Puppeteer failures)

PR #2880, opened 2024-12-10
chore: detect a frame started navigating target event

OrKoN:


sadym-chromium:
Thanks! Merged into https://github.com/GoogleChromeLabs/chromium-bidi/pull/2879

PR #2881, opened 2024-12-10
chore: refactor navigation

sadym-chromium:
Addressing [Align navigation with the spec](https://github.com/GoogleChromeLabs/chromium-bidi/issues/2856).

Navigation tracker relies on the following events:
* `Page.frameNavigated`
* TargetEvents.FrameStartedNavigating (`Network.requestWillBeSent`)
* `Page.navigatedWithinDocument`
* `Page.frameRequestedNavigation`
* `Page.javascriptDialogOpening:beforeunload`


src/bidiMapper/modules/context/NavigationTracker.ts
```diff
@@ -19,50 +19,69 @@
...
+      timestamp: getTimestamp(),
+      url: this.url,
+    };
   }
 }
 
+/**
+ * Keeps track of navigations. Details: http://go/webdriver:bidi-navigation
+ */
 export class NavigationTracker {
```
OrKoN, line 71: could we cover major use cases with unit tests?

OrKoN:
There are a few new Puppeteer failures as well as far as I can see.

src/bidiMapper/modules/context/NavigationTracker.ts
sadym-chromium: added the major cases

tests/browsing_context/test_create.py
```diff
@@ -297,14 +300,41 @@ async def test_browsingContext_create_withUserGesture_eventsEmitted(
...
+            'params': {
+                'context': 'stable_0',
+                'navigation': 'stable_1',
+                'timestamp': ANY_TIMESTAMP,
+                'url': url_example,
+            },
+        }, {
+            'method': 'browsingContext.navigationStarted',
+            'params': {
+                'context': 'stable_0',
```
OrKoN, line 332: the order seems wrong, why is the navigationStarted after domContentLoaded?

tests/browsing_context/test_navigate_events.py
```diff
@@ -180,7 +180,52 @@ async def test_scriptNavigate_aboutBlank_checkEvents(websocket, context_id,
...
 
-    messages = await read_sorted_messages(9,
+    messages = await read_sorted_messages(7,
+                                          keys_to_stabilize=KEYS_TO_STABILIZE,
+                                          check_no_other_messages=True)
+    assert messages == snapshot(exclude=SNAPSHOT_EXCLUDE)
+
+
+@pytest.mark.asyncio
+async def test_aboutBlank_reload_checkEvents(websocket, context_id, html,
```
OrKoN, line 190: let's add tests for history updated and fragment navigations in before unload?

OrKoN, line 190: let's add tests that traverse history causes appropriate events?

tests/browsing_context/test_create.py
sadym-chromium: `read_sorted_messages` sorts messages alphabetically. We could introduce sorting-persisting method, but let's keep it out of scope of this PR

tests/browsing_context/test_navigate_events.py
sadym-chromium: follow-up?

tests/browsing_context/test_create.py
OrKoN: order seems to be important for these tests. I think we should use a non-sorted message reader for all new tests where the order matters.

OrKoN: > let's keep it out of scope of this PR

we could open and land a PR adding a new helper first and then use it for new tests in this PR

OrKoN:
I have landed the roll with existing changes into Puppeteer so that new regressions are clear. If anything fails in the Puppeteer job now, it is caused by the changes in this PR.

OrKoN:
> I have landed the roll with existing changes into Puppeteer so that new regressions are clear. If anything fails in the Puppeteer job now, it is caused by the changes in this PR.

I'd suggest we add e2e tests/fixes for Puppeteer scenarios that this change affects.

OrKoN:
"[oopif.spec] OOPIF should load oopif iframes with subresources and request interception" seems to be a Linux-only failure upstream.

src/bidiMapper/modules/context/NavigationTracker.ts
```diff
@@ -72,210 +102,365 @@ export class NavigationTracker {
...
+        'navigation canceled by concurrent navigation',
+      ),
+    );
+    const navigation = this.#createNavigation(url);
+    this.#pendingNavigation = navigation;
+    return navigation;
+  }
+
+  #setListeners(navigation: NavigationState) {
+    void navigation.started
```
OrKoN, line 171: so the way of dispatching events is not working as expected. Using Deferred.then creates a new microtask that makes events emitted to be emitted at a different relative order to the events scheduled when the deferred is resolved. I believe this is the reason for the failing data URL tests. The correct way to handle event would be to register the event at the same time as the deferred is resolved.

OrKoN (changes requested):

src/bidiMapper/modules/context/NavigationTracker.ts
OrKoN: in the case of the data URLs it allows network storage to dispatch the beforeRequest event before navigationStarted because network storage does not delay the event registration

src/bidiMapper/modules/context/NavigationTracker.spec.ts
```diff
@@ -0,0 +1,394 @@
...
+const YET_ANOTHER_URL = 'YET_ANOTHER_URL';
+// keep-sorted end
+
+describe('NavigationTracker', () => {
+  let navigationTracker: NavigationTracker;
+  let eventManager: SinonStubbedInstance<EventManager>;
+  let initialNavigationId: string;
+
+  async function assertNoNavigationEvents() {
+    await new Promise((resolve) => setTimeout(resolve, 1));
```
OrKoN, line 44: I think draining of the microtask queue masks the issues of when events have to happen

OrKoN: I would suggest we remove this and make the assertion sync

tests/browsing_context/test_create.py
```diff
@@ -112,67 +111,62 @@ async def test_browsingContext_windowOpen_nonBlank_eventsEmitted(
...
+        "params": {
+            'context': 'stable_0',
+            "url": "about:blank",
+            "children": None,
+            "parent": None,
+            "userContext": "default",
+            "originalOpener": ANY_STR,
+            'clientWindow': ANY_STR,
+        }
+    },
```
OrKoN, line 131: nit: I am curious why formatting gets weird here

OrKoN (approved):

PR #2893, opened 2024-12-13
test: split puppeteer tests into 4 shards

OrKoN:
Brings total duration from 7m to 3m.


OrKoN:
The setup and builds seems to be expensive for the job, so perhaps it's not worth it, it saves about 2-3 minutes only.

OrKoN:
> The setup and builds seems to be expensive for the job, so perhaps it's not worth it, it saves about 2-3 minutes only.

added wireit to see if it improves.

sadym-chromium (approved):

PR #2908, opened 2024-12-18
feat: throw a proper error if the browsing context was destroyed during action dispatching

sadym-chromium:
Get browsing context each time from the storage to guarantee it is not disposed.

+ e2e test

### DIFF
--- a/src/bidiMapper/modules/input/ActionDispatcher.ts
+++ b/src/bidiMapper/modules/input/ActionDispatcher.ts
@@ -28,6 +28,7 @@ import {
   isSingleGrapheme,
 } from '../../../utils/GraphemeTools.js';
 import type {BrowsingContextImpl} from '../context/BrowsingContextImpl.js';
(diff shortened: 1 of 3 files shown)


OrKoN (approved):

src/bidiMapper/modules/input/ActionDispatcher.ts
```diff
@@ -90,21 +91,33 @@ export class ActionDispatcher {
...
   #isMacOS: boolean;
+
   constructor(
     inputState: InputState,
-    context: BrowsingContextImpl,
+    browsingContextStorage: BrowsingContextStorage,
+    contextId: string,
     isMacOS: boolean,
   ) {
+    this.#browsingContextStorage = browsingContextStorage;
```
sadym-chromium, line 108: We could pass a delegate for getting the context, or add the logic to the browsing context to throw exception when accessing it's target when the browsing context is disposed. But I think this way is the most readable.

PR #2919, opened 2024-12-19
feat: provide proper network timings

sadym-chromium:
+ make e2e test offline.

Tested manually + WPT.

### DIFF
--- a/src/bidiMapper/modules/network/NetworkRequest.ts
+++ b/src/bidiMapper/modules/network/NetworkRequest.ts
@@ -23,27 +23,27 @@
 import type {Protocol} from 'devtools-protocol';
 
 import {
-  Network,
   type BrowsingContext,
   ChromiumBidi,
+  Network,
   type NetworkEvent,
 } from '../../../protocol/protocol.js';
 import {assert} from '../../../utils/assert.js';
 import {Deferred} from '../../../utils/Deferred.js';
-import {LogType, type LoggerFn} from '../../../utils/log.js';
+import {type LoggerFn, LogType} from '../../../utils/log.js';
 import type {CdpTarget} from '../cdp/CdpTarget.js';
 import type {EventManager} from '../session/EventManager.js';
 
 import type {NetworkStorage} from './NetworkStorage.js';
 import {
-  computeHeadersSize,
-  bidiNetworkHeadersFromCdpNetworkHeaders,
-  cdpToBiDiCookie,
-  cdpFetchHeadersFromBidiNetworkHeaders,
-  cdpAuthChallengeResponseFromBidiAuthContinueWithAuthAction,
   bidiBodySizeFromCdpPostDataEntries,
-  networkHeaderFromCookieHeaders,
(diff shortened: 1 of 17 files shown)


src/bidiMapper/modules/network/NetworkRequest.ts
```diff
@@ -304,26 +304,66 @@ export class NetworkRequest {
...
       // TODO: Verify this is correct
       // https://source.chromium.org/chromium/chromium/src/+/main:net/base/load_timing_info.h;l=145
-      fetchStart: getTiming(this.#response.info?.timing?.requestTime),
-      dnsStart: getTiming(this.#response.info?.timing?.dnsStart),
-      dnsEnd: getTiming(this.#response.info?.timing?.dnsEnd),
-      connectStart: getTiming(this.#response.info?.timing?.connectStart),
-      connectEnd: getTiming(this.#response.info?.timing?.connectEnd),
-      tlsStart: getTiming(this.#response.info?.timing?.sslStart),
-      requestStart: getTiming(this.#response.info?.timing?.sendStart),
+      fetchStart: getTiming(
```
OrKoN, line 329: I think these have to be durations from timeOrigin

src/bidiMapper/modules/network/NetworkRequest.ts
```diff
@@ -304,26 +304,66 @@ export class NetworkRequest {
...
+    const responseTimeOffset = getTiming(
+      getTiming(this.#response.info?.timing?.requestTime) -
+        getTiming(this.#request.info?.timestamp),
+    );
+
     return {
       // TODO: Verify this is correct
-      timeOrigin: getTiming(this.#response.info?.timing?.requestTime),
-      requestTime: getTiming(this.#response.info?.timing?.requestTime),
+      timeOrigin: Math.round(getTiming(this.#request.info?.wallTime) * 1000),
```
OrKoN, line 319: I think timeOrigin is the time of the creation of the navigable. I think we should take the navigation request wallTime for that.

OrKoN:
I think failures are due to the ubuntu 24.04 rollout (we probably need this https://github.com/puppeteer/puppeteer/blob/main/.github/workflows/ci.yml#L138)

OrKoN (approved):

src/bidiMapper/modules/network/NetworkRequest.ts
sadym-chromium: I assume CDP `response.timing` is relative to the baseline CDP `response.timing.requestTime`. The `responseTimeOffset` represents the diff between request and response baseline. It's my speculations which ends up to a meaningful result.

src/bidiMapper/modules/network/NetworkRequest.ts
sadym-chromium: I'm not sure if we can reliably get the time of the creation of the navigable. Especially if it was created before the Mapper.

OrKoN: it is performance.timeOrigin in the web APIs

PR #2928, opened 2024-12-22
fix: regression in serialization

OrKoN:
Deeply nested nodes do not include values in CDP, thus, Object.hasOwn throws an error because only objects are allowed as the first arg.

Drive-by: improve types
Caused by: https://github.com/GoogleChromeLabs/chromium-bidi/pull/2622
Addresses: https://github.com/puppeteer/puppeteer/issues/13433
WPT test: https://github.com/web-platform-tests/wpt/pull/49821/files


sadym-chromium:
@OrKoN I extended an e2e test

sadym-chromium (approved):

sadym-chromium (approved):

PR #2954, opened 2025-01-07
feat: implement subscription IDs

OrKoN:
This PR implements refactoring to support subscription IDs. In-scope:

- Internal refactoring to align with the spec.
- Generation of subscription IDs.

To be addressed in follow-ups:
- unsubscribe by IDs
- unit tests should not assert only channels but the entire subscription struct.

Notable changes:

- removal of subscribe priority (the insertion order reflects the priority).
- fix in findTopLevelBrowsingContext to not return an id for non-existing contexts.
- clearer way to handle global events (that are not related to any given context).
- checks like isSubscribed should use specific events instead of module event names.


src/bidiMapper/modules/session/SubscriptionManager.ts
```diff
@@ -75,146 +74,104 @@ export function unrollEvents(
...
-          contextId,
-          channel,
-        ),
-        channel,
-      }))
-      .filter(({priority}) => priority !== null) as {
-      priority: number;
-      channel: BidiPlusChannel;
-    }[];
+    const channels = new Set<BidiPlusChannel>();
```
sadym-chromium, line 98: Note: the `set` preserves order.

src/bidiMapper/modules/session/SubscriptionManager.ts
```diff
@@ -232,164 +189,189 @@ export class SubscriptionManager {
...
+    );
+
+    const newSubscriptions: Subscription[] = [];
+    let found = false;
+    for (const subscription of this.#subscriptions) {
+      if (found || subscription.channel !== channel) {
+        newSubscriptions.push(subscription);
+        continue;
+      }
+      if (intersection(subscription.eventNames, eventNames).size === 0) {
```
sadym-chromium, line 254: Can you please add comment here?

src/bidiMapper/modules/session/SubscriptionManager.ts
```diff
@@ -232,164 +189,189 @@ export class SubscriptionManager {
...
+        // If it is non-global subscription, keep it as is because we
+        // are only unsubscribing globally.
+        if (subscription.topLevelTraversableIds.size !== 0) {
+          newSubscriptions.push(subscription);
+          continue;
+        }
+        const subscriptionEventNames = new Set(subscription.eventNames);
+        for (const eventName of eventNames) {
+          if (subscriptionEventNames.has(eventName)) {
+            eventNames.delete(eventName);
```
sadym-chromium, line 268: I'm failing to understand something: if we have 2 subscriptions with the same `eventName`, will not it be removed only from the first subscription?

src/bidiMapper/modules/session/SubscriptionManager.ts
```diff
@@ -232,164 +189,189 @@ export class SubscriptionManager {
...
+        }
+      } else {
+        // Skip global subscriptions.
+        if (subscription.topLevelTraversableIds.size === 0) {
+          newSubscriptions.push(subscription);
+          continue;
+        }
+        if (
+          intersection(eventNames, subscription.eventNames).size !==
+          eventNames.size
```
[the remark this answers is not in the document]
OrKoN: It matches the spec one to one.

src/bidiMapper/modules/session/SubscriptionManager.ts
sadym-chromium: Can you please add the following test:
1. Subscribe to (a, b, c).
2. Subscribe to (b, c, d).
3. Unsubscribe from (b).
4. Assert no subscriptions to `b` left.

src/bidiMapper/modules/session/SubscriptionManager.ts
```diff
@@ -232,164 +189,189 @@ export class SubscriptionManager {
...
+          subscription.eventNames = subscriptionEventNames;
+          newSubscriptions.push(subscription);
+        }
+      } else {
+        // Skip global subscriptions.
+        if (subscription.topLevelTraversableIds.size === 0) {
+          newSubscriptions.push(subscription);
+          continue;
+        }
+        if (
```
OrKoN, line 288:
```suggestion
        // If event name sets are not exactly the same,
        // it is not a match for unsubscribe.
        if (
```

tests/browsing_context/test_create.py
```diff
@@ -418,7 +418,9 @@ async def test_browsingContext_subscribe_to_contextCreated_emits_for_existing(
     assert resp == {
         'id': subscribe_command_id,
         'type': 'success',
-        'result': {}
+        'result': {
+            'subscription': ANY_STR,
```
sadym-chromium, line 422: nit
```suggestion
            'subscription': ANY_UUID,
```

src/bidiMapper/modules/session/SubscriptionManager.ts
OrKoN: Yeah, that case would not work yet.

OrKoN: Let's see maybe it will work now, I wonder if the same needs to be done for non-global branch.

src/bidiMapper/modules/session/SubscriptionManager.spec.ts
```diff
@@ -215,12 +180,54 @@ describe('SubscriptionManager', () => {
...
+          [],
+          SOME_CHANNEL,
+        );
+        subscriptionManager.unsubscribe([SOME_EVENT], [], SOME_CHANNEL);
+        expect(
+          subscriptionManager.getChannelsSubscribedToEvent(
+            SOME_EVENT,
+            SOME_CONTEXT,
+          ),
+        ).to.deep.equal([]);
```
sadym-chromium, line 201: 👍

src/bidiMapper/modules/session/SubscriptionManager.ts
```diff
@@ -232,164 +189,204 @@ export class SubscriptionManager {
...
+        if (subscription.topLevelTraversableIds.size === 0) {
+          newSubscriptions.push(subscription);
+          continue;
+        }
+        // If event name sets are not exactly the same,
+        // it is not a match for unsubscribe.
+        if (!equal(eventNames, subscription.eventNames)) {
+          newSubscriptions.push(subscription);
+          continue;
+        }
```
sadym-chromium, line 290: I'm not sure if this is correct. Does it mean that after subscribe(a, b), calling unsubscribe(a) does not have any effect? I don't think it matches to the specification.

src/bidiMapper/modules/session/SubscriptionManager.ts
OrKoN: Yeah, I think the other branch needs the same treatment.

src/bidiMapper/modules/session/SubscriptionManager.ts
OrKoN: Updated.

sadym-chromium (approved):

src/bidiMapper/modules/session/SubscriptionManager.ts
sadym-chromium: I wonder why unit test didn't catch this case

PR #2987, opened 2025-01-15
fix: multi headers reporting incorrectly

Lightning00Blade:
We need to do proper deduplication for Binary Headers so this can land, feel free to take over.

### DIFF
--- a/src/bidiMapper/modules/network/NetworkRequest.ts
+++ b/src/bidiMapper/modules/network/NetworkRequest.ts
@@ -29,6 +29,7 @@ import {
   type NetworkEvent,
 } from '../../../protocol/protocol.js';
 import {assert} from '../../../utils/assert.js';
+import {DefaultMap} from '../../../utils/DefaultMap.js';
 import {Deferred} from '../../../utils/Deferred.js';
 import {type LoggerFn, LogType} from '../../../utils/log.js';
 import type {CdpTarget} from '../cdp/CdpTarget.js';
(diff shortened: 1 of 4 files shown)


src/bidiMapper/modules/network/NetworkRequest.ts
```diff
@@ -258,7 +258,23 @@ export class NetworkRequest {
   get #requestHeaders(): Network.Header[] {
     let headers: Network.Header[] = [];
     if (this.#requestOverrides?.headers) {
-      headers = this.#requestOverrides.headers;
+      const headerMap = new Map<string, string[]>();
```
sadym-chromium, line 261:
```suggestion
      const headerMap = new DefaultMap<string, string[]>(()=>[]);
```

src/bidiMapper/modules/network/NetworkRequest.ts
```diff
@@ -258,7 +258,23 @@ export class NetworkRequest {
...
+      for (const header of this.#requestOverrides.headers) {
+        if (headerMap.has(header.name)) {
+          headerMap.get(header.name)?.push(header.value.value);
+        } else {
+          headerMap.set(header.name, [header.value.value]);
+        }
+      }
```
sadym-chromium, line 268:
```suggestion
      const headerMap = new DefaultMap<string, string[]>(()=>[]);
      for (const header of this.#requestOverrides.headers) {
        headerMap.get(header.name).push(header.value.value);
      }
```

sadym-chromium (approved):

PR #2998, opened 2025-01-17
fix: wait for context should respect already existing contexts

sadym-chromium:
This should address Puppeteer test failures: https://github.com/GoogleChromeLabs/chromium-bidi/actions/runs/12827546451/job/35769733285?pr=2993

Tested manually. 1000 times run Puppeteer test "[locator.spec] Locator Locator.prototype.filter should resolve as soon as the predicate matches" with and without the change. Problem is gone.


OrKoN (approved):

OrKoN:
should it be a `fix: `

PR #3029, opened 2025-01-24
fix: rely on frameSubtreeWillBeDetached to emit contextDestroyed events

OrKoN:


Lightning00Blade (commented):
Is this a breaking change due to this requiring newer version of Chrome?

OrKoN:
it is available since M130.

Lightning00Blade (approved):

PR #3035, opened 2025-01-24
ci: don't use always()

Lightning00Blade:
https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#always

### DIFF
--- a/.github/workflows/wpt.yml
+++ b/.github/workflows/wpt.yml
@@ -51,7 +51,7 @@ jobs:
     name: '[Required] WPT sink'
     needs: [wpt]
     runs-on: ubuntu-latest
-    if: always()
+    if: ${{ !cancelled() }}
     steps:
       - if: ${{ needs.wpt.result != 'success' }}
         run: 'exit 1'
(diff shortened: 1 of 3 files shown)


.github/workflows/wpt.yml
```diff
@@ -235,9 +234,6 @@ jobs:
     name: Commit updated expectations
     runs-on: ubuntu-latest
     needs: [update_expectations]
-    # Experiment.
-    # TODO: remove.
-    if: success() || failure()
```
sadym-chromium, line 240: Are you sure this job will work if the main WPT job failed?

Lightning00Blade: I think we can check the `update_expectations` that it succeed.

sadym-chromium (approved):

.github/workflows/wpt.yml
sadym-chromium: Apparently it does not: https://github.com/GoogleChromeLabs/chromium-bidi/actions/runs/13078541005/job/36496854875

Lightning00Blade: I will take a look.

Lightning00Blade: Fixed in https://github.com/GoogleChromeLabs/chromium-bidi/pull/3062

Issue #2983, opened 2025-01-15
Implement user context subscriptions 


OrKoN:
Spec: https://github.com/w3c/webdriver-bidi/pull/836

- [x] WPT
   - https://github.com/web-platform-tests/wpt/pull/50290
   - https://github.com/web-platform-tests/wpt/pull/50434
- [x] implementation
   - https://github.com/GoogleChromeLabs/chromium-bidi/pull/3039

OrKoN:
I think WPT tests cover all important use cases (let me know if anything was overlooked).

PR #3039, opened 2025-01-27
feat: implement user context subscriptions

OrKoN:
- Introduces UserContextStorage to abstract and re-use operations on user contexts. For now, only the get operation is moved. 

Issue https://github.com/GoogleChromeLabs/chromium-bidi/issues/2983

### DIFF
--- a/src/bidiMapper/modules/session/SubscriptionManager.ts
+++ b/src/bidiMapper/modules/session/SubscriptionManager.ts
@@ -17,6 +17,7 @@
 
 import type {BidiPlusChannel} from '../../../protocol/chromium-bidi.js';
 import {
+  type Browser,
   type BrowsingContext,
   ChromiumBidi,
   InvalidArgumentException,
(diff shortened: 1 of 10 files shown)


src/bidiMapper/modules/session/SubscriptionManager.ts
```diff
@@ -157,23 +159,34 @@ export class SubscriptionManager {
       return false;
     }
 
-    // global subscription.
-    if (subscription.topLevelTraversableIds.size === 0) {
-      return true;
-    }
+    // user context subscription.
+    if (subscription.userContextIds.size !== 0) {
+      if (!contextId) {
```
sadym-chromium, line 164: nit: `contextId` became opaque. Consider renaming to `browsingContextId`.

tests/session/test_subscription.py
```diff
@@ -69,6 +69,35 @@ async def test_subscribeWithoutContext_subscribesToEventsInAllContexts(
     assert resp["params"]["context"] == context_id
 
 
+@pytest.mark.asyncio
+async def test_subscribeUserContext(websocket, context_id, html):
```
sadym-chromium, line 73: it would be nice to have a test for a cusom user context as well. However we can delegate it to WPT tests, up to you

sadym-chromium (approved):
in general LGTM, but I have a few minor comments

tests/session/test_subscription.py
OrKoN: I suggest to use WPT, could you please review the first test in WPT https://github.com/web-platform-tests/wpt/pull/50290 ?

sadym-chromium (approved):

PR #3070, opened 2025-02-04
chore: small changes

Lightning00Blade:
Update TypeScript to target newer versions of ECMAScript - we run it in the browser so almost all feature are there.
Made use of these new things to remove the `[...Iterator]` syntax as the new one should have better performence.
Replaced the `divider` div from the template as there were small issue around hiding things using border-top does the same thing.


OrKoN (approved):
This runs in Node 18 right?

PR #3072, opened 2025-02-04
feat: support UserContext in PreloadScripts

Lightning00Blade:
Need to add Unit tests mayble.

### DIFF
--- a/src/bidiMapper/modules/script/PreloadScriptStorage.ts
+++ b/src/bidiMapper/modules/script/PreloadScriptStorage.ts
@@ -80,12 +80,14 @@ export class PreloadScriptStorage {
   }
 
   onCdpTargetCreated(targetId: string, userContext: Browser.UserContext) {
-    for (const script of [...this.#scripts].filter((script) => {
+    const scriptInUserContext = [...this.#scripts].filter((script) => {
+      // Global scripts
       if (!script.userContexts && !script.contexts) {
         return true;
       }
       return script.userContexts?.includes(userContext);
-    })) {
+    });
+    for (const script of scriptInUserContext) {
       script.targetIds.add(targetId);
     }
   }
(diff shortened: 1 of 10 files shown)


src/bidiMapper/modules/script/PreloadScriptStorage.ts
```diff
@@ -68,9 +62,31 @@ export class PreloadScriptStorage {
...
+  getPreloadScript(id: string): PreloadScript {
+    const script = [...this.#scripts].find((script) => script.id === id);
+    if (script === undefined) {
+      throw new NoSuchScriptException(`No preload script with id '${id}'`);
+    }
+    return script;
+  }
+
+  onCdpTargetCreated(targetId: string, userContext: Browser.UserContext) {
+    for (const script of [...this.#scripts].filter((script) => {
```
OrKoN, line 83: nit: let's filter before the loop for readability.

src/bidiMapper/modules/script/PreloadScriptStorage.ts
```diff
@@ -68,9 +62,31 @@ export class PreloadScriptStorage {
...
+      if (!script.userContexts && !script.contexts) {
+        return true;
+      }
+      return script.userContexts?.includes(userContext);
```
OrKoN, line 87: should not this use the same logic as the find with targetId method?

OrKoN (commented):
I think we better refactor for clarify with this feature. I think current PR while working is a bit too confusing.

src/bidiMapper/modules/script/PreloadScriptStorage.ts
Lightning00Blade: That is impossible as at this point the Target is not in the list of the Target that the PreloadScript need to run.

OrKoN (approved):
Alright, tests seem to be happy so let's land it

Lightning00Blade:
Add interception seems to be slowed down I would want to fix that before we land. 
Only on Mapper without ChromeDriver.

PR #3078, opened 2025-02-05
chore: fix run all

Lightning00Blade:
This script didn't run pure mapper at all.

### DIFF
--- a/tools/run-wpt.mjs
+++ b/tools/run-wpt.mjs
@@ -39,7 +39,7 @@ function usage() {
   log(
     `Usage:
       [BROWSER_BIN=<path, default: download pinned chrome>]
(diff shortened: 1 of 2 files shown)


tools/run-wpt.mjs
```diff
@@ -39,7 +39,7 @@ function usage() {
   log(
     `Usage:
       [BROWSER_BIN=<path, default: download pinned chrome>]
-      [CHROMEDRIVER=<true | default: false>]
+      [CHROMEDRIVER=<true | default: true>]
```
sadym-chromium, line 42:
```suggestion
      [CHROMEDRIVER=<false | default: true>]
```

sadym-chromium (approved):

PR #3104, opened 2025-02-10
chore: rely on `Page.frameStartedNavigating` CDP event

sadym-chromium:


sadym-chromium:
This PR is blocked for now by edge case with `window.open`, where navigation is started before the CDP target is throttled.

sadym-chromium:
https://github.com/GoogleChromeLabs/chromium-bidi/pull/3236

PR #3133, opened 2025-02-20
build: remove comments from esm script

sadym-chromium:
Required as a work-around for `keep-sorted` artifacts. [Example](https://crrev.com/c/6105891/1).

Keep license headers.

### DIFF
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -16,6 +16,7 @@
     "noUncheckedIndexedAccess": true,
     "noUnusedLocals": false,
     "noUnusedParameters": false,
+    "removeComments": true,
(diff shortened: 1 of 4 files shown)


tsconfig.base.json
```diff
@@ -16,6 +16,7 @@
     "noUncheckedIndexedAccess": true,
     "noUnusedLocals": false,
     "noUnusedParameters": false,
+    "removeComments": true,
```
OrKoN, line 19: does it keep license comments?

sadym-chromium: oh, not really, so another solution is required.

OrKoN (approved):

PR #3151, opened 2025-02-27
test: use sleep to mitigate a race condition

OrKoN:
Somehow if the test fails too fast the websocket connection to WebDriver server fails.


LiviuArln:
Maybe we should not be so eager to create the websocket. Returning a factory might be better. But this seems good enough for now.

LiviuArln (approved):

PR #3236, opened 2025-03-20
chore: rely on Page.frameStartedNavigating CDP event

sadym-chromium:
`Page.frameStartedNavigating` is emitted for new navigations, so we can rely on it.

Disabling Network domain is out of scope to reduce risks of the PR.
Cleaning up TargetEvents and CdpTarget is out of scope.

Addressing http://b/401500518

### DIFF
--- a/src/bidiMapper/modules/context/NavigationTracker.spec.ts
+++ b/src/bidiMapper/modules/context/NavigationTracker.spec.ts
@@ -122,7 +122,11 @@ describe('NavigationTracker', () => {
           initialNavigationId,
         );
 
-        navigationTracker.frameStartedNavigating(ANOTHER_URL, LOADER_ID);
+        navigationTracker.frameStartedNavigating(
+          ANOTHER_URL,
+          LOADER_ID,
+          'differentDocument',
+        );
 
         assertNavigationEvent(
           ChromiumBidi.BrowsingContext.EventNames.NavigationStarted,
@@ -140,7 +144,22 @@ describe('NavigationTracker', () => {
 
         assertNoNavigationEvents();
         assert.equal(navigation.url, ANOTHER_URL);
-        assert.equal(navigationTracker.url, ANOTHER_URL);
+        assert.equal(navigationTracker.url, INITIAL_URL);
+        assert.equal(
+          navigationTracker.currentNavigationId,
+          navigation.navigationId,
+        );
+
+        navigationTracker.frameNavigated(YET_ANOTHER_URL, LOADER_ID);
+
+        assertNavigationEvent(
+          ChromiumBidi.BrowsingContext.EventNames.NavigationCommitted,
+          navigation.navigationId,
+          YET_ANOTHER_URL,
+        );
+
+        assert.equal(navigation.url, YET_ANOTHER_URL);
+        assert.equal(navigationTracker.url, YET_ANOTHER_URL);
         assert.equal(
           navigationTracker.currentNavigationId,
           navigation.navigationId,
@@ -162,7 +181,11 @@ describe('NavigationTracker', () => {
 
     it('canceled by script-initiated navigation', async () => {
       const navigation = navigationTracker.createPendingNavigation(SOME_URL);
-      navigationTracker.frameStartedNavigating(ANOTHER_URL, LOADER_ID);
+      navigationTracker.frameStartedNavigating(
+        ANOTHER_URL,
+        LOADER_ID,
+        'differentDocument',
+      );
 
       assertNavigationEvent(
         ChromiumBidi.BrowsingContext.EventNames.NavigationStarted,
@@ -170,7 +193,11 @@ describe('NavigationTracker', () => {
         ANOTHER_URL,
       );
 
-      navigationTracker.frameRequestedNavigation(YET_ANOTHER_URL);
+      navigationTracker.frameStartedNavigating(
+        YET_ANOTHER_URL,
+        ANOTHER_LOADER_ID,
+        'differentDocument',
+      );
 
       assertNavigationEvent(
         ChromiumBidi.BrowsingContext.EventNames.NavigationFailed,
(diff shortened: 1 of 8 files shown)


sadym-chromium:
Puppeteer is failing, as a "[navigation.spec] navigation Page.goto should work when page calls history API in beforeunload" test is fixed.

src/bidiMapper/modules/context/NavigationTracker.ts
```diff
@@ -283,13 +301,10 @@ export class NavigationTracker {
   frameNavigated(url: string, loaderId: string, unreachableUrl?: string) {
     this.#logger?.(LogType.debug, `frameNavigated ${url}`);
 
-    if (
-      unreachableUrl !== undefined &&
-      !this.#loaderIdToNavigationsMap.has(loaderId)
-    ) {
-      // The navigation failed before started. Get or create pending navigation and fail
-      // it.
+    if (unreachableUrl !== undefined) {
```
Lightning00Blade, line 304: Nit: 
```suggestion
    if (unreachableUrl) {
```

src/bidiMapper/modules/context/NavigationTracker.ts
```diff
@@ -340,7 +357,7 @@ export class NavigationTracker {
     // one.
     const fragmentNavigation =
       this.#pendingNavigation !== undefined &&
```
Lightning00Blade, line 359: Nit:
```suggestion
      this.#pendingNavigation &&
```

src/bidiMapper/modules/context/NavigationTracker.ts
```diff
@@ -400,64 +411,48 @@ export class NavigationTracker {
...
-    this.#currentNavigation = navigation;
+  frameStartedNavigating(
+    url: string,
+    loaderId: string,
+    navigationType: string,
+  ) {
+    this.#logger?.(LogType.debug, `frameStartedNavigating ${url}, ${loaderId}`);
 
-    if (this.#pendingNavigation === navigation) {
+    if (![loaderId, undefined].includes(this.#pendingNavigation?.loaderId)) {
```
Lightning00Blade, line 423: Can we not use this pattern but instead have two checks? 
Also with the inversion at the start it makes it even harder to understand. 
It looks confusing and you need to create an Array object, which is extra overhead.

src/bidiMapper/modules/context/NavigationTracker.ts
```diff
@@ -400,64 +411,48 @@ export class NavigationTracker {
...
-   * In case of `beforeunload` handler, the pending navigation should be marked as started
-   * for consistency, as the `browsingContext.navigationStarted` should be emitted before
-   * user prompt.
-   */
-  beforeunload() {
-    this.#logger?.(LogType.debug, `beforeunload`);
+    pendingNavigation.isFragmentNavigation = [
+      'historySameDocument',
+      'sameDocument',
+    ].includes(navigationType);
```
Lightning00Blade, line 451: Can we extract this check and reuse it we always make this check for both exiting and pending navigation.

Lightning00Blade (commented):
Some minor comments on the implementation. 
Will LGTM after I review the test changes.

src/bidiMapper/modules/context/NavigationTracker.spec.ts
```diff
@@ -182,18 +209,31 @@ describe('NavigationTracker', () => {
         (await navigation.finished).eventName,
         NavigationEventName.NavigationFailed,
       );
-      assert.equal(navigationTracker.currentNavigationId, initialNavigationId);
+      // Current navigation is the not yet finished `YET_ANOTHER_URL` navigation.
+      assert.notEqual(
```
Lightning00Blade, line 213: If the navigation to `YET_ANOTHER_URL` is not finished why is this assert.notEqual .
Which Id is expected here, from which navigation.

sadym-chromium: the navigation to `YET_ANOTHER_URL` is the current one, but we don't have it's ID

src/bidiMapper/modules/context/NavigationTracker.ts
sadym-chromium: I don't think if correctly covers edge case of `unreachableUrl =""`

src/bidiMapper/modules/context/NavigationTracker.ts
sadym-chromium: Actually it's excessive, removed

src/bidiMapper/modules/context/NavigationTracker.ts
sadym-chromium: done

src/bidiMapper/modules/context/NavigationTracker.ts
sadym-chromium: done

src/bidiMapper/modules/context/NavigationTracker.spec.ts
sadym-chromium: updated comment

src/bidiMapper/modules/context/NavigationTracker.ts
Lightning00Blade: ACK

src/bidiMapper/modules/context/NavigationTracker.ts
```diff
@@ -400,66 +409,57 @@ export class NavigationTracker {
...
+    url: string,
+    loaderId: string,
+    navigationType: string,
+  ) {
+    this.#logger?.(LogType.debug, `frameStartedNavigating ${url}, ${loaderId}`);
 
-    if (this.#pendingNavigation === navigation) {
+    if (
+      this.#pendingNavigation &&
+      this.#pendingNavigation?.loaderId !== undefined &&
```
Lightning00Blade, line 423: Nit: You can omit the optional chain for `this.#pendingNavigation`

Lightning00Blade (approved):
LGTM, found 1 small nit.

src/bidiMapper/modules/context/NavigationTracker.ts
sadym-chromium: correct, but I find it slightly better for readability

PR #3248, opened 2025-03-24
feat: `browsingContext.downloadWillBegin` event

sadym-chromium:
Implement [`browsingContext.downoadWillBegin`](https://www.w3.org/TR/webdriver-bidi/#event-browsingContext-downoadWillBegin) event.

### DIFF
--- a/src/bidiMapper/modules/cdp/CdpTarget.ts
+++ b/src/bidiMapper/modules/cdp/CdpTarget.ts
@@ -167,10 +167,6 @@ export class CdpTarget {
         this.#cdpClient.sendCommand('Page.enable', {
           enableFileChooserOpenedEvent: true,
         }),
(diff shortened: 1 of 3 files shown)


src/bidiMapper/modules/cdp/CdpTarget.ts
```diff
@@ -167,6 +167,10 @@ export class CdpTarget {
         this.#cdpClient.sendCommand('Page.enable', {
           enableFileChooserOpenedEvent: true,
         }),
+        this.#cdpClient.sendCommand('Browser.setDownloadBehavior', {
```
Lightning00Blade, line 170: Do we need this for every CDP target, isn't the browser domain only on the Browser Target?

sadym-chromium: indeed. Fixed

Lightning00Blade (approved):

PR #3255, opened 2025-03-26
New WebDriver Bidi command bluetooth.disableSimulation

chengweih001:
Add a new WebDriver Bidi command bluetooth.disableSimulation for tearing down the Bluetooth simulation.


src/protocol/chromium-bidi.ts
```diff
@@ -121,6 +121,7 @@ export type Command = (
   | ExternalSpecCommand<WebDriverBidiPermissions.PermissionsCommand>
   | ExternalSpecCommand<WebDriverBidiBluetooth.Bluetooth.HandleRequestDevicePrompt>
   | ExternalSpecCommand<WebDriverBidiBluetooth.Bluetooth.SimulateAdapter>
+  | ExternalSpecCommand<WebDriverBidiBluetooth.Bluetooth.DisableSimulation>
```
sadym-chromium, line 124: FYI: we can avoid adding those commands manually after this PR is merged: https://github.com/WebBluetoothCG/web-bluetooth/pull/649

tests/bluetooth/test_emulation.py
```diff
@@ -62,6 +62,15 @@ async def setup_device(websocket, context_id):
             }
         })
 
+async def disable_simulation(websocket, context_id):
```
sadym-chromium, line 65: NIT: no need in this trivial helper

tests/bluetooth/test_emulation.py
```diff
@@ -95,6 +104,8 @@ async def test_simulate_create_adapter_twice(websocket, context_id, state_1,
             }
         })
 
+    await disable_simulation(websocket, context_id)
```
sadym-chromium, line 107: It looks like `disable_simulation` is added for cleaning up the bluetooth state. If so, it should be moved to the beginning of the test, so that failing test would not influence another tests.

tests/bluetooth/test_emulation.py
```diff
@@ -62,6 +62,15 @@ async def setup_device(websocket, context_id):
...
 
+async def disable_simulation(websocket, context_id):
+    await execute_command(
+        websocket, {
+            'method': 'bluetooth.disableSimulation',
+            'params': {
+                'context': context_id,
+            }
+        })
+
```
sadym-chromium, line 74: Please add the following tests:
* `bluetooth.disableSimulation` disables simulation (we should assert the simulation is really disabled)
* `bluetooth.disableSimulation` disables simulation only in the provided context (it will fail for now, so mark it as `pytest.xfail` until the simulation per context is supported)
* `bluetooth.disableSimulation` twice

tests/bluetooth/test_emulation.py
chengweih001: Done. Added a fixture with `autouse=true` to be used in every test.

tests/bluetooth/test_emulation.py
chengweih001: Done.

Thanks for the suggestion! I also updated those to remove `pytest.xfail` on old headless mode since old headless can work with bluetooth now (after https://crrev.com/c/6165347)

sadym-chromium (approved):

sadym-chromium:
@chengweih001 please note that at least one WPT wdspec test is needed as well for the new command

sadym-chromium:
@chengweih001 please run `npm run format` to fix pre-commit

chengweih001:
> @chengweih001 please note that at least one WPT wdspec test is needed as well for the new command

Yes, I will take that into account for the PR to WPT.

PR #3261, opened 2025-03-28
chore: hide Mapper tab in nodejs runner

sadym-chromium:
Run Mapper in a hidden target, unless `verbose` flag is set.

### DIFF
--- a/src/bidiServer/MapperCdpConnection.ts
+++ b/src/bidiServer/MapperCdpConnection.ts
@@ -147,7 +147,8 @@ export class MapperServerCdpConnection {
       'Target.createTarget',
       {
         url: 'about:blank#MAPPER_TARGET',
-        hidden: true,
+        hidden: !verbose,
+        background: true,
       } as any,
     );
 


src/bidiServer/MapperCdpConnection.ts
```diff
@@ -143,34 +143,22 @@ export class MapperServerCdpConnection {
...
-      {},
-    )) as Protocol.Target.GetTargetsResponse;
-    const mapperTabTargetId = targets.targetInfos.filter(
-      (target) => target.type === 'page',
-    )[0]!.targetId;
+    const {targetId: mapperTargetId} = await browserClient.sendCommand(
+      'Target.createTarget',
+      {
+        url: 'about:blank#MAPPER_TARGET',
+        hidden: !verbose,
```
sadym-chromium, line 150: Using verbose can be a bit unstable, as there can be issues reproducible only in hidden targets. If so, we can add another flag for running Mapper in a visible Tab if needed.

OrKoN (approved):

PR #3276, opened 2025-04-01
Autofill commands

theindra:
Implements the suggested autofill commands in the bidi protocol


sadym-chromium:
Could you please provide some links to WebDriver BiDi PR, spec or draft?

PR #3311, opened 2025-04-14
feat: implement `emulation.setGeolocationOverride`

sadym-chromium:
CDP [`Emulation.setGeolocationOverride`](https://chromedevtools.github.io/devtools-protocol/tot/Emulation/#method-setGeolocationOverride) allows emulation only of latitude, longitude and accuracy.


src/bidiMapper/CommandProcessor.ts
```diff
@@ -266,8 +266,8 @@ export class CommandProcessor extends EventEmitter<CommandProcessorEventsMap> {
       // Emulation module
       // keep-sorted start block=yes
       case 'emulation.setGeolocationOverride':
-        throw new UnknownErrorException(
-          `Method ${command.method} is not implemented.`,
+        return await this.#browsingContextProcessor.setGeolocationOverride(
```
Lightning00Blade, line 269: From my understanding we wanted to have a processor per module?
This breaks this assumption.

sadym-chromium: good point. Will do.

sadym-chromium: done

src/bidiMapper/modules/emulation/EmulationProcessor.ts
```diff
@@ -0,0 +1,145 @@
...
+          await context.cdpTarget.setGeolocationOverride(params.coordinates),
+      ),
+    );
+    return {};
+  }
+
+  /**
+   * Returns a list of top-level browsing contexts.
+   */
+  async #getRelatedTopLevelBrowsingContexts(
```
Lightning00Blade, line 89: Non blocking:
Can we maybe move this function the UserContextStorage? 
It has the `BrowsingContextStorage` so we can reuse the logic?

src/bidiMapper/modules/emulation/EmulationProcessor.ts
```diff
@@ -0,0 +1,145 @@
+/**
+ * Copyright 2021 Google LLC.
```
Lightning00Blade, line 2: Correct year.

Lightning00Blade (approved):

src/bidiMapper/modules/emulation/EmulationProcessor.ts
sadym-chromium: `UserContextStorage` does not have `BrowsingContextStorage`. It can be extended with one, but it seems to be an odd dependency.

PR #3359, opened 2025-04-30
feat: emulate geolocation positionUnavailable

sadym-chromium:
Spec is landed in https://github.com/w3c/webdriver-bidi/pull/911

### DIFF
--- a/src/bidiMapper/modules/emulation/EmulationProcessor.ts
+++ b/src/bidiMapper/modules/emulation/EmulationProcessor.ts
@@ -45,7 +45,7 @@ export class EmulationProcessor {
       );
     }
 
-    let geolocationOverride:
+    let geolocation:
       | Emulation.GeolocationCoordinates
       | Emulation.GeolocationPositionError
       | null = null;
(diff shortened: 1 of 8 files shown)


src/bidiMapper/modules/emulation/EmulationProcessor.ts
```diff
@@ -39,15 +39,40 @@ export class EmulationProcessor {
...
+        (params.coordinates?.altitudeAccuracy ?? null) !== null
+      ) {
+        throw new InvalidArgumentException(
+          'Geolocation altitudeAccuracy can be set only with altitude',
+        );
+      }
+
+      geolocationOverride = params.coordinates;
+    } else if ('error' in params) {
+      if (params.error.type !== 'positionUnavailable') {
```
Lightning00Blade, line 65: Isn't this handled by the parsers?

Lightning00Blade (approved):
LGTM, with one nit - we can even drop the Override suffix, as everything on the config is technically and override.

PR #3366, opened 2025-05-02
feat: support GATT connection simulation

chengweih001:
Support bluetooth.simulateGattConnectionResponse command and bluetooth.gattConnectionAttempted event.

This PR also moves some common utilities like setup and request device into one place for the Bluetooth emulation test.


tests/bluetooth/__init__.py
```diff
@@ -0,0 +1,108 @@
...
+
+from test_helpers import (execute_command, goto_url, send_JSON_command,
+                          subscribe, wait_for_event)
+
+HTML_SINGLE_PERIPHERAL = """
+<div>
+    <button id="bluetooth">bluetooth</button>
+    <script>
+        let device;
+        const options = {filters: [{name:"SomeDevice"}]};
```
sadym-chromium, line 25: nit: consider extracting `SomeDevice` to a const.

tests/bluetooth/__init__.py
```diff
@@ -0,0 +1,108 @@
...
+        document.getElementById('bluetooth').addEventListener('click', async () => {
```
sadym-chromium, line 26: Should it be a page script, or maybe it can be evaluated via `script.evaluate`? Then there would be no need in a custom page at all.

tests/bluetooth/test_gatt_connection.py
```diff
@@ -0,0 +1,67 @@
...
+
+import pytest
+import pytest_asyncio
+from test_helpers import (execute_command, send_JSON_command, subscribe,
+                          wait_for_event)
+
+from . import disable_simulation, setup_granted_device
+
+
+@pytest_asyncio.fixture(autouse=True)
```
sadym-chromium, line 24: Can this fixture be moved to `__init__.py`?

sadym-chromium (approved):
LGTM. The few comments I left are optional.

tests/bluetooth/__init__.py
chengweih001: Done. I removed the custom page and replace it with a new `request_device` utility function.

tests/bluetooth/test_gatt_connection.py
chengweih001: I tried, but it doesn't take effect. I did some searching and it seems to be suggested to have the fixture inside the test file rather than inside `__init__.py`.

PR #3372, opened 2025-05-05
fix: allow no frameId in dialog events

OrKoN:


src/bidiMapper/modules/context/BrowsingContextImpl.ts
```diff
@@ -695,7 +698,10 @@ export class BrowsingContextImpl {
     });
 
     this.#cdpTarget.cdpClient.on('Page.javascriptDialogOpening', (params) => {
-      if (this.id !== params.frameId) {
+      // Checking for `params.frameId` for comptaibility with Chrome
+      // versions that do not have a frameId. TODO: remove once
+      // https://crrev.com/c/6487891 is in stable.
+      if (params.frameId && this.id !== params.frameId) {
```
sadym-chromium, line 704: This will cause duplicated events for pages with same-process iFrames. We need to check if the parent's CDP client is different from the current one to avoid duplicates.

OrKoN: Something like this?

src/bidiMapper/modules/context/BrowsingContextImpl.ts
```diff
@@ -661,7 +661,19 @@ export class BrowsingContextImpl {
...
+        return;
+      }
+      if (
+        !params.frameId &&
+        this.#parentId &&
+        this.#cdpTarget.cdpClient !==
+          this.#browsingContextStorage.getContext(this.#parentId)?.cdpTarget
+            .cdpClient
+      ) {
```
sadym-chromium, line 676:
```suggestion
      ) {
        // If CDP event `Page.javascriptDialogClosed` does not have a frameId, this
        // heuristic emits the event only for top-level per-cdp target context, ignoring
        // the event for same-process iframes. So the event will be emitted only once per
        // CDP target. TODO: remove once https://crrev.com/c/6487891 is in stable.
```

sadym-chromium (approved):

src/bidiMapper/modules/context/BrowsingContextImpl.ts
```diff
@@ -695,7 +707,19 @@ export class BrowsingContextImpl {
...
```
sadym-chromium, line 722: consider adding the same comment in order not to forget it later.

PR #3425, opened 2025-05-20
feat: implement `emulation.setLocaleOverride`

sadym-chromium:
Spec is landed: https://github.com/w3c/webdriver-bidi/pull/924

### DIFF
--- a/tests/emulation/test_locale.py
+++ b/tests/emulation/test_locale.py
@@ -15,24 +15,47 @@
 
 import pytest
 import pytest_asyncio
-from test_helpers import execute_command
+from test_helpers import execute_command, goto_url
+
+LOCALES = ["de-DE", "es-ES", "fr-FR", "it-IT"]
 
 
 @pytest_asyncio.fixture
-async def initial_locale(websocket, context_id):
+async def default_locale(websocket, context_id):
+    """
+    Returns default locale.
+    """
     return await get_locale(websocket, context_id)
 
 
 @pytest.fixture
-def some_locale(initial_locale):
-    """Returns some locale which is not equal to initial one."""
-    return "es-ES" if initial_locale != "es-ES" else "it-IT"
+def some_locale(default_locale):
+    """
+    Returns some locale which is not equal to `default_locale` nor to
+    `another_locale`.
+    """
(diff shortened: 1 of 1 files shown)


src/bidiMapper/modules/emulation/EmulationProcessor.ts
```diff
@@ -180,3 +208,17 @@ export class EmulationProcessor {
     return [...new Set(result).values()];
   }
 }
+
+// Export for testing.
+export function isValidLocale(locale: string): boolean {
+  try {
+    new Intl.Locale(locale);
```
OrKoN, line 215: theoretically locales supported in Chrome and Node might go out of sync for Node consumers of BiDi. Don't we do any validation on the CDP side? Could we rely on it instead?

tests/emulation/test_locale.py
```diff
@@ -0,0 +1,137 @@
...
+    emulated_locale_1 = await get_locale(websocket, browsing_context_id_1)
+    assert emulated_locale_1 == some_locale
+
+    browsing_context_id_2 = await create_context(user_context_id)
+    emulated_locale_2 = await get_locale(websocket, browsing_context_id_2)
+    assert emulated_locale_2 == another_locale
+
+
+@pytest.mark.asyncio
+async def test_locale_per_browsing_context(websocket, context_id,
```
OrKoN, line 138: could you please add a test with OOPIFs?

OrKoN (approved):

src/bidiMapper/modules/emulation/EmulationProcessor.ts
sadym-chromium: User can set locale override for a user context without navigables. This happens if the user creates a new user context, close all the navigables in the default one and sets locale override for the default user context. At that moment we don't have any navigable to set locale into.

tests/emulation/test_locale.py
sadym-chromium: Good point. The problem is wider, so I'll address it altogether with geolocation and screen orientation in https://github.com/GoogleChromeLabs/chromium-bidi/issues/3532.

PR #3427, opened 2025-05-20
feat: implement `browsingContext.downloadFinished` event

sadym-chromium:
Spec: https://github.com/w3c/webdriver-bidi/pull/923


tools/run_local_http_server.py
```diff
@@ -36,3 +37,7 @@
   - hangs forever: {local_server_http.url_hang_forever()}
   - bad ssl: {local_server_bad_ssl.url_200()}
 """)
+
+while True:
```
OrKoN, line 41: curious why is it needed?

sadym-chromium: the script is expected to be run manually. Probably there is a more elegant solution, but this simply keeps the main thread running in order to keep the thread with the http server running as well.

OrKoN (approved):

PR #3446, opened 2025-05-27
test: reduce LocalHttpServer shutdown timeout

sadym-chromium:
It does not make sense for waiting for 5s after the test cases are finished.


OrKoN:
Could you add an explanation why it works and why previous approach does not? Having an HTTP method on a server to stop itself seems unusual

Issue #1610, opened 2023-12-07
Fix race condition in local http server in e2e tests


sadym-chromium:
In #1591 the flag `TrackingProtection3pcd` was disabled, because of some e2e tests started failing. Investigation showed that the tests are passing when some external addresses are used instead of one provided by local http server.

The problem is in the python http server, which is hang in case of OOPiF.

AC:
* Re-enable `TrackingProtection3pcd`.
* Add test verifying OOPiF is supported by local server.

OrKoN:
Probably the same issue https://github.com/GoogleChromeLabs/chromium-bidi/pull/1805#issuecomment-1954261564

OrKoN:
cc @Lightning00Blade this might be affecting network interception tests as well

PR #3496, opened 2025-06-17
test: test_browsingContext_navigateWaitComplete_oopif_navigated

sadym-chromium:
Test for https://github.com/GoogleChromeLabs/chromium-bidi/issues/1610


tests/browsing_context/test_navigate.py
```diff
@@ -227,6 +227,89 @@ async def test_browsingContext_navigateWaitComplete_navigated(
...
+            },
+            'type': 'event',
+        },
+        {
+            'method': 'browsingContext.domContentLoaded',
+            'params': {
+                'context': Not(context_id),
+                'navigation': ANY_UUID,
+                'timestamp': ANY_TIMESTAMP,
+                'url': '',
```
OrKoN, line 293: where is this event coming from?

tests/browsing_context/test_navigate.py
```diff
@@ -227,6 +227,89 @@ async def test_browsingContext_navigateWaitComplete_navigated(
...
+            },
+            'type': 'event',
+        },
+        {
+            'method': 'browsingContext.domContentLoaded',
+            'params': {
+                'context': Not(context_id),
```
OrKoN: let's get the iframe's context id and assert using it?

tests/browsing_context/test_navigate.py
sadym-chromium: When the frame is moved to OOPIF, the CDP `Page.lifecycleEvent` events `commit', and `DOMContentLoaded` are emitted for the initial navigation.

OrKoN: IIRC this should not be emitted from the WebDriver BiDi point of view? let's add a comment that this is potentially incorrect

sadym-chromium: done

tests/browsing_context/test_navigate.py
sadym-chromium: done

OrKoN (approved):

PR #3511, opened 2025-06-20
test: add a server with a valid SSL

sadym-chromium:
This should ease testing cases where secure context is required, e.g. permissions.


tests/conftest.py
```diff
@@ -118,6 +130,8 @@ async def create_session(connection):
             "webSocketUrl": True,
             "goog:chromeOptions": {
                 "args": [
+                    # Required for navigating to `local_server_good_ssl`.
+                    "--ignore-certificate-errors-spki-list=0BFjPjhH1jzif+9C8nnl+d94xL0i/PK6o1CJnqnHKps=,0Rt4mT6SJXojEMHTnKnlJ/hBKMBcI4kteBlhR1eTTdk=",
```
alexnj: nit: move this to a config (or a variable)?

tests/tools/ssl_good_cert.key
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
alexnj: I'm assuming these these certs are self signed. Can they (keys, cert pems) be generated during test time instead of storing in repo? If these are not self signed, then they likely shouldn't be stored this way.

alexnj: Updating per the offline chat: Let's put a README.md in the ssl cert dir with steps we used to get public key/cert from WPT. Instructions would make it easier for the future maintainer to keep them updated/maintained.

alexnj (approved):
LGTM % nits/changes required.

tests/conftest.py
sadym-chromium: done

tests/tools/ssl_good_cert.key
sadym-chromium: > I'm assuming these these certs are self signed. Can they (keys, cert pems) be generated during test time instead of storing in repo? If these are not self signed, then they likely shouldn't be stored this way.

They are self-signed, but generating them certificates for each test run creates unnecessary overhead.

> Updating per the offline chat: Let's put a README.md in the ssl cert dir with steps we used to get public key/cert from WPT. Instructions would make it easier for the future maintainer to keep them updated/maintained.

Added `tests/tools/certs/README.md`.

Issue #3532, opened 2025-06-27
Support emulation in OOPiF


sadym-chromium:
Currently, emulations are not apply to OOPiF and are reset when an iFrame is moved out of process. Applies for geolocation, screen orientation and locale emulations.

Lightning00Blade:
@sadym-chromium From the PR this seems fixed, close if so.

PR #3534, opened 2025-06-27
test: test emulations in iframes

sadym-chromium:
Repro for https://github.com/GoogleChromeLabs/chromium-bidi/issues/3532


tests/conftest.py
```diff
@@ -626,7 +626,8 @@ def html(content="", same_origin=True):
 def iframe():
     """Return a factory for <iframe> with the given src."""
     def iframe(src=""):
-        return f'<iframe src="{src}" />'
+        # Geolocation is required for Geolocation testing in iframes.
+        return f'<iframe allow="geolocation *" src="{src}" />'
```
alexnj, line 630: Should we parameterize enabling geolocation vs. enabling it for all? Is this function going to be reused elsewhere or for negative tests?

sadym-chromium: it can be parametrized, but we use the iframes indirectly via fixture [`iframe_id`](https://github.com/GoogleChromeLabs/chromium-bidi/blob/e310753565dbd079a2d05f008daa92362a571cf9/tests/conftest.py#L636), which creates the iframe before the test run, so I believe it's safe to add all the `allow` in the fixture by default.

PR #3546, opened 2025-07-03
feat: implement network data collectors

sadym-chromium:
Start implementation of network collectors.


src/bidiMapper/modules/network/NetworkStorage.ts
```diff
@@ -53,6 +61,12 @@ export class NetworkStorage {
   /** A map from intercept ID to track active network intercepts. */
   readonly #intercepts = new Map<Network.Intercept, NetworkInterception>();
 
+  readonly #collectors = new Map<string, NetworkCollector>();
+  readonly #collectedResponses = new Map<
+    Network.Request,
+    {data: Script.GetDataResult; collectors: Set<string>}
```
OrKoN: I think we should not keep the data on the chromium-bidi side and always call into CDP

src/bidiMapper/modules/network/NetworkStorage.ts
```diff
@@ -202,10 +216,133 @@ export class NetworkStorage {
...
+      data: response,
+      collectors: new Set(collectors.map((collector) => collector.collectorId)),
+    });
+  }
+
   getInterceptionStages(browsingContextId: BrowsingContext.BrowsingContext) {
     const stages = {
       request: false,
-      response: false,
+      // In CDP the collectors work only on the intercepted requests.
```
OrKoN: I do not think this is true.

src/bidiMapper/modules/network/NetworkStorage.ts
```diff
@@ -202,10 +216,133 @@ export class NetworkStorage {
...
+    );
+    return [...collectors.values()];
+  }
+
+  async collectResponse(
+    request: NetworkRequest,
+    collectors: NetworkCollector[],
+  ) {
+    const result = await request.cdpTarget.cdpClient.sendCommand(
+      'Fetch.getResponseBody',
```
OrKoN: I think we should be using https://chromedevtools.github.io/devtools-protocol/tot/Network/#method-getResponseBody

src/bidiMapper/modules/network/NetworkStorage.ts
sadym-chromium: I referred to [`Fetch.getResponseBody`](https://chromedevtools.github.io/devtools-protocol/tot/Fetch/#method-getResponseBody), however the `Network` has a different scope.

OrKoN: Right, we should not be using Fetch and request interception for this feature.

src/bidiMapper/modules/network/NetworkStorage.ts
sadym-chromium: yup. This is not supported in Fetch, but is supported in Network domain.

sadym-chromium: Switched to Network

src/bidiMapper/modules/network/NetworkStorage.ts
sadym-chromium: Switched to Network

sadym-chromium:
> I believe we should not be using request interception to avoid performance regressions for this use case and we should not be caching data on the chromium-bidi side as it will be done on the browser side.

Switched to Network

src/bidiMapper/modules/cdp/CdpTarget.ts
```diff
@@ -521,7 +521,11 @@ export class CdpTarget {
 
   async toggleNetwork() {
     const stages = this.#networkStorage.getInterceptionStages(this.topLevelId);
-    const fetchEnable = Object.values(stages).some((value) => value);
+    const collectors = this.#networkStorage.getCollectorsForBrowsingContext(
+      this.topLevelId,
+    );
+    const fetchEnable =
```
OrKoN: fetch does not need to be enabled for collectors to work, I think the change here needs to be reverted

src/bidiMapper/modules/network/NetworkProcessor.ts
```diff
@@ -467,6 +471,69 @@ export class NetworkProcessor {
...
+          throw new InvalidArgumentException(
+            `Data collectors are available only on top-level browsing contexts`,
+          );
+        }
+      }
+    }
+    const collectorId = this.#networkStorage.addDataCollector(params);
+
+    // Enable fetch domain on impacted CDP targets.
+    await Promise.all(
```
OrKoN: currently, the Network is unconditionally enabled, we can keep this code but inside toggleNetwork we do not need to turn on Fetch

src/bidiMapper/modules/network/NetworkRequest.ts
```diff
@@ -555,6 +556,20 @@ export class NetworkRequest {
     if (event.responseStatusCode || event.responseErrorReason) {
       this.#response.paused = event;
 
+      this.#logger?.(
```
OrKoN: is this log generally useful?

src/bidiMapper/modules/network/NetworkStorage.ts
OrKoN: Comment outdated?

src/bidiMapper/modules/network/NetworkStorage.ts
```diff
@@ -327,7 +475,11 @@ export class NetworkStorage {
     this.#requests.set(request.id, request);
   }
 
-  deleteRequest(id: Network.Request) {
+  disposeRequest(id: Network.Request) {
+    if (this.#requestCollectors.get(id)?.size ?? 0 > 0) {
+      // Keep request is collected. Keep it, as it's data can be accessed later.
+      return;
+    }
```
OrKoN, line 473: let's add a TODO here to call CDP once we have a command to remove a specific request

OrKoN (changes requested):

src/bidiMapper/modules/network/NetworkProcessor.ts
sadym-chromium: Yup,I already addressed it.

src/bidiMapper/modules/network/NetworkStorage.ts
sadym-chromium: done

src/bidiMapper/modules/network/NetworkRequest.ts
sadym-chromium: nope. Removed.

src/bidiMapper/modules/cdp/CdpTarget.ts
sadym-chromium: yes, removed this part.

OrKoN (approved):
LGTM thanks. I see that there is still no WPT?

PR #3606, opened 2025-07-25
test: open extension page

sadym-chromium:
Verify http://b/412926721 is not reproducible on the protocol level.

### DIFF
--- a/tests/webExtension/test_web_extension.py
+++ b/tests/webExtension/test_web_extension.py
@@ -107,7 +107,7 @@ async def test_extensions_can_open_manifest(websocket,
                                             unpacked_extension_location,
                                             test_headless_mode,
                                             create_context):
-    """ Verify the extension page can be opened. http://b/412926721 """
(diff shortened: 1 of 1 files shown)


tests/webExtension/test_web_extension.py
```diff
@@ -95,6 +95,42 @@ async def test_extensions_can_install(websocket, unpacked_extension_location,
...
+        'args':
+            ['--enable-unsafe-extension-debugging', '--remote-debugging-pipe']
+    },
+}],
+                         indirect=True)
+async def test_extensions_can_open_manifest(websocket,
+                                            unpacked_extension_location,
+                                            test_headless_mode,
+                                            create_context):
+    """ Verify the extension page can be opened. http://b/412926721 """
```
Lightning00Blade: Let's use crbugs when the issue is on the Chromium tracker.
```suggestion
    """ Verify the extension page can be opened. https://crbug.com/412926721 """
```

PR #3633, opened 2025-08-07
fix: do not override context config with undefined

sadym-chromium:
Addressing Puppeteer failures: https://github.com/puppeteer/puppeteer/actions/runs/16803966784/job/47591715262.

### DIFF
--- a/src/bidiMapper/modules/browser/ContextConfig.ts
+++ b/src/bidiMapper/modules/browser/ContextConfig.ts
@@ -44,4 +44,27 @@ export class ContextConfig {
   // Timezone is kept in CDP format with GMT prefix for offset values.
   timezone?: string | null;
   userPromptHandler?: Session.UserPromptHandler;
+
+  /**
+   * Merges multiple `ContextConfig` objects. The configs are merged in the
+   * order they are provided. For each property, the value from the last config
+   * that defines it (i.e., the value is not `undefined`) will be used.
+   * The final result will not contain any `undefined` properties.
+   */
+  static merge(...configs: (ContextConfig | undefined)[]): ContextConfig {
+    const result: ContextConfig = {};
+
+    for (const config of configs) {
+      if (!config) {
+        continue;
+      }
+      for (const key in config) {
+        const value = config[key as keyof ContextConfig];
+        if (value !== undefined) {
+          (result as any)[key] = value;
+        }
+      }
+    }
+    return result;
+  }
 }
--- a/src/bidiMapper/modules/browser/ContextConfig.ts
+++ b/src/bidiMapper/modules/browser/ContextConfig.ts
@@ -52,7 +52,7 @@ export class ContextConfig {
    * The final result will not contain any `undefined` properties.
    */
   static merge(...configs: (ContextConfig | undefined)[]): ContextConfig {
-  const result = new ContextConfig()
+    const result = new ContextConfig();
 
     for (const config of configs) {
       if (!config) {
--- a/src/bidiMapper/modules/browser/ContextConfig.ts
+++ b/src/bidiMapper/modules/browser/ContextConfig.ts
@@ -52,7 +52,7 @@ export class ContextConfig {
    * The final result will not contain any `undefined` properties.
    */
   static merge(...configs: (ContextConfig | undefined)[]): ContextConfig {
-    const result: ContextConfig = {};
+  const result = new ContextConfig()
 
     for (const config of configs) {
       if (!config) {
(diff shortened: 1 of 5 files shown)


src/bidiMapper/modules/browser/ContextConfigStorage.ts
```diff
@@ -23,13 +23,14 @@ export class ContextConfigStorage {
...
   updateGlobalConfig(config: ContextConfig) {
-    this.#global = {...this.#global, ...config};
+    this.#global = {...this.#global, ...this.#removeUndefined(config)};
   }
 
   updateBrowsingContextConfig(
     browsingContextId: string,
     config: ContextConfig,
   ) {
+    config = this.#removeUndefined(config);
```
OrKoN: how do we get undefined in the first place? and why does it matter? removal of undefines on a strictly typed data structure does not sound good

sadym-chromium: The `undefined` can be set here. And if so, it will override the global setting (but it should not): https://github.com/GoogleChromeLabs/chromium-bidi/blob/c66d125b659c704242111570782cb20afb4b3f75/src/bidiMapper/modules/browser/BrowserProcessor.ts#L114-L115

Another approach would be to implement `mergeConfig` method ignoring the `undefined`. WDYT?

OrKoN: yeah it should not override the global setting if undefined is passed. I think that would be a better solution.

src/bidiMapper/modules/browser/ContextConfig.ts
```diff
@@ -43,4 +44,27 @@ export class ContextConfig {
...
   userPromptHandler?: Session.UserPromptHandler;
+
+  /**
+   * Merges multiple `ContextConfig` objects. The configs are merged in the
+   * order they are provided. For each property, the value from the last config
+   * that defines it (i.e., the value is not `undefined`) will be used.
+   * The final result will not contain any `undefined` properties.
+   */
+  static merge(...configs: (ContextConfig | undefined)[]): ContextConfig {
+    const result: ContextConfig = {};
```
OrKoN:
```suggestion
  const result = new ContextConfig()
```

src/bidiMapper/modules/browser/ContextConfigStorage.ts
```diff
@@ -17,57 +17,80 @@
...
+        : this.#userContextConfigs.get(userContext);
+
+    const browsingContextConfig =
+      topLevelBrowsingContextId === undefined
+        ? undefined
+        : this.#browsingContextConfigs.get(topLevelBrowsingContextId);
+
+    return ContextConfig.merge(
+      this.#global,
+      userContextConfig,
```
OrKoN:
```suggestion
      this.#userContextConfigs.get(userContext),
```

src/bidiMapper/modules/browser/ContextConfigStorage.ts
```diff
@@ -17,57 +17,80 @@
...
+      browsingContextConfig,
```
OrKoN:
```suggestion
      this.#browsingContextConfigs.get(topLevelBrowsingContextId),
```

src/bidiMapper/modules/browser/ContextConfigStorage.ts
```diff
@@ -17,57 +17,80 @@
...
-    return result;
+    const userContextConfig =
+      userContext === undefined
+        ? undefined
+        : this.#userContextConfigs.get(userContext);
+
+    const browsingContextConfig =
+      topLevelBrowsingContextId === undefined
+        ? undefined
+        : this.#browsingContextConfigs.get(topLevelBrowsingContextId);
```
OrKoN:
```suggestion
```

src/bidiMapper/modules/browser/ContextConfigStorage.ts
sadym-chromium: `userContextConfig` can be `undefined`, so it wouldn't work unless we declare `#userContextConfigs` as `Map<string|undefined, ContextConfig>()` to simplify this construction.

src/bidiMapper/modules/browser/ContextConfigStorage.ts
sadym-chromium: the same as with `userContextConfig`

src/bidiMapper/modules/browser/ContextConfigStorage.ts
sadym-chromium: done

src/bidiMapper/modules/browser/ContextConfigStorage.ts
OrKoN: right, why are the arguments to this function optional? make it required + add getGlobalConfig getter without merging?

sadym-chromium: done

PR #3662, opened 2025-08-22
feat!: validate `browser.createUserContext:proxy`

sadym-chromium:
### Blocked on https://github.com/w3c/webdriver/issues/1920

Spec: https://github.com/w3c/webdriver-bidi/pull/988

**Breaking change:** `proxy` parameter does not allow for schema.


OrKoN (approved):

sadym-chromium:
> It looks like it would not be compatible with Puppeteer?

It is compatible but it's a breaking change. In Puppeteer, we will have to parse the proxy string and strip out the schema from  proxies per schema.

Issue #3698, opened 2025-09-05
Sending locale override multiple times fails


OrKoN:
The error

```
Another locale override is already in effect
```

can be emitted and block the target.

PR #3699, opened 2025-09-05
fix: handle target config errors gracefully

OrKoN:
There is no reason to fail target workflow if there was some error on the CDP level. For troubleshooting logs should be used and testing should rely on observable effects.

Related: https://github.com/GoogleChromeLabs/chromium-bidi/issues/3698

### DIFF
--- a/src/bidiMapper/modules/cdp/CdpTarget.ts
+++ b/src/bidiMapper/modules/cdp/CdpTarget.ts
@@ -201,67 +201,59 @@ export class CdpTarget {
    * Enables all the required CDP domains and unblocks the target.
    */
   async #unblock() {
-    try {
-      await Promise.all([
-        this.#cdpClient.sendCommand('Page.enable', {
-          enableFileChooserOpenedEvent: true,
-        }),
-        ...(this.#ignoreFileDialog()
-          ? []
-          : [
-              this.#cdpClient.sendCommand(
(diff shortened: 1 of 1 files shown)


src/bidiMapper/modules/cdp/CdpTarget.ts
```diff
@@ -254,15 +254,11 @@ export class CdpTarget {
...
-      this.#logger?.(LogType.debugError, 'Failed to unblock target', error);
-      // The target might have been closed before the initialization finished.
-      if (!this.#cdpClient.isCloseError(error)) {
-        this.#unblocked.resolve({
-          kind: 'error',
-          error,
-        });
-        return;
-      }
+      this.#logger?.(
```
Lightning00Blade: Let move to Promise.allSettled and throw after so that we know the other thing are already set up.

OrKoN: we do not need to throw here. Like 50% of thse calls throw  see https://github.com/GoogleChromeLabs/chromium-bidi/issues/3702

Lightning00Blade (approved):

src/bidiMapper/modules/cdp/CdpTarget.ts
Lightning00Blade: Sorry, I meant to say log.

PR #3701, opened 2025-09-05
fix: do no duplicate headers

OrKoN:


src/bidiMapper/modules/network/NetworkRequest.ts
```diff
@@ -908,16 +908,14 @@ export class NetworkRequest {
...
-      // ),
-    ];
+    const cdpHeaders = this.#response.info?.headers ?? {};
+    const cdpRawHeaders = this.#response.extraInfo?.headers ?? {};
+    for (const [key, value] of Object.entries(cdpRawHeaders)) {
+      cdpHeaders[key] = value;
+    }
+    const headers = bidiNetworkHeadersFromCdpNetworkHeaders(cdpHeaders);
+
+    // TODO: Verify how to dedupe this.#response.paused?.responseHeaders.
```
Lightning00Blade: Nit: the comment needs to be update to and moved before the const headers.

Lightning00Blade (approved):

PR #3722, opened 2025-09-12
test: locale emulation in popup

sadym-chromium:
Repro for https://crbug.com/444674636

### DIFF
--- a/tests/emulation/test_locale.py
+++ b/tests/emulation/test_locale.py
@@ -117,8 +117,7 @@ async def test_locale_set_and_clear(websocket, context_id, default_locale,
 async def test_locale_set_in_opened_window(websocket, context_id,
                                            default_locale, some_locale,
                                            another_locale):
-    pytest.xfail(
-        "https://github.com/GoogleChromeLabs/chromium-bidi/issues/3698")
+    pytest.xfail("crbug.com/444674636")
     await execute_command(
         websocket, {
             "method": "session.subscribe",


tests/emulation/test_locale.py
```diff
@@ -102,6 +113,73 @@ async def test_locale_set_and_clear(websocket, context_id, default_locale,
...
+            "params": {
+                "events": ["browsingContext.contextCreated"]
+            }
+        })
+
+    await execute_command(
+        websocket, {
+            'method': 'emulation.setLocaleOverride',
+            'params': {
+                'contexts': [context_id],
```
OrKoN, line 133: let's set the override for the user context instead

tests/emulation/test_locale.py
```diff
@@ -102,6 +113,73 @@ async def test_locale_set_and_clear(websocket, context_id, default_locale,
...
+            }
+        })
+
+    # Get the new context id.
+    new_context_id = resp["result"]["value"]["context"]
+
+    # Assert the locale emulation inherited.
+    assert (await get_locale(websocket, new_context_id)) == some_locale
+
+    # Assert the locale can be overridden.
```
OrKoN, line 159: here, let's check that the popup event is actually emitted and the locale in the new page is set

sadym-chromium: Which pop event are your referring to? Isn't it what is done a few lines above:
```
    # Assert the locale emulation inherited.
    assert (await get_locale(websocket, new_context_id)) == some_locale
```

tests/emulation/test_locale.py
sadym-chromium: It would be another use-case.

tests/emulation/test_locale.py
sadym-chromium: UPD: I updated the assertion above to make sure the emulation is not inherited (it is failing now).

OrKoN: I am referring to the `browsingContext.contextCreated` which in Puppeteer and Playwright APIs is called a `popup` when the context created is a window opened by another window.

tests/emulation/test_locale.py
OrKoN: That what was the issue about https://github.com/GoogleChromeLabs/chromium-bidi/issues/3698

OrKoN: See the linked https://github.com/GoogleChromeLabs/chromium-bidi/pull/3699

tests/emulation/test_locale.py
sadym-chromium: `window.open` returns the window proxy with it's ID, so no need in waiting for the event.

OrKoN: no, that is not equivalent: you get the ID from serialization of the window. But the problem is actually in CdpTarget unblocking mechanism which might never unblock the target on errors or wrong implementation.

sadym-chromium: > no, that is not equivalent: you get the ID from serialization of the window. But the problem is actually in CdpTarget unblocking mechanism which might never unblock the target on errors or wrong implementation.

Updated repro link: https://crbug.com/444674636

PR #3782, opened 2025-09-25
feat: respect `network.addDataCollector: maxEncodedDataSize` param

sadym-chromium:
Align data collectors with BiDi spec. Do not mark requests exceeding the collector's `maxEncodedDataSize` as collected.

### DIFF
--- a/src/bidiMapper/modules/network/CollectorsStorage.ts
+++ b/src/bidiMapper/modules/network/CollectorsStorage.ts
@@ -31,6 +31,10 @@ import type {NetworkRequest} from './NetworkRequest.js';
 
 type NetworkCollector = Network.AddDataCollectorParameters;
 
+// The default total data size limit in CDP.
+// https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/inspector/inspector_network_agent.cc;drc=da1f749634c9a401cc756f36c2e6ce233e1c9b4d;l=133
+const MAX_TOTAL_COLLECTED_SIZE = 200 * 1000 * 1000;
+
 export class CollectorsStorage {
   readonly #collectors = new Map<string, NetworkCollector>();
   readonly #requestCollectors = new Map<Network.Request, Set<string>>();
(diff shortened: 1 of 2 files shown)


src/bidiMapper/modules/network/CollectorsStorage.ts
```diff
@@ -38,6 +41,16 @@ export class CollectorsStorage {
...
 
   addDataCollector(params: Network.AddDataCollectorParameters) {
+    if (
+      params.maxEncodedDataSize < 1 ||
+      params.maxEncodedDataSize > 200 * 1000 * 1000
+    ) {
+      // 200 MB is the default limit in CDP:
+      // https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/inspector/inspector_network_agent.cc;drc=da1f749634c9a401cc756f36c2e6ce233e1c9b4d;l=133
+      throw new InvalidArgumentException(
+        `Max encoded data size should be between 1 and ${200 * 1000 * 1000}`,
```
OrKoN:
```suggestion
        `Max encoded data size should be between 1 and ${200_000_000}`,
```

src/bidiMapper/modules/network/CollectorsStorage.ts
```diff
@@ -38,6 +41,16 @@ export class CollectorsStorage {
   }
 
   addDataCollector(params: Network.AddDataCollectorParameters) {
+    if (
+      params.maxEncodedDataSize < 1 ||
+      params.maxEncodedDataSize > 200 * 1000 * 1000
```
OrKoN: Let's extract into a const and use `200_000_000`

OrKoN: the name for the const could be `MAX_TOTAL_DATA_SIZE`

OrKoN (approved):

src/bidiMapper/modules/network/CollectorsStorage.ts
sadym-chromium: done

PR #3797, opened 2025-09-29
fix: don't report interception for DataUrl and servedFromCache

Lightning00Blade:
Closes: https://github.com/GoogleChromeLabs/chromium-bidi/issues/3762

### DIFF
--- a/src/bidiMapper/modules/network/NetworkRequest.ts
+++ b/src/bidiMapper/modules/network/NetworkRequest.ts
@@ -197,6 +197,15 @@ export class NetworkRequest {
     return this.url.startsWith('data:');
   }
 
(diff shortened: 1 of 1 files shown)


src/bidiMapper/modules/network/NetworkRequest.ts
```diff
@@ -197,6 +197,15 @@ export class NetworkRequest {
     return this.url.startsWith('data:');
   }
 
+  #isNotInterpretable(): boolean {
```
OrKoN:
```suggestion
  #isNonInterceptable(): boolean {
```
?

OrKoN (approved):

PR #3809, opened 2025-10-02
feat: support request data collection

sadym-chromium:
Step 1. Throw unsupported exception for request collector.


src/bidiMapper/modules/network/CollectorsStorage.ts
```diff
@@ -84,7 +94,12 @@ export class CollectorsStorage {
     return true;
   }
 
-  disown(requestId: Network.Request, collectorId?: string) {
+  isRequestCollected() {
```
Lightning00Blade: If you want to have this let's do the correct types so 

```suggestion
  isRequestCollected(_requestId: Network.Request, _collectorId?: string) {
```

Lightning00Blade (approved):
LGTM with one comment.

PR #3815, opened 2025-10-06
feat: support request post data collection

sadym-chromium:


src/bidiMapper/modules/network/CollectorsStorage.ts
```diff
@@ -138,36 +160,60 @@ export class CollectorsStorage {
       // Collector is aimed for a different top-level browsing context.
       return false;
     }
-    if (collector.maxEncodedDataSize < request.bytesReceived) {
+    if (!collector.dataTypes.includes(dataType)) {
```
Lightning00Blade, line 163: I guess the current implementation is to have `dataTypes` as Array?
Does a change to Set make sense there?

src/bidiMapper/modules/network/NetworkStorage.ts
```diff
@@ -271,25 +256,47 @@ export class NetworkStorage {
...
+        );
+    }
+
+    if (params.disown && params.collector !== undefined) {
+      this.#collectorsStorage.disownData(
+        request.id,
+        params.dataType,
+        params.collector,
       );
+      // `disposeRequest` disposes request only if no other collectors for it are left.
```
Lightning00Blade, line 284: Nit: I would prefer this be a TSDoc comment on the method, or at least copy it there as well

Lightning00Blade (approved):

src/bidiMapper/modules/network/CollectorsStorage.ts
sadym-chromium: `collector` is stored as a raw BiDi params. From the performance perspective, the `Array.includes` is slightly faster on the small arrays (not more then 2 in our case) then `Set.has`.

src/bidiMapper/modules/network/NetworkStorage.ts
sadym-chromium: added a JSDOC on `disposeRequest`

src/bidiMapper/modules/network/CollectorsStorage.ts
Lightning00Blade: I see, interesting I didn't know about the performance parts

PR #3819, opened 2025-10-07
feat: implement `emulation.setNetworkConditions:offline`

sadym-chromium:


src/bidiMapper/modules/cdp/CdpTarget.ts
```diff
@@ -904,4 +911,35 @@ export class CdpTarget {
       userAgent: userAgent ?? '',
     });
   }
+
+  async setEmulatedNetworkConditions(
+    networkConditions: Emulation.NetworkConditions | null,
+  ): Promise<void> {
+    if (networkConditions !== null && networkConditions.type !== 'offline') {
```
Lightning00Blade, line 918: Isn't this verified in the Parser?
And that will run first so it will be `invalid param exception`

src/bidiMapper/modules/emulation/EmulationProcessor.ts
```diff
@@ -351,12 +351,65 @@ export class EmulationProcessor {
...
+      this.#contextConfigStorage.updateGlobalConfig({
+        emulatedNetworkConditions: params.networkConditions,
+      });
+    }
+
+    if (
+      params.networkConditions !== null &&
+      params.networkConditions.type !== 'offline'
+    ) {
+      throw new UnsupportedOperationException(
```
Lightning00Blade, line 393: Same here is this not handle in the parser?

src/bidiMapper/modules/cdp/CdpTarget.ts
```diff
@@ -904,4 +911,35 @@ export class CdpTarget {
...
+      throw new UnsupportedOperationException(
```
Lightning00Blade, line 919: Is this not handled by the Param parser in zod?

src/bidiMapper/modules/cdp/CdpTarget.ts
sadym-chromium: It will until the new condition will be added to the BiDi protocol. It his condition is removed, after the new condition is added, Mapper will consider it as removing condition instead of failing with `Unsupported exception`. So this is required for future extensions of BiDi protocol.

sadym-chromium: So the problem is that we pass the raw protocol `Emulation.NetworkConditions` type here, which can be unexpectedly extended. And we want to be aware of this extension.

src/bidiMapper/modules/emulation/EmulationProcessor.ts
sadym-chromium: Replied above

src/bidiMapper/modules/cdp/CdpTarget.ts
sadym-chromium: > Is this not handled by the Param parser in zod?

Replied in https://github.com/GoogleChromeLabs/chromium-bidi/pull/3819#discussion_r2413605705

src/bidiMapper/modules/cdp/CdpTarget.ts
sadym-chromium: I consider the concern is addressed and resolving it in order to merge. Feel free to un-resolve and discuss.

PR #3832, opened 2025-10-10
feat: validate headers set in `network.setExtraHeaders`

sadym-chromium:


src/bidiMapper/modules/network/NetworkProcessor.ts
```diff
@@ -629,13 +629,74 @@ function unescapeURLPattern(pattern: string) {
   return result;
 }
 
+// https://fetch.spec.whatwg.org/#header-name
+const FORBIDDEN_HEADER_NAME_SYMBOLS = [
```
OrKoN: let's use a set for both lists

sadym-chromium: done

PR #3844, opened 2025-10-15
test: test_network_collector_get_data_response_image

sadym-chromium:
Try to repro the scenario @jimevans mentioned, when the chrome crashed on collected response data retrieval of an image.


sadym-chromium:
The problem was in the websocket's message size limit.

PR #3851, opened 2025-10-16
feat: get request body size from Content-Length

sadym-chromium:
Intended to be used instead of `hasPostData` CDP flag in Puppeteer.

### DIFF
--- a/tests/network/test_network.py
+++ b/tests/network/test_network.py
@@ -13,6 +13,8 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 #
+import json
+
 import pytest
 from anys import ANY_DICT, ANY_LIST, ANY_NUMBER, ANY_STR, AnyOr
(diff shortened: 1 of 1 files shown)


src/bidiMapper/modules/network/NetworkRequest.ts
```diff
@@ -245,16 +245,45 @@ export class NetworkRequest {
...
   }
 
+  #getBodySizeFromHeaders(
+    headers: Protocol.Network.Headers | undefined,
+  ): number | undefined {
+    if (headers === undefined) {
+      return undefined;
+    }
+
+    if (headers['Content-Length'] !== undefined) {
```
OrKoN, line 255: IIRC content-length is optional so might not be a good signal

sadym-chromium: Right, it can be omitted in case of `Transfer-Encoding: chunked`. But having `Content-Length` headers IS a good signal.

PR #3854, opened 2025-10-17
fix: data collection do not influence redirect network events

sadym-chromium:
Force-creating requests for redirects.


src/bidiMapper/modules/network/NetworkStorage.ts
```diff
@@ -97,7 +97,8 @@ export class NetworkStorage {
     redirectCount?: number,
   ): NetworkRequest {
     let request = this.getRequestById(id);
-    if (request) {
+    if (redirectCount === undefined && request) {
+      // Force re-creating requests for redirects.
```
sadym-chromium, line 101: Otherwise if the original request was collected, the new request will not be created, and no events will be emitted.

PR #3902, opened 2025-11-10
test: measure screenshot performance

sadym-chromium:
Step 1: just upload performance test results to artifacts.

Analogue of Selenium performance tests.

### Out of scope
* Results presentation.
* Exclude performance tests from normal e2e runs.


tools/run-e2e.mjs
```diff
@@ -194,6 +194,7 @@ if (HEADLESS === 'false' && !argv.k) {
 if (argv.k) {
   e2eArgs.push('-k', argv.k);
 }
+if (argv.s) e2eArgs.push('-s');
```
Lightning00Blade: Style here is wrong, is this file not picked up by the linter?

Lightning00Blade (approved):
LGTM after all the actions are ran again.

tools/run-e2e.mjs
sadym-chromium: `ignores: ['**/*.spec.ts', 'src/bidiServer/**', 'tools/**']`

sadym-chromium: formatted manually

Lightning00Blade: Can I get a link I only see 1 rule disabled for this this should be Prettier that handles this, not EsLint

sadym-chromium: Oh, `npm format` considers this is a valid string: `if (verbose) console.log('!!@@##');`

PR #3941, opened 2025-12-11
fix: ensure `network.beforeRequestSent` is emitted before `network.authRequired`

sadym-chromium:
Addressing https://github.com/puppeteer/puppeteer/issues/14499

### DIFF
--- a/src/bidiMapper/modules/network/NetworkStorage.spec.ts
+++ b/src/bidiMapper/modules/network/NetworkStorage.spec.ts
@@ -403,6 +403,35 @@ describe('NetworkStorage', () => {
       events = await getEvents('network.authRequired');
       expect(events).to.have.length(2);
     });
+
+    it('should emit beforeRequestSent before authRequired', async () => {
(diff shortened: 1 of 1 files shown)


src/bidiMapper/modules/network/NetworkRequest.ts
```diff
@@ -689,6 +692,9 @@ export class NetworkRequest {
       this.#fetchId !== this.id
     ) {
       this.#interceptPhase = Network.InterceptPhase.AuthRequired;
+      // Make sure the `network.beforeRequestSent` is emitted before
+      // `network.authRequired`.
+      this.#emitEventsIfReady();
```
OrKoN, line 697: Let's add a unit test

OrKoN: (to NetworkStorage.spec.ts)

sadym-chromium: done

PR #3952, opened 2025-12-16
feat: implement `emulation.setTouchOverride`

sadym-chromium:
Spec: https://www.w3.org/TR/webdriver-bidi/#commands-emulationsettouchoverride


src/bidiMapper/modules/context/BrowsingContextImpl.ts
```diff
@@ -1974,6 +1974,14 @@ export class BrowsingContextImpl {
     );
   }
 
+  async setTouchOverride(maxTouchPoints: number | null) {
+    await Promise.all(
```
OrKoN: I think we should do await Promise.allSettled here since for some targets the command will error and we probably do not want to return early.

OrKoN: this probably applies to other methods

OrKoN (approved):

src/bidiMapper/modules/context/BrowsingContextImpl.ts
OrKoN: if we want to throw on errors we would need to do it manually, but not throwing is fine I think

sadym-chromium: done for this command

PR #3970, opened 2026-01-08
fix: remove sameParty Cookie attribute

bershanskiy:
Chrome never shipped support for HTTP Set-Cookie directive and Cookie.sameParty APIs. This commit removes code for mapping `Cookie.sameParty` values which were never fully supported.

## Background
- In 2022, [Chrome deprecated](https://github.com/WICG/first-party-sets/issues/92) `SameParty` cookie attribute (in all contexts, including HTTP Set-Cookie header and all JS APIs).
- In 2023, [CL 4219615](https://chromium-review.googlesource.com/c/chromium/src/+/4219615) removed `sameParty` attributes from `Cookie` and `CookieInit` JS APIs  and setting corresponding internal Boolean to `false` ([diff](https://chromium.googlesource.com/chromium/src/+/1b2855445a38ce9258cdbffde8bfcd0150ec12bf%5E%21/)).
- In 2024 Chromium BiDi Mapper PR #1759 added mapping for `sameParty` and in tests remarked that miraculously [`sameParty` is not set for whatever reason](https://github.com/GoogleChromeLabs/chromium-bidi/pull/1759/files#diff-bb27db29812eb6dd20a6902cf9e7ab0efe6022a736f6c9df08c9c333ef3f7d3cR250).

### DIFF
--- a/src/bidiMapper/modules/network/NetworkUtils.ts
+++ b/src/bidiMapper/modules/network/NetworkUtils.ts
@@ -201,7 +201,6 @@ export function cdpToBiDiCookie(
   // Extending with CDP-specific properties with `goog:` prefix.
   result[`goog:session`] = cookie.session;
   result[`goog:priority`] = cookie.priority;
-  result[`goog:sameParty`] = cookie.sameParty;
   result[`goog:sourceScheme`] = cookie.sourceScheme;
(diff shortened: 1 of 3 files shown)


sadym-chromium:
merged in https://github.com/GoogleChromeLabs/chromium-bidi/pull/3980

PR #3976, opened 2026-01-13
chore: update pre-commit

sadym-chromium:
and re-format python files


wpt
OrKoN, line 1: did you mean to include wpt?

sadym-chromium: not really, reverted. Thanks!

PR #4002, opened 2026-01-22
ci: measure perf via Puppeteer

sadym-chromium:
Another approach for performance measurement. Evaluate script changing DOM many times via Puppeteer CDP and BiDi. Normalize the results by CDP to negotiate hardware differences.

I keep the e2e for now, so that we can compare the stability of the metrics.

### DIFF
--- a/tools/benchmark-puppeteer.mjs
+++ b/tools/benchmark-puppeteer.mjs
@@ -22,6 +22,7 @@ import puppeteer from 'puppeteer';
 import {installAndGetChromePath} from './path-getter/path-getter.mjs';
 
 const ITERATIONS = 10000;
+const WARMUP_ITERATIONS = ITERATIONS * 0.1;
 
 function calculateStats(latencies) {
   latencies.sort((a, b) => a - b);
@@ -88,7 +89,6 @@ async function runBenchmark(name, launchOptions) {
         </div>`;
     }, name);
 
-    const WARMUP_ITERATIONS = ITERATIONS * 0.1;
     console.log(`Warming up for ${WARMUP_ITERATIONS} iterations...`);
     for (let i = 0; i < WARMUP_ITERATIONS; i++) {
       await page.evaluate(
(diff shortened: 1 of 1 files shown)


tools/perf-pup.mjs
```diff
@@ -0,0 +1,174 @@
...
+    `BiDi is ${Math.abs(diff).toFixed(4)}ms (${Math.abs(diffPercent).toFixed(4)}%) ${slowerOrFaster} than CDP`,
+  );
+  console.log(
+    `95% Confidence Interval of the difference: [${ciLowerPercent.toFixed(2)}%, ${ciUpperPercent.toFixed(2)}%] (±${moeDiffPercent.toFixed(2)}%)`,
+  );
+
+  console.log(`PERF_METRIC:diff_bidi_vs_cdp:${diffPercent.toFixed(4)}`);
+}
+
+run().catch(console.error);
```
Lightning00Blade: Should this exit silently? I think we want to fail the run if there are errors, no?

tools/benchmark-puppeteer.mjs
```diff
@@ -0,0 +1,174 @@
...
+      document.body.innerHTML = `
+        <div style='font-family:Segoe UI, sans-serif; padding:20px; background:#f4f7f6;'>
+          <h2>${browserName} Protocol Benchmark</h2>
+          <div style='display:flex; gap:15px;'>
+            <div id='some-box' style='flex:1; padding:15px; background:white; border-left:5px solid #3498db;'>Some counter<div id='some-counter' style='font-size:24px;'>0</div><div id='some-res'>-</div></div>
+          </div>
+        </div>`;
+    }, name);
+
+    const WARMUP_ITERATIONS = ITERATIONS * 0.1;
```
Lightning00Blade: Nit: give this is const you can extract it next to the Iterations.

tools/benchmark-puppeteer.mjs
```diff
@@ -0,0 +1,174 @@
...
+  });
+
+  try {
+    const page = await browser.newPage();
+    await page.goto('about:blank');
+
+    // Setup DOM
+    await page.evaluate((browserName) => {
+      document.body.innerHTML = `
+        <div style='font-family:Segoe UI, sans-serif; padding:20px; background:#f4f7f6;'>
```
Lightning00Blade, line 84: Let's add the `<!DOCTYPE html>` So the file is not executed in Quriks mode

Lightning00Blade: Sorry we are setting this on the about:blank, so I think it should be fine.

tools/benchmark-puppeteer.mjs
```diff
@@ -0,0 +1,174 @@
+#!/usr/bin/env node
```
Lightning00Blade, line 1: Super Nit: I don't think there is a reason to shorten the file name, it makes it harder to know what it is.

sadym-chromium: done

tools/benchmark-puppeteer.mjs
sadym-chromium: resolving

tools/benchmark-puppeteer.mjs
sadym-chromium: done

tools/perf-pup.mjs
sadym-chromium: agree. Removed.

PR #4004, opened 2026-01-23
ci: improve Puppeteer benchmark

sadym-chromium:
1. Re-run browser for more stable results.
2. Changed the test script.
3. As long as the metric mechanism is changed, the metric name changed as well to prevent confusion.

### DIFF
--- a/tools/benchmark-puppeteer.mjs
+++ b/tools/benchmark-puppeteer.mjs
@@ -116,9 +116,7 @@ async function runBenchmarkRun(name, launchOptions, chromePath) {
   try {
     const page = await browser.newPage();
     await page.goto('about:blank');
-    await page.evaluate((html) => {
-      document.body.innerHTML = html;
-    }, BENCHMARK_HTML);
+    await page.setContent(BENCHMARK_HTML);
 
     // Warmup
     for (let i = 0; i < WARMUP_ITERATIONS; i++) {


tools/benchmark-puppeteer.mjs
```diff
@@ -62,13 +106,9 @@ function calculateStats(latencies) {
...
-async function runBenchmark(name, launchOptions) {
-  const executablePath = installAndGetChromePath();
-  console.log(`\n=== Testing Browser: ${name} ===`);
-  console.log(`Using Chrome: ${executablePath}`);
-
+async function runBenchmarkRun(name, launchOptions, chromePath) {
   const browser = await puppeteer.launch({
-    executablePath,
+    executablePath: chromePath,
     args: ['--no-sandbox', '--disable-setuid-sandbox'],
```
OrKoN: we should not use these args.

tools/benchmark-puppeteer.mjs
```diff
@@ -62,13 +106,9 @@ function calculateStats(latencies) {
...
     headless: true,
```
OrKoN, line 112: I would suggest headless shell here for more stable results

tools/benchmark-puppeteer.mjs
```diff
@@ -99,72 +131,118 @@ async function runBenchmark(name, launchOptions) {
...
+  const stats = {
+    cdp: [],
+    bidi: [],
+  };
+
+  for (let i = 0; i < RUNS; i++) {
+    process.stdout.write(`Run ${i + 1}/${RUNS}: `);
+
+    // Run CDP.
+    process.stdout.write(`running CDP... `);
```
Lightning00Blade, line 184: Why not just console.log?

Lightning00Blade: Pref one style of logging, or even better a function to abstract this.

sadym-chromium: `process.stdout.write` allows for string appending, while console.log does not. This allows for a clean output.

tools/benchmark-puppeteer.mjs
sadym-chromium: done

tools/benchmark-puppeteer.mjs
sadym-chromium: done

tools/benchmark-puppeteer.mjs
```diff
@@ -62,34 +106,21 @@ function calculateStats(latencies) {
...
-  const executablePath = installAndGetChromePath();
-  console.log(`\n=== Testing Browser: ${name} ===`);
-  console.log(`Using Chrome: ${executablePath}`);
-
+async function runBenchmarkRun(name, launchOptions, chromePath) {
   const browser = await puppeteer.launch({
-    executablePath,
-    args: ['--no-sandbox', '--disable-setuid-sandbox'],
-    headless: true,
+    executablePath: chromePath,
```
OrKoN, line 111: why do we need executable path? puppeteer finds the default one. It is going to override the shell param if the path is not shell

OrKoN: If you run npm install puppeteer@24.36.0 you do not need any path resolution stuff we do we chromium-bidi that only uses @puppeteer/browsers

sadym-chromium: We might need canary functionality e.g. when we use it during startup. If so, the puppeteer's default browser is outdated. Also this would allow us for detecting regressions due to Chromium rolls.

OrKoN: I believe you are currently not running with shell due to this.

sadym-chromium: oh, you are right. But the shell fails with
```
Run 1/100: running CDP... running BiDi... node_modules/puppeteer-core/lib/esm/puppeteer/common/CallbackRegistry.js:99
    #error = new ProtocolError();
             ^
ProtocolError: Protocol error (Browser.close): 'Browser.close' wasn't found`
```

OrKoN: Interesting, we actually do not run CI for bidi with shell, so probably a bug

sadym-chromium: Let's run it with headless and switch to shell later on

tools/benchmark-puppeteer.mjs
```diff
@@ -62,34 +106,21 @@ function calculateStats(latencies) {
...
-    args: ['--no-sandbox', '--disable-setuid-sandbox'],
+    executablePath: chromePath,
     headless: true,
     ...launchOptions,
   });
 
   try {
     const page = await browser.newPage();
     await page.goto('about:blank');
+    await page.evaluate((html) => {
```
OrKoN: await page.setContent(html)?

tools/benchmark-puppeteer.mjs
OrKoN: looks like a bug in chromium-bidi

OrKoN: https://gist.github.com/OrKoN/d85abf050c9cd71c86ad0abf2c63e990 log for later

tools/benchmark-puppeteer.mjs
sadym-chromium: done

tools/benchmark-puppeteer.mjs
OrKoN: fix https://github.com/GoogleChromeLabs/chromium-bidi/pull/4005

PR #4005, opened 2026-01-23
fix: handle uncaught exceptions in close method

OrKoN:
Added error handling for uncaught exceptions when closing the browser.

### DIFF
--- a/src/bidiMapper/modules/browser/BrowserProcessor.ts
+++ b/src/bidiMapper/modules/browser/BrowserProcessor.ts
@@ -54,7 +54,10 @@ export class BrowserProcessor {
     // Ensure that it is put at the end of the event loop.
     // This way we send back the response before closing the tab.
     // Always catch uncaught exceptions.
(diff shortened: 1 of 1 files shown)


src/bidiMapper/modules/browser/BrowserProcessor.ts
```diff
@@ -53,7 +53,8 @@ export class BrowserProcessor {
   close(): EmptyResult {
     // Ensure that it is put at the end of the event loop.
     // This way we send back the response before closing the tab.
-    setTimeout(() => this.#browserCdpClient.sendCommand('Browser.close'), 0);
+    // Always catch uncaught exceptions.
+    setTimeout(() => this.#browserCdpClient.sendCommand('Browser.close').catch(() => {}), 0);
```
OrKoN:
```suggestion
    setTimeout(
      () => this.#browserCdpClient.sendCommand('Browser.close').catch(() => {}),
      0,
    );
```

PR #4012, opened 2026-01-26
ci: extend performance metrics

sadym-chromium:


package.json
```diff
@@ -227,6 +227,7 @@
   },
   "dependencies": {
     "mitt": "^3.0.1",
+    "puppeteer": "^24.36.0",
```
OrKoN, line 230: overlooked this, it should not be a prod dependency

PR #4041, opened 2026-01-30
chore: benchmark.md

sadym-chromium:


docs/benchmark.md
```diff
@@ -0,0 +1,151 @@
+# Performance Benchmarks for Chromium-BiDi
```
Lightning00Blade, line 1: Let's put such files under docs/ 
It make is easier to find and keeps the root cleaner
WDYT?

sadym-chromium: done

PR #4068, opened 2026-02-13
test: extend puppeteer benchmark

sadym-chromium:
Measure evaluating of `1` and 100 DOM nodes

### DIFF
--- a/tools/benchmark-puppeteer.mjs
+++ b/tools/benchmark-puppeteer.mjs
@@ -115,6 +115,31 @@ const benchmarks = [
       await page.evaluate(() => 1);
     },
   },
+  {
+    id: 'nodes',
+    name: '100 Nodes Evaluation',
+    warmup: async (page) => {
+      await page.evaluate(() => {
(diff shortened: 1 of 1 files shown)


tools/benchmark-puppeteer.mjs
```diff
@@ -83,16 +103,44 @@ async function runBenchmarkRun(launchOptions, chromePath) {
...
+        for (let i = 0; i < 100; i++) {
+          const node = document.createElement('div');
+          node.innerText = `Node ${i}`;
+          container.appendChild(node);
+        }
+      });
+    },
+    measure: async (page) => {
+      await page.evaluate(() => {
+        return Array.from(document.querySelectorAll('#container > div'));
```
OrKoN: Puppeteer does not allow returning nodes from evaluate. $$eval should be used to get content from handles.

OrKoN: You can measure evaluateHandle for this

PR #4138, opened 2026-04-24
feat: switch to URLPattern

OrKoN:


src/bidiMapper/modules/network/NetworkProcessor.ts
```diff
@@ -653,25 +521,6 @@ export class NetworkProcessor {
 /**
  * See https://w3c.github.io/webdriver-bidi/#unescape-url-pattern
```
Lightning00Blade: Left over comment.

PR #4152, opened 2026-05-22
build: remove pre-commit

OrKoN:
I have spent some time to get pre-commit to install in the right pipenv but was not successful. It is anyway too heavy of a dependency so for python linting I replaced with ruff. Also using keep-sorted directly. It is also a little faster. The second commit is the caught linter issues.


Lightning00Blade (approved):

OrKoN:
I tried to minimize the changes but we were quite inconsistent in styles and I think the new linter is better at formatting python (looks nicer to me).

Lightning00Blade:
Yeah I agree that it looks better.

PR #4158, opened 2026-05-22
refactor: migrate from wireit to GN

OrKoN:
This PR migrates the build system to GN and Chromium's depot tools. Note that this is only the initial step and we are likely require many more changes to GN templates (remove hard-coded paths, support tests from gen folder, build.gn for e2e tests).


.github/workflows/e2e.yml
```diff
@@ -75,11 +75,15 @@ jobs:
         with:
           node-version-file: '.nvmrc'
           cache: npm
+          cache-dependency-path: 'package-lock.json'
```
Lightning00Blade: Why do we need this now, usally this is auto handled.

.github/workflows/playwright.yml
```diff
@@ -19,20 +19,26 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           path: 'bidi'
-      - name: Checkout
+      - name: Checkout Playwright
```
Lightning00Blade, line 22: Should we remove this similar to other 3rd party?

GEMINI.md
```diff
@@ -31,8 +31,8 @@ To restore the project to a known good state, run the following commands in orde
   extension.
 - They are co-located with the code they are testing.
 - Run unit tests with `npm run unit`. This command first compiles the TypeScript code
-  (including tests) into JavaScript in the `lib` directory and then runs the tests
-  using the Node.js native test runner.
+  (including tests) into JavaScript in the `out/Default/gen/src/` directory and then runs the tests
+  using Mocha.
```
Lightning00Blade: pls revert.

.github/workflows/playwright.yml
OrKoN: No, this does not add any dependencies and can still run on GitHub actions.

OrKoN: Although let us remove it, I am not 100% sure it runs and we rarely look at it. We should re-add it once the refactor settles and it is fairly easy to re-add. I will do it in a follow-up.

Lightning00Blade (approved):

build_overrides/BUILDCONFIG.gn
```diff
@@ -0,0 +1,52 @@
+# Copyright 2026 The Chromium Authors
```
OrKoN: TODO: update headers

PR #4170, opened 2026-06-02
fix: set mobile device metrics on Android targets

juliopiubello:
I want to support for BiDi viewport emulation through `browsingContext.setViewport for Android. while doing it, I found that Android targets need CDP mobile viewport semantics for the emulated viewport and DPR to match the actual Android layout behavior.

Android BiDi viewport emulation was using desktop CDP viewport semantics, which can produce incorrect viewport sizing after `browsingContext.setViewport`.

So I am pushing this PR to set the CDP mobile flag for android targets when BiDi applies viewport emulation

Android needs mobile viewport semantics for viewport and DPR overrides to match the target’s actual layout behavior. Desktop targets keep the existing `mobile: false` behavior.

## Testing

- `npm run gn:gen`
- `npm run build`
- `npm run unit`
- `npx eslint src/bidiMapper/modules/cdp/CdpTarget.ts src/bidiMapper/modules/cdp/CdpTarget.test.ts`
- `npx prettier --check package.json src/bidiMapper/modules/cdp/CdpTarget.ts src/bidiMapper/modules/cdp/CdpTarget.test.ts`

### DIFF
--- a/src/BUILD.gn
+++ b/src/BUILD.gn
@@ -174,6 +174,7 @@ ts_library("unit_tests") {
   sources = [
     "bidiMapper/modules/browser/BrowserProcessot.test.ts",
     "bidiMapper/modules/browser/ContextConfigStorage.test.ts",
+    "bidiMapper/modules/cdp/CdpTarget.test.ts",
     "bidiMapper/modules/context/BrowsingContextStorage.test.ts",
     "bidiMapper/modules/context/NavigationTracker.test.ts",
     "bidiMapper/modules/emulation/EmulationProcessor.test.ts",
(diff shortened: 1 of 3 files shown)


Lightning00Blade:
Hello,

Thanks for the PR!
Looking at the Chromium code I am surprised this is the case 
as the following two lines should just mean that we default to the original in case of `mobile: false`
- https://crsrc.org/c/content/browser/devtools/protocol/emulation_handler.cc;l=756;drc=77010a16c40f098e9a0d9022eed415ad8e491540;bpv=1;bpt=1?q=EmulatedScreenType::kMobile&ss=chromium%2Fchromium%2Fsrc

- https://crsrc.org/c/third_party/blink/renderer/core/inspector/dev_tools_emulator.cc;l=313;drc=77010a16c40f098e9a0d9022eed415ad8e491540;bpv=1;bpt=1 

In general it may be more appropriate to fix (if there is an issue) upstream so CDP user also benefit for it.

juliopiubello:
Thanks @Lightning00Blade ! it makes sense

PR #4180, opened 2026-06-12
build: add a tool to apply wpt metadata updates locally

OrKoN:
Currently we do not run automation with Github Actions so this PR adds a basic script to fetch expectations from PRs manually.

### DIFF
--- a/tools/apply-wpt-expectations.mjs
+++ b/tools/apply-wpt-expectations.mjs
@@ -1,0 +1,238 @@
+#!/usr/bin/env node
+
+/**
+ * Copyright 2026 Google LLC.
+ * Copyright (c) Microsoft Corporation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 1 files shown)


tools/apply-wpt-expectations.mjs
```diff
@@ -0,0 +1,238 @@
...
+      {encoding: 'utf-8'},
+    );
+    const runs = JSON.parse(runListOutput);
+    if (runs.length === 0) {
+      throw new Error(`No WPT workflow runs found for branch "${branch}"`);
+    }
+    const run = runs[0];
+    runId = run.databaseId;
+    console.log(
+      `Found run ID ${runId} (Status: ${run.status}, Conclusion: ${run.conclusion})`,
```
samiyac, line 88: Do we want to filter on run status before retrieving artifacts or does that not matter?

samiyac (approved):

tools/apply-wpt-expectations.mjs
OrKoN: the status should not matter (we need the latest run). In case it's wrong, we can pass exact run id as an arg.

PR #4183, opened 2026-06-22
test: use assert style instead of expect

Lightning00Blade:
The style provides better type safety and aligns more with the other project we maintain.


OrKoN:
I realized we lost some unit test coverage. I am working on a fix, let's land this after the fix.

OrKoN (approved):

PR #4189, opened 2026-06-24
build: add rdb reporter for unit tests

OrKoN:
The discrepancy in reported counts is due to test.ts files that define no tests which the default reporter prints and the rdb reporter filters out.

Drive-by: removed duplicate tests

### DIFF
--- a/src/bidiMapper/modules/emulation/EmulationProcessor.test.ts
+++ b/src/bidiMapper/modules/emulation/EmulationProcessor.test.ts
@@ -92,8 +92,6 @@ describe('EmulationProcessor helper functions', () => {
       'Latn',
       // Locale with script before language.
       'Latn-en',
-      // Repeated separator.
-      'en--US',
       // Invalid character in an otherwise valid structure.
       'en-US-!',
       // Too many subtags of a specific type (e.g., multiple script tags).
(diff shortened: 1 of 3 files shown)


src/bidiMapper/modules/emulation/EmulationProcessor.test.ts
```diff
@@ -93,7 +93,6 @@ describe('EmulationProcessor helper functions', () => {
       // Locale with script before language.
       'Latn-en',
       // Repeated separator.
```
Lightning00Blade: Left out comment.

tools/unit-resultsink-reporter.mjs
```diff
@@ -0,0 +1,161 @@
...
+
+      const testIdName = `${sourcePath} > ${fullName}`;
+      const testResult = {
+        testId: testIdName
+          .replace(/[\r\n]+/g, ' ')
+          .normalize('NFC')
+          .substring(0, 512),
+        status,
+        expected,
+        duration: `${durationSec}.${durationNanos.toString().padStart(9, '0')}s`,
```
Lightning00Blade: If the `durationNanos` is some float in JS this may exceed the Protobuf length expectation. So we should trim it down to an expected one 
We can also just do `(durationMs / 1000).toFixed(3)`. I don't think we need super great precision above MS

Lightning00Blade (approved):

PR #4193, opened 2026-07-01
build: add rdb reporter to e2e

OrKoN:


tests/resultsink_reporter.py
```diff
@@ -0,0 +1,117 @@
+# Copyright 2021 Google LLC.
```
Lightning00Blade: Wrong year?

Lightning00Blade (approved):

PR #4224, opened 2026-07-17
feat(digital-credentials): support non-top-level contexts in setVirtualWalletBehavior

mohamedamir:
This CL expands support for digitalCredentials.setVirtualWalletBehavior
to accept non-top-level contexts (subframes).

If a subframe does not have an explicitly configured virtual wallet behavior,
it falls back to the global configuration (or user context configuration),
rather than inheriting from its parent context. This is achieved by explicitly
sending the resolved behavior (which defaults to 'clear') to the subframe via
CDP to override Chromium's internal parent-inheritance behavior.

TAG=agy
CONV=ce3d0ebd-0f95-4b7d-ab74-fc357eac15ed

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,7 +23,7 @@
         "@types/sinon": "21.0.1",
         "@types/websocket": "1.0.10",
         "chai": "6.2.2",
-        "devtools-protocol": "0.0.1655900",
+        "devtools-protocol": "0.0.1662358",
         "eslint": "10.5.0",
         "eslint-config-prettier": "10.1.8",
         "eslint-import-resolver-typescript": "4.4.5",
@@ -1956,9 +1956,9 @@
       }
     },
     "node_modules/devtools-protocol": {
-      "version": "0.0.1655900",
-      "resolved": "https://registry.npmjs.org/devtools-protocol/-/devtools-protocol-0.0.1655900.tgz",
-      "integrity": "sha512-mgjlBJ7tTnTcIAks1HQKvd1XBRiZPzHd/HOW8MUr7XqsfZWtrBFdRG4XOM9wnrZ2bx6OLGWkbD6y5whjh36GgQ==",
+      "version": "0.0.1662358",
+      "resolved": "https://registry.npmjs.org/devtools-protocol/-/devtools-protocol-0.0.1662358.tgz",
+      "integrity": "sha512-2QEjq2DqI9JjRYImbJmhdpzcoibPEfpcv7sydS9S05KJurVudhyBYTVdXJZrHE45B1doI/xJbRP2JJ4Is/AhNw==",
       "dev": true,
       "license": "BSD-3-Clause"
     },
--- a/src/bidiMapper/modules/cdp/CdpTarget.ts
+++ b/src/bidiMapper/modules/cdp/CdpTarget.ts
@@ -23,6 +23,7 @@ import {
   type Browser,
   type BrowsingContext,
   type ChromiumBidi,
+  DigitalCredentials,
   Emulation,
   Session,
   type UAClientHints,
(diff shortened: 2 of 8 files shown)


package-lock.json
```diff
@@ -23,7 +23,7 @@
         "@types/sinon": "21.0.1",
         "@types/websocket": "1.0.10",
         "chai": "6.2.2",
-        "devtools-protocol": "0.0.1655900",
+        "devtools-protocol": "0.0.1662358",
```
OrKoN, line 26: you would need to upload node_modules to the gcs bucket (same as with chromium) as we are preparing for a move (you should have access, if you have Chromium access).

mohamedamir: Done (or at least I think so)

src/bidiMapper/modules/browser/ContextConfigStorage.ts
```diff
@@ -79,6 +79,22 @@ export class ContextConfigStorage {
     return this.#global;
   }
 
+  /**
+   * Returns the configuration for a specific browsing context.
+   */
+  getBrowsingContextConfig(
```
OrKoN: can we use getActiveConfig instead of adding new getters?

src/bidiMapper/modules/context/BrowsingContextImpl.ts
```diff
@@ -166,31 +172,34 @@ export class BrowsingContextImpl {
...
-              params: {
-                ...context.serializeToBidiValue(),
-                // Hack to provide the initial URL of the context, as it can be changed
-                // between the page target is attached and unblocked, as the page is not
-                // fully paused in MPArch session (https://crbug.com/372842894).
-                // TODO: remove once https://crbug.com/372842894 is addressed.
-                url,
+      context
+        .targetUnblockedOrThrow()
+        .then(() => context.applyDigitalCredentialsBehavior())
```
OrKoN: I think this is violating the architecture again. We should do this in src/bidiMapper/modules/cdp/CdpTarget.ts like it was before.

src/bidiMapper/modules/browser/ContextConfigStorage.ts
mohamedamir: Done!
Sorry!

src/bidiMapper/modules/context/BrowsingContextImpl.ts
mohamedamir: Sorry, I tried to update the project GEMINI.md to make sure this mistake doesn't happen again!

src/bidiMapper/modules/context/BrowsingContextImpl.ts
```diff
@@ -160,6 +166,9 @@ export class BrowsingContextImpl {
     browsingContextStorage.addContext(context);
     if (!context.isTopLevelContext()) {
       context.parent!.addChild(context.id);
+      if (context.cdpTarget === context.parent!.cdpTarget) {
```
OrKoN, line 169: I think this might be not needed (and flaky). If the local frame is just being created, there should be no behavior targeting it directly by id yet. If there is a global target behavior, it should apply via CdpTarget. WDYT?

src/bidiMapper/modules/cdp/CdpTarget.ts
```diff
@@ -719,22 +720,27 @@ export class CdpTarget {
...
-            protocol: config.digitalCredentialsBehavior.protocol,
-            response: config.digitalCredentialsBehavior.response,
-          })
-          .catch(() => {
-            // Ignore CDP errors, as the command is not supported by iframe targets
-            // or older browser versions.
-          }),
-      );
-    }
+    const dcConfig = this.contextConfigStorage.getActiveConfig(
```
OrKoN, line 723: do we always have a dcConfig? should not sometimes nothing be configured?

src/bidiMapper/modules/context/BrowsingContextImpl.ts
mohamedamir: OK, I have been investigating this.
The problem is the [logic in Chromium](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/digital_credentials/digital_credential_environment.cc;l=54-64;drc=14743e7bf12bf554772a7f42f43471a598b2c10e) inherits the behaviour from the parent which we decided against here.

I am sorry I overlooked this during the code review in Chromium.

What do you think is the best path forward now?
The spec doesn't define this parent-frame behaviour inheritance.
Should I update the logic in Chromium and then comeback to this PR?
Do you see any other path forward?

src/bidiMapper/modules/cdp/CdpTarget.ts
mohamedamir: It seems [getAtiveConfig](https://github.com/GoogleChromeLabs/chromium-bidi/blob/63c1596f3fe83f2c38fa46919170b2e19b0ee56e/src/bidiMapper/modules/browser/ContextConfigStorage.ts#L106-L130) always returns something, even when nothing is configured.
But then the logic below 
`const dcBehavior = dcConfig.digitalCredentialsBehavior;` might indeed be undefined, which is handled in the code AFAICT.

Am I overlooking something?

OrKoN: I think it does kind of handle it but we might also save sending setVirtualWalletBehavior if nothing was configured at all (as opposed to clear action). Maybe it's something for a follow-up.

src/bidiMapper/modules/context/BrowsingContextImpl.ts
OrKoN: Yeah I think in Chromium we should not inherit. To summarize, if Chromium does not inherit and the spec does not define inheriting:

1) when a new target is created, the client could not have known the targetId so we only need to enable if we have userContext level configuration. The spec does not specify the userContext level configuration so we should not expect userContext configuration in the implementation.
2) when frame exists, it can be targeted by context ID.

So indeed I think we should be good with just doing the configuration when unblocking CDP targets and removing this part. Does it sound correct?

PR #4226, opened 2026-07-23
chore: fix chromium integration

OrKoN:


tools/run_unittests.py
```diff
@@ -0,0 +1,64 @@
...
+    parser.add_argument("--gen-dir", required=True)
+    parser.add_argument("--node-py", required=True)
+    parser.add_argument("args", nargs=argparse.REMAINDER)
+    args = parser.parse_args()
+
+    # The current directory may be the root of the checkout or the build dir (e.g. out/Default).
+    # Use __file__ to reliably find the source directory for chromium-bidi.
+    src_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
+    dst_dir = os.path.abspath(
+        os.path.join(args.gen_dir, "third_party", "chromium-bidi")
```
sadym-chromium, line 37: are you sure this script would work in stand-alone checkout (outside of the chromium tree)?

sadym-chromium (approved):

tools/run_unittests.py
OrKoN: it does not need to work in a stand-alone checkout. It's only for chromium.

OrKoN: in standalone checkout use `npm run unit` as usual.

PR #4231, opened 2026-07-28
build: support vpython envs

OrKoN:
Docs https://source.chromium.org/chromium/chromium/src/+/main:.vpython3;l=3;drc=6b7ca6159eaea5574341e2c6570ec600f9747601


sadym-chromium (approved):
LGTM after CI is fixed

PR #4241, opened 2026-08-19
build: fix local build

sadym-chromium:


OrKoN (changes requested):
Could you please elaborate how this fixes the local build?

PR #4242, opened 2026-08-19
build: sync package versions

OrKoN:


sadym-chromium (approved):

package.json
```diff
@@ -58,7 +58,7 @@
     "@types/sinon": "21.0.1",
     "@types/websocket": "1.0.10",
     "chai": "6.2.2",
-    "devtools-protocol": "0.0.1662358",
+    "devtools-protocol": "0.0.1655900",
```
sadym-chromium, line 61: was it intended to downgrade CDP version?

OrKoN: yes, because the downgraded version was effectively used.

